PR #55, opened 2020-08-05
add password for curvefs_tool && disable error log output to console

wu-hanqing:


curvefs_python/test.py
```diff
@@ -37,6 +37,14 @@ def exec_cmd(cmd):
 exec_cmd(cmd)
 cmd = "curve list --user k8s --dirname /k8s"
 exec_cmd(cmd)
+# 不是root，失败
```
aspirer, line 40: use English pls :)

wu-hanqing: Done

PR #85, opened 2020-09-02
fix clone has same dest bug

xu-chaojie:

Issue Number: close #xxx 

Problem Summary:

fix a bug in snapshotcloneserver when we clone a file and delete it，then clone a file which has same filename will return a failure code -19。

What's Changed:

change snapshotclone server code in clone.cpp to fix it。

How it Works:

When we clone a file which has same filename and notexist ，we try to search it on mds to find it notexist， then we continue to clone。

nothing

### Check List

### DIFF
--- a/test/integration/snapshotcloneserver/snapshotcloneserver_common_test.cpp
+++ b/test/integration/snapshotcloneserver/snapshotcloneserver_common_test.cpp
@@ -938,5 +938,58 @@ TEST_F(SnapshotCloneServerTest, TestSnapAndCloneWhenSnapHasError) {
     ASSERT_EQ(kErrCodeFileNotExist, retCode);
 }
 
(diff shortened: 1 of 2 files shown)


wu-hanqing (approved):

bai-charisu (approved):

test/integration/snapshotcloneserver/snapshotcloneserver_common_test.cpp
```diff
@@ -938,5 +938,34 @@ TEST_F(SnapshotCloneServerTest, TestSnapAndCloneWhenSnapHasError) {
     ASSERT_EQ(kErrCodeFileNotExist, retCode);
 }
```
ilixiaocui, line 940: source不一样的情况也测试一下？

ilixiaocui (approved):

PR #235, opened 2021-01-26
mds : fix clean manager retry

xu-chaojie:

fix clean manager retry & clone from snapshot do not send clonesource

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/src/mds/nameserver2/clean_task.h
+++ b/src/mds/nameserver2/clean_task.h
@@ -37,8 +37,13 @@ namespace mds {
 
 typedef  uint64_t TaskIDType;
 
+// default clean task retry times
+const uint32_t kDefaultTaskRetryTimes = 5;
+
 class Task {
  public:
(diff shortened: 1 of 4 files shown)


src/mds/nameserver2/clean_task.h
```diff
@@ -67,9 +73,24 @@ class Task {
...
+    }
+
+    void Retry() {
+        retry_--;
+        progress_ = TaskProgress();
+    }
+
+    bool RetryTimesExceed() {
+        return retry_ == 0;
+    }
```
wu-hanqing, line 87: 这两个是不是可以放到一个函数里面去？

xu-chaojie: 考虑过好像不太灵活

ilixiaocui (approved):

wu-hanqing (approved):

PR #288, opened 2021-03-10
fix mds log level, fix snaptool

cw123:

Problem Summary: fix mds log level, fix snaptool

What's Changed: fix mds log level, fix snaptool

How it Works: fix mds log level, fix snaptool

Side effects(Breaking backward compatibility? Performance regression?):  NO

### Check List


cw123:
recheck

wu-hanqing (approved):

PR #299, opened 2021-03-24
mds,client,snapshotcloneserver: fix bug CreateCloneFile missing param…

xu-chaojie:
… stripeUnit stripeCount

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/proto/snapshotcloneserver.proto
+++ b/proto/snapshotcloneserver.proto
@@ -47,6 +47,8 @@ message SnapshotInfoData {
     required uint64 fileLength = 8;
     required uint64 time = 9;
     required int32 status = 10;
+    optional uint64 stripeUnit = 11;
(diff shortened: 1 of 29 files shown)


ilixiaocui (approved):

wu-hanqing (approved):

proto/snapshotcloneserver.proto
```diff
@@ -47,6 +47,8 @@ message SnapshotInfoData {
     required uint64 fileLength = 8;
     required uint64 time = 9;
     required int32 status = 10;
+    optional uint64 stripeUint = 11;
```
wu-hanqing, line 50: should be `stripeUnit`

xu-chaojie: done

PR #302, opened 2021-03-25
nbd: fix concurrent nbd map

wu-hanqing:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/nbd/src/util.cpp
+++ b/nbd/src/util.cpp
@@ -89,54 +89,6 @@ int get_nbd_max_count() {
     return nbds_max;
 }
 
-std::string find_unused_nbd_device() {
-    int index = 0;
-    int devfd = 0;
-    int nbds_max = get_nbd_max_count();
-    char dev[64];
-    int sockfd[2];
-
-    int ret = socketpair(AF_UNIX, SOCK_STREAM, 0, sockfd);
-    if (ret < 0) {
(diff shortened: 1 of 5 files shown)


SeanHai (approved):

nbd/src/util.cpp
```diff
@@ -441,6 +393,74 @@ int load_module(NBDConfig *cfg) {
     return ret;
 }
 
+int map_on_unused_nbd_device(int sockfd, std::string* devpath, int* nbdindex) {
```
xu-chaojie, line 396: This function can be protect function in the parent class ’NBDController' ?

wu-hanqing: Done

nbd/src/util.cpp
```diff
@@ -441,6 +393,74 @@ int load_module(NBDConfig *cfg) {
...
+             << nbds_max << ", last try nbd index: " << index << std::endl;
+        return -1;
+    } else {
+        *devpath = dev;
+        *nbdindex = index;
+        return devfd;
+    }
+}
+
+int map_on_nbd_device_by_devpath(int sockfd, const std::string& devpath,
```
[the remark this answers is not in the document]
wu-hanqing: Done

xu-chaojie (approved):

Issue #327, opened 2021-04-21
【C-Plan选题一：清理代码中的TODO】add return value check


zzl200012:
**Describe the task you choose (描述你选择的任务)**
`// TODO(wuhanqing): 判断返回值` at [src/client/libcbd_libcurve.cpp](https://github.com/opencurve/curve/blob/master/src/client/libcbd_libcurve.cpp#L93)

**Describe alternatives you've considered (optional) (描述你想到的方案(可选))**

**Additional context/screenshots (更多上下文/截图)**

ilixiaocui:
thank u for submit！

PR #328, opened 2021-04-21
fix a todo: add return value check

zzl200012:

Issue Number: #327  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/src/client/libcbd_libcurve.cpp
+++ b/src/client/libcbd_libcurve.cpp
@@ -91,8 +91,12 @@ int64_t cbd_libcurve_filesize(const char* filename) {
     memset(&info, 0, sizeof(info));
 
     // TODO(wuhanqing): 判断返回值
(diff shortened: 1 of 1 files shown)


src/client/libcbd_libcurve.cpp
```diff
@@ -91,8 +91,12 @@ int64_t cbd_libcurve_filesize(const char* filename) {
     memset(&info, 0, sizeof(info));
 
     // TODO(wuhanqing): 判断返回值
-    StatFile4Qemu(filename, &info);
-    return info.length;
+    int ret = StatFile4Qemu(filename, &info);
```
wu-hanqing, line 94: remove TODO annotation

zzl200012: fix

wu-hanqing (approved):

ilixiaocui (approved):

Issue #335, opened 2021-04-29
【C-Plan选题二：捉虫计划】fix a initialization error.


Super-long:
**Describe alternatives you've considered (optional) **

in /curve/test/chunkserver/multiple_copysets_io_test.cpp function update_leader()

The leader field in the info structure is not explicitly initialized, which means that its initial value is 0, but in function update_leader, if the first if is judged to be flase because of the failure of the election of the leader, it will directly return to copyset->leader, but zero is a valid one Value, this will cause an error.

So I think that the initial value of copyset->leader should be set to -1, and a meaningless judgment in update_leader should be deleted, which will cause the boundary to be crossed after the initialization is -1

PR #336, opened 2021-04-29
fix bug;

Super-long:

Issue Number: [335](https://github.com/opencurve/curve/issues/335)

Problem Summary:

copyset->leader is not explicitly initialized, which means that its initial value is 0, and the leader will be returned directly when it is judged in update_leader() that the election of the leader fails, which results in a correct value being returned in the case of an error.

/curve/test/chunkserver/multiple_copysets_io_test.cpp
- function update_leader()
- function init_copysets()

Initialize copyset->leader to -1, and delete a meaningless judgment in update_leader.

What's Changed:

How it Works:

It is necessary to traverse all the configurations every time to find the index corresponding to the leader.

### Check List


cw123:
recheck

PR #346, opened 2021-05-13
add ParseUri ErrorCode;

Super-long:

src/chunkserver/copyset_node.cpp need to add an error code of ParseUri, so this PR solves the problem.

I modified the return value of the ParseUri function and defined three ErrorCodes, the type is ParseUrlErrorCode. From the analysis of the source code, ParseUri is actually just doing a parsing of the existing storage protocol, and the return value is only used to exit when the protocol parsing fails, so only three error codes are defined. As far as the storage protocol is concerned, only "local" and "curve" are currently valid.

Because of the modification of the return value type, I also modified all the code that references function ParseUri.

In the previous parseuri, uri.find_last_not_of(" ") was used for the analysis of Uri, but in fact, the space is also a valid file character, you can refer to here: https://zh.wikipedia.org/wiki/%E6%AA%94%E6%A1%88%E5%90%8D%E7%A8%B1

So I deleted this code.

### Check List


src/chunkserver/uri_paser.h
```diff
@@ -32,29 +32,51 @@ namespace chunkserver {
...
-    static std::string ParseUri(const std::string& uri,
-                             std::string * param) {
+    // Other error codes can be added later.
+    enum ParseUrlErrorCode{
+        SUCESS = 0,
+        EPROTOCALEMPTY = 1,
+        EPROTOCALINVAILD = 2,
+    };
+
+    static ParseUrlErrorCode ParseUri(const std::string& uri,
```
wu-hanqing, line 42: I think you have changed this function purpose. The previous version parses a Uri, and corresponding protocol and parameters are returned by return value and argument separately, also if the Uri is invalid, the return value is empty. But the after version returns parameters and whether the protocol is valid, the protocol type is lossed.

ilixiaocui:
cc @Super-long Please resubmit to the master branch if necessary.

PR #348, opened 2021-05-14
nbd: prevent mounted device to unmap

Wine93:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/nbd/src/main.cpp
+++ b/nbd/src/main.cpp
@@ -88,6 +88,7 @@ static void Usage() {
         << "  --timeout <seconds>     Set nbd request timeout\n"
         << "  --try-netlink           Use the nbd netlink interface\n"
         << "Unmap options:\n"
+        << "  -f, --force                 Force unmap even if the device is mounted\n"              // NOLINT
         << "  --retry_times <limit>       The number of retries waiting for the process to exit\n"  // NOLINT
         << "                              (default: " << nbdConfig->retry_times << ")\n"            // NOLINT
         << "  --sleep_ms <milliseconds>   Retry interval in milliseconds\n"                         // NOLINT
--- a/nbd/src/util.cpp
+++ b/nbd/src/util.cpp
@@ -153,6 +153,8 @@ int parse_args(std::vector<const char*>& args, std::ostream *err_msg,   // NOLIN
             }
         } else if (argparse_flag(args, i, "--try-netlink", (char *)NULL)) { // NOLINT
             cfg->try_netlink = true;
+        }  else if (argparse_flag(args, i, "-f", "--force", (char *)NULL)) {  // NOLINT
+            cfg->force_unmap = true;
         } else if (argparse_witharg(args, i, &cfg->retry_times, err, "--retry_times", (char*)(NULL))) {  // NOLINT
             if (!err.str().empty()) {
                 *err_msg << "curve-nbd: " << err.str();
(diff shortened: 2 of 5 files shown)


Wine93:
I hava test the tool option in my own , it works well, see below:

* STEP1: map image
```
root@work:~# curve-nbd map cbd:pool//test_test_
root@work:~# curve-nbd list-mapped
id      image                device
1654834 cbd:pool//test_test_ /dev/nbd0
```

* STEP 2: unmap without mount -> success
```
root@work:~# curve-nbd unmap /dev/nbd0
root@work:~# curve-nbd list-mapped
```

* STEP 3: remap image and mount it
```
root@work:~# curve-nbd map cbd:pool//test_test_
root@work:~# curve-nbd list-mapped
id      image                device
1655046 cbd:pool//test_test_ /dev/nbd0

root@work:~# mount /dev/nbd0 /home/data
```

* STEP 4: unmap without `-f` parm -> failed
```
root@work:~# curve-nbd unmap /dev/nbd0
2021-05-14 16:17:45 1657770 nbd/src/util.cpp:309] curve-nbd: the /dev/nbd0 is still mount on /home/data, you can't unmap it or specify -f parameter
root@work:~# curve-nbd list-mapped
id      image                device
1655046 cbd:pool//test_test_ /dev/nbd0
```

* STEP 5: map with `-f` parm -> success
```
root@work:~# curve-nbd unmap -f /dev/nbd0
2021-05-14 16:18:03 1657900 nbd/src/util.cpp:304] curve-nbd: the /dev/nbd0 is still mount on /home/data, force unmap it
root@work:~# curve-nbd list-mapped
```

nbd/src/util.h
```diff
@@ -60,6 +60,8 @@ extern int check_block_size(int nbd_index, uint64_t expected_size);
 extern int check_device_size(int nbd_index, uint64_t expected_size);
 // 如果当前系统还未加载nbd模块，则进行加载；如果已经加载，则不作任何操作
 extern int load_module(NBDConfig *cfg);
+// Check the mount status and judge whether it can be unmap
+extern int check_mount_status(const NBDConfig *cfg);
```
wu-hanqing, line 64: `extern` in here is redundant, and maybe `check_dev_can_unmap` is more obvious

nbd/src/util.cpp
```diff
@@ -278,6 +280,37 @@ int get_mapped_info(int pid, NBDConfig *cfg) {
     return 0;
 }
 
+int check_mount_status(const NBDConfig *cfg) {
+    std::ifstream ifs;
+    ifs.open("/proc/mounts", std::ifstream::in);
```
wu-hanqing, line 285: You can merge this as one line code `std::ifstream ifs("/proc/mounts", std::ifstream::in);`

nbd/src/util.cpp
```diff
@@ -278,6 +280,37 @@ int get_mapped_info(int pid, NBDConfig *cfg) {
...
+
+    if (!ifs.is_open()) {
+        dout << "curve-nbd: failed to open /proc/mounts" << std::endl;
+        return -EINVAL;
+    }
+
+    std::string device, mountPath;
+    bool mounted = false;
+    while (ifs >> device >> mountPath) {
```
wu-hanqing, line 294: Each line of `/proc/mounts` has six columns, so the input in `while condition` has a problem.
In the first time, `device` stores the first column, and `mountPath` stores the second column, but the next time, `device` will store the third column - filesystem type, and `mountPath` will store the fourth column which is the mount options.

SeanHai, line 294: maybe can read line and deal after?

nbd/src/util.h
Wine93: > check_dev_can_unmap

done

nbd/src/util.cpp
Wine93: > td::ifstream ifs("/proc/mounts", std::ifstream::in);

done.

nbd/src/util.cpp
Wine93: > Each line of `/proc/mounts` has six columns, so the input in `while condition` has a problem.
> In the first time, `device` stores the first column, and `mountPath` stores the second column, but the next time, `device` will store the third column - filesystem type, and `mountPath` will store the fourth column which is the mount options.

yeap, it's a **BUG**, i will fix it.

Wine93: > maybe can read line and deal after?

yeap, it's a way to deal with it.

wu-hanqing (approved):

SeanHai (approved):

nbd/src/main.cpp
```diff
@@ -88,6 +88,7 @@ static void Usage() {
         << "  --timeout <seconds>     Set nbd request timeout\n"
         << "  --try-netlink           Use the nbd netlink interface\n"
         << "Unmap options:\n"
+        << "  -f, --force                 Force unmap even if the device is mounted\n"              // NOLINT
```
ilixiaocui, line 91: force unmap will lead to filesystem readonly

Wine93: > force unmap will lead to filesystem readonly

yeap, it will lead to filesystem become readonly.

BTW: I have constructed a test, that client write one 4K IO not success (RPC error) and keep RPC retrying, and then i do the `umount` and `curve-nbd unmap` operation, the result is:

* we can't umount the device, even if we specify `-f` option
* but we can do `curve-nbd unmap` success
* after we do the `curve-nbd unmap` operation, the filesystem is still on `read-write` mode, but if we read/write on the filesystem, it will become `read-only` mode.

ilixiaocui (approved):

PR #351, opened 2021-05-14
fix memory order;

Super-long:

Modified relaxed semantics to acquire-release semantics.

Problem Summary:

memory_order_relaxed semantics is not suitable for this scenario.

### Check List

### DIFF
--- a/src/chunkserver/copyset_node.cpp
+++ b/src/chunkserver/copyset_node.cpp
@@ -505,7 +505,7 @@ std::string CopysetNode::GetCopysetDir() const {
 
 uint64_t CopysetNode::GetConfEpoch() const {
     std::lock_guard<std::mutex> lockguard(confLock_);
-    return epoch_.load(std::memory_order_relaxed);
+    return epoch_.load(std::memory_order_acquire);
 }
 
 int CopysetNode::LoadConfEpoch(const std::string &filePath) {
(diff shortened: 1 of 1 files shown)


src/chunkserver/copyset_node.cpp
```diff
@@ -526,7 +526,7 @@ int CopysetNode::LoadConfEpoch(const std::string &filePath) {
                        << ", Copyset: " << GroupIdString();
             ret = -1;
         } else {
-            epoch_.store(loadEpoch, std::memory_order_relaxed);
+            epoch_.store(loadEpoch, std::memory_order_release);
```
ilixiaocui, line 529: LoadConfEpoch is called on copyset in initialization，there is no competition at this time.

src/chunkserver/copyset_node.cpp
```diff
@@ -505,7 +505,7 @@ std::string CopysetNode::GetCopysetDir() const {
 
 uint64_t CopysetNode::GetConfEpoch() const {
     std::lock_guard<std::mutex> lockguard(confLock_);
-    return epoch_.load(std::memory_order_relaxed);
+    return epoch_.load(std::memory_order_acquire);
```
ilixiaocui, line 508: GetConfEpoch()  already get lock

PR #358, opened 2021-05-19
client: fix segment fault in SourceReader

wu-hanqing:

Problem Summary:

After #172 and #209, source volume data of the clone volume can be read within the client through SourceReader, and each source volume was also opened and represented by a FileInstance, but `FileInstance::mdsclient_` points to `FileClient::mdsClient_`, so after FileClient is destroyed, `FileInstance::mdsclient_` becomes dangling pointer.

What's Changed:

To fix this problem, we let `FileClient::mdsClient_` be a shared_ptr, and each FileInstance holds ownership of it.

### Check List

### DIFF
--- a/src/client/client_metric.h
+++ b/src/client/client_metric.h
@@ -30,6 +30,7 @@
 
 #include "src/common/timeutility.h"
 #include "src/client/client_common.h"
+#include "src/common/string_util.h"
 
 using curve::common::TimeUtility;
 
@@ -200,8 +201,7 @@ struct MDSClientMetric {
     explicit MDSClientMetric(const std::string& prefix_ = "")
         : prefix(!prefix_.empty()
                      ? prefix_
-                     : "curve_mds_client_" +
-                           std::to_string(reinterpret_cast<uint64_t>(this))),
+                     : "curve_mds_client_" + common::ToHexString(this)),
           metaserverAddress(prefix, "current_metaserver_addr", GetStringValue,
                             &metaserverAddr),
           openFile(prefix, "openFile"),
--- a/src/client/mds_client.cpp
+++ b/src/client/mds_client.cpp
@@ -22,12 +22,12 @@
 #include "src/client/mds_client.h"
 
 #include <glog/logging.h>
+#include <brpc/errno.pb.h>
 
 #include <utility>
 #include <algorithm>
 
 #include "src/client/lease_executor.h"
-#include "src/client/metacache.h"
 #include "src/common/net_common.h"
 #include "src/common/string_util.h"
 #include "src/common/timeutility.h"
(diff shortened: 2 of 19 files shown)


src/client/mds_client.h
```diff
@@ -39,25 +38,19 @@
...
-using curve::common::RWLock;
-using curve::common::ReadLockGuard;
-using curve::common::Authenticator;
-
-class MetaCache;
 struct LeaseRefreshResult;
 
 // MDSClient是client与MDS通信的唯一窗口
-class MDSClient {
+class MDSClient : public MDSClientBase,
```
xu-chaojie, line 48: Why dose ‘MDSClient’ change to inherit from 'MDSClientBase' ?

wu-hanqing: After removing the option from MDSClientBase, it becomes an empty class. So if we define a member variable of class MDSClientBase, it at least occupies 1-byte memory. To avoid taking up extra space, inherit it is a good solution. Also, make MDSClient inherits from MDSClientBase has no bad effect.

src/client/client_metric.h
```diff
@@ -200,8 +201,7 @@ struct MDSClientMetric {
     explicit MDSClientMetric(const std::string& prefix_ = "")
         : prefix(!prefix_.empty()
                      ? prefix_
-                     : "curve_mds_client_" +
-                           std::to_string(reinterpret_cast<uint64_t>(this))),
+                     : "curve_mds_client_" + common::ToHexString(this)),
```
SeanHai, line 204: why change to hexString? give an example for prefix

src/client/mds_client_base.h
```diff
@@ -100,7 +96,7 @@ class MDSClientBase {
      * @param: metaServerOpt为mdsclient的配置信息
```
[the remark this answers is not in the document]
wu-hanqing: Now, this function is useless, so I just delete it.

src/client/mds_client.cpp
```diff
@@ -81,20 +63,28 @@ LIBCURVE_ERROR MDSClient::Initialize(const MetaServerOption& metaServerOpt) {
 
     metaServerOpt_ = metaServerOpt;
 
-    int rc = mdsClientBase_.Init(metaServerOpt_);
+    int rc = MDSClientBase::Init();
```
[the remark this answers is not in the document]
wu-hanqing: fix

src/client/client_metric.h
wu-hanqing: The prefix in here is just to distinguish different MDSClientMetric's instances because for a specific name, bvar only expose once. So, I just add the instance's address in the prefix, and change the address to hex string is more obvious.

SeanHai (approved):

xu-chaojie (approved):

PR #373, opened 2021-05-31
tools: fixed remove broken copyset after remove peer

Wine93:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/src/chunkserver/copyset_service.cpp
+++ b/src/chunkserver/copyset_service.cpp
@@ -134,6 +134,31 @@ void CopysetServiceImpl::CreateCopysetNode2(RpcController *controller,
     LOG(INFO) << "Create " << request->copysets().size() << " copysets success";
 }
 
+void CopysetServiceImpl::DeleteBrokenCopyset(RpcController* controller,
+                                             const CopysetRequest* request,
+                                             CopysetResponse* response,
+                                             Closure* done) {
+    LOG(INFO) << "Receive delete broken copyset request";
+
+    brpc::ClosureGuard doneGuard(done);
+
+    auto poolId = request->logicpoolid();
+    auto copysetId = request->copysetid();
+    auto groupId = ToGroupIdString(poolId, copysetId);
+
+    // if copyset node exist in the manager means its data is complete
+    if (copysetNodeManager_->IsExist(poolId, copysetId)) {
+        response->set_status(COPYSET_OP_STATUS_COPYSET_IS_HEALTHY);
+        LOG(WARNING) << "Delete broken copyset, " << groupId  << " is healthy";
+    } else if (!copysetNodeManager_->DeleteBrokenCopyset(poolId, copysetId)) {
+        response->set_status(COPYSET_OP_STATUS_FAILURE_UNKNOWN);
+        LOG(ERROR) << "Delete broken copyset " << groupId << " failed";
+    } else {
+        response->set_status(COPYSET_OP_STATUS_SUCCESS);
+        LOG(INFO) << "Delete broken copyset " << groupId << " success";
+    }
+}
+
 void CopysetServiceImpl::GetCopysetStatus(RpcController *controller,
                                         const CopysetStatusRequest *request,
                                         CopysetStatusResponse *response,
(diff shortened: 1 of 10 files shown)


src/chunkserver/copyset_service.cpp
```diff
@@ -134,6 +134,30 @@ void CopysetServiceImpl::CreateCopysetNode2(RpcController *controller,
...
+    brpc::ClosureGuard doneGuard(done);
+
+    auto poolId = request->logicpoolid();
+    auto copysetId = request->copysetid();
+    auto groupId = ToGroupIdString(poolId, copysetId);
+
+    if (!copysetNodeManager_->IsExist(poolId, copysetId)) {
+        response->set_status(COPYSET_OP_STATUS_COPYSET_NOTEXIST);
+        LOG(WARNING) << "Remove copyset node " << groupId  << " not exist";
+    } else if (!copysetNodeManager_->PurgeCopysetNodeData(poolId, copysetId)) {
```
wu-hanqing, line 152: Should we check copyset status before purge its data？

Wine93: > Should we check copyset status before purge its data？

I think there is a bug in my implemention. actually, the copyset node will not exist in the `copysetNodeManager_ ` if its data has broken, because the `copysetNodeManager_` will ignore the data broken's copyset in the load phase when chunkserver start. So we will never delete the copyset node success.

Wine93: > Should we check copyset status before purge its data？

In the latest implementation, i will only delete the broken copyset.

Wine93:
recheck

src/chunkserver/copyset_service.h
```diff
@@ -30,6 +30,9 @@ namespace chunkserver {
 
 using ::google::protobuf::RpcController;
 using ::google::protobuf::Closure;
+using COPYSET_OP_STATUS::COPYSET_OP_STATUS_COPYSET_IS_HEALTHY;
```
wu-hanqing, line 33: For legacy C enum define, its internal enumerators are already exposed to its outer namespace, you can use `COPYSET_OP_STATUS_COPYSET_IS_HEALTHY` directly without the prefix `COPYSET_OP_STATUS::`,  so no need to use those using.

Wine93: > For legacy C enum define, its internal enumerators are already exposed to its outer namespace, you can use `COPYSET_OP_STATUS_COPYSET_IS_HEALTHY` directly without the prefix `COPYSET_OP_STATUS::`, so no need to use those using.

yeap, it's a bad use of plain enums, i will improve it.

src/tools/curve_cli.cpp
```diff
@@ -35,6 +35,8 @@ DEFINE_string(peer,
               "", "Id of the operating peer");
 DEFINE_string(new_conf,
               "", "new conf to reset peer");
+DEFINE_bool(remove_copyset, false, "Whether need to remove broken copyset "
```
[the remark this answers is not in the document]
Wine93: Actually, it's a dangerous action to delete copyset even if it's a broken copyset, we want to let user know what he is doning now, instead of silently deleting it.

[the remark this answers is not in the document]
Wine93: > why default is false? if the broken copyset data keep there, there will some problem

Actually, it's a dangerous action to delete copyset even if it's a broken copyset, we want to let user know what he is doning now, instead of silently deleting it.

src/chunkserver/copyset_service.cpp
```diff
@@ -134,6 +134,31 @@ void CopysetServiceImpl::CreateCopysetNode2(RpcController *controller,
     LOG(INFO) << "Create " << request->copysets().size() << " copysets success";
 }
 
+void CopysetServiceImpl::DeleteBrokenCopyset(RpcController* controller,
+                                             const CopysetRequest* request,
+                                             CopysetResponse* response,
+                                             Closure* done) {
+    LOG(INFO) << "Receive delete broken copyset request";
```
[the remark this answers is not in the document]
Wine93: > Add groupId in this LOG ?

The below logic has guranteed the group id will be print.

wu-hanqing (approved):

SeanHai (approved):

ilixiaocui (approved):

PR #409, opened 2021-07-04
mds: fix clone volume missing throttle params and may start two or more tasks to clean same segment

wu-hanqing:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/src/client/splitor.cpp
+++ b/src/client/splitor.cpp
@@ -387,7 +387,7 @@ bool Splitor::MarkDiscardBitmap(IOTracker* iotracker, FileSegment* fileSegment,
     fileSegment->SetBitmap(offset, len);
 
     if (fileSegment->IsAllBitSet()) {
-        iotracker->discardSegments_.push_back(segmentIndex);
+        iotracker->discardSegments_.emplace(segmentIndex);
     }
 
     return true;
--- a/src/mds/nameserver2/clean_manager.cpp
+++ b/src/mds/nameserver2/clean_manager.cpp
@@ -23,6 +23,7 @@
 #include <memory>
 #include <vector>
 #include "src/mds/nameserver2/clean_manager.h"
+#include "src/common/concurrent/count_down_event.h"
 
 namespace curve {
 namespace mds {
@@ -63,7 +64,8 @@ bool CleanManager::SubmitDeleteCommonFileJob(const FileInfo &fileInfo) {
 
 bool CleanManager::SubmitCleanDiscardSegmentJob(
     const std::string& cleanSegmentKey,
-    const DiscardSegmentInfo& discardSegmentInfo) {
+    const DiscardSegmentInfo& discardSegmentInfo,
+    curve::common::CountDownEvent* counter) {
     // get dlock
     dlockOpts_->pfx = std::to_string(discardSegmentInfo.fileinfo().id());
     dlock_ = std::make_shared<DLock>(*dlockOpts_);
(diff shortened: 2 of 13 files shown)


wu-hanqing:
recheck

src/mds/nameserver2/clean_manager.cpp
```diff
@@ -155,13 +157,22 @@ void CleanDiscardSegmentTask::ScanAndExecTask() {
...
-                                                             kv.second)) {
+            if (!cleanManager_->SubmitCleanDiscardSegmentJob(
+                    kv.first, kv.second, &counter)) {
                 LOG(ERROR) << "SubmitCleanDiscardSegmentJob failed";
-                continue;
+                counter.Signal();
             }
         }
+
+        // wait all submitted jobs finish
```
ilixiaocui, line 174: wait all jobs submitted?

src/client/splitor.cpp
```diff
@@ -387,7 +387,7 @@ bool Splitor::MarkDiscardBitmap(IOTracker* iotracker, FileSegment* fileSegment,
     fileSegment->SetBitmap(offset, len);
 
     if (fileSegment->IsAllBitSet()) {
-        iotracker->discardSegments_.push_back(segmentIndex);
+        iotracker->discardSegments_.emplace(segmentIndex);
```
ilixiaocui, line 390: Repeat push segmentIndex before?

src/mds/nameserver2/clean_manager.cpp
wu-hanqing: Not true, the logic here is loading persistent tasks and issuing asynchronous jobs for each task.
And after the job is completed, it deletes the corresponding persistent task.
So, if don't wait for all submitted jobs to finish, the next loop will load persistent tasks and issue jobs for those tasks, but there may exist the same jobs that do the same thing because the previous job hasn't finished and its corresponding persistent task remain undeleted.

src/client/splitor.cpp
wu-hanqing: Yes, if a segment is discarded, internal bitmap is all set, the condition here `fileSegment->IsAllBitSet()` is true.
The internal bitmap will be cleared by reading/writing operations or segment has been discarded, e.g., DeAllocateSegment RPC to MDS succeeded.
If RPC hasn't sent to MDS, and the same segment is been discarded again(can be triggered by `fstrim`), this condition will be always true.
Because the logic here is processing a chunk, so it will push segment_size/chunk_size same segment index.

wu-hanqing:
recheck

wu-hanqing:
recheck

src/mds/nameserver2/metric.h
```diff
@@ -66,15 +66,15 @@ class SegmentDiscardMetric {
...
+          totalCleanedSegments_(prefix_ + "_total_cleaned_segment_count"),
+          pendingSegments_(prefix_ + "_pending_segment_count"),
+          totalCleanedSize_(prefix_ + "_total_cleaned_size"),
+          pendingSize_(prefix_ + "_pending_size") {}
 
     ~SegmentDiscardMetric() = default;
 
-    void OnReceiveDiscardRequest(uint64_t size);
-    void OnDiscardFinish(uint64_t size);
+    void OnReceiveDiscardRequest(int64_t size);
```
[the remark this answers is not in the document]
wu-hanqing: Because in `OnDiscardFinish`, it will take a negative value of parameters, but for an unsigned value its result is unexpected. So, change all the parameters type to `int64_t`.

src/mds/nameserver2/clean_manager.cpp
ilixiaocui: ok

ilixiaocui (approved):

SeanHai (approved):

PR #414, opened 2021-07-07
curvefs: added block device client adapter

Wine93:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


curvefs/src/client/block_device_client.h
```diff
@@ -25,80 +25,148 @@
...
         const char* buf, off_t offset, size_t length) = 0;
 };
 
 class BlockDeviceClientImpl : public BlockDeviceClient {
  public:
-    explicit BlockDeviceClientImpl(std::shared_ptr<FileClient> fileClient) :
-        fileClient_(fileClient) {}
-    virtual ~BlockDeviceClientImpl() {}
+    explicit BlockDeviceClientImpl(std::shared_ptr<FileClient> fileClient)
+        : fileClient_(fileClient) {}
```
wu-hanqing, line 125: move `fileClient` to `fileClient_`.

And you forget set other member variables, before entering the function body, it will call its default constructor for each variable except built-in types, e.g., isOpen is not guaranteed to false, and fd is not guaranteed to 0.
So this will make `if (!isOpen_ || fd_ < 0) {` unexpected.

Either you can init member variables in the constructor initializer list or just give a default value in the member variable definition.

curvefs/src/client/block_device_client.h
```diff
@@ -25,80 +25,148 @@
...
 
-    CURVEFS_ERROR Init(const BlockDeviceClientOptions &options) override;
+    BlockDeviceClientImpl() = default;
```
wu-hanqing, line 127: why add an extra default constructor?

curvefs/src/client/block_device_client.h
```diff
@@ -25,80 +25,148 @@
...
+struct BlockDeviceStat {
+    uint64_t length;
+
+    // 0 -> FileCreated
+    // 1 -> FileDeleting
+    // 2 -> FileCloning
+    // 3 -> FileCloneMetaInstalled
+    // 4 -> FileCloned
+    // 5 -> FileBeingCloned
+    int status;
```
wu-hanqing, line 54: should we define this as enum?

curvefs/src/client/block_device_client.cpp
```diff
@@ -21,22 +21,199 @@
...
+
+CURVEFS_ERROR BlockDeviceClientImpl::WritePadding(char* writeBuffer,
+                                                  off_t writeStart,
+                                                  off_t writeEnd,
+                                                  off_t offset,
+                                                  size_t length) {
+    off_t readEnd = 0;
+    size_t writeLength = writeEnd - writeStart;
+
+    if (offset != writeStart) {  // Padding leading
```
wu-hanqing, line 156: there is exists one case that can be optimized, e.g., IO_ALIGNED_BLOCK_SIZE is 4k, and the original write request's size is 2k, but it crosses a block boundary, like the last 1k of the first block and the first 1k of the second block.
so, you can just issue one read 8k request, rather than two 4k read requests.

and, read/write here are all synchronous operations, but I'm not sure that whether fuse support asynchronous processing, if it supports it's better to use the asynchronous version of read/write API.

BTW, it's not urgent to optimized this, so you can leave it.

curvefs/src/client/block_device_client.h
```diff
@@ -25,80 +25,148 @@
...
 #include "src/client/libcurve_file.h"
 #include "curvefs/src/client/config.h"
 #include "curvefs/src/client/error_code.h"
 
-using ::curve::client::FileClient;
-
 namespace curvefs {
 namespace client {
 
+#define CURVE_ALIGN(d, a) (((d) + (a - 1)) & ~(a - 1))
```
wu-hanqing, line 39: change this to a function, and it probably is inlined.

curvefs/src/client/block_device_client.cpp
Wine93: > there is exists one case that can be optimized, e.g., IO_ALIGNED_BLOCK_SIZE is 4k, and the original write request's size is 2k, but it crosses a block boundary, like the last 1k of the first block and the first 1k of the second block.
> so, you can just issue one read 8k request, rather than two 4k read requests.

This is indeed a good optimization that I have overlooked, i will improve it.

> and, read/write here are all synchronous operations, but I'm not sure that whether fuse support asynchronous processing, if it supports it's better to use the asynchronous version of read/write API.
> 
> BTW, it's not urgent to optimized this, so you can leave it.

I will to confirm whether the fuse support asynchronous processing and imporve it if it supports.

curvefs/src/client/block_device_client.h
Wine93: > change this to a function, and it probably is inlined.

yep, i will improve it.

curvefs/src/client/block_device_client.h
Wine93: > should we define this as enum?

That would be clearer.

curvefs/src/client/block_device_client.h
Wine93: > why add an extra default constructor?

This is an unnecessary constructor, i will delete it.

curvefs/src/client/block_device_client.h
Wine93: > move `fileClient` to `fileClient_`.
> 
> And you forget set other member variables, before entering the function body, it will call its default constructor for each variable except built-in types, e.g., isOpen is not guaranteed to false, and fd is not guaranteed to 0.
> So this will make `if (!isOpen_ || fd_ < 0) {` unexpected.
> 
> Either you can init member variables in the constructor initializer list or just give a default value in the member variable definition.

This is indeed not a qualified constructor, i will fix it.

xu-chaojie: using "const std::shared_ptr<FileClient> &fileClient" as parameter  will be better.

curvefs/src/client/block_device_client.h
xu-chaojie: I need a constructor as follows, please see my pr: https://github.com/opencurve/curve/pull/402/files#diff-97023361921fe6f18ece5ed622166ee6036bab57b9553e4cdebfa094dfe7e898

And the "fd_" & "isOpen_" shoule be initailize in initailize list !

BlockDeviceClientImpl()  
      : fileClient_(std::make_shared<FileClient>()),
        fd_(-1),
        isOpen_(false)  {}

curvefs/src/client/block_device_client.h
```diff
@@ -25,80 +25,148 @@
...
+    CURVEFS_ERROR AlignRead(char* buf, off_t offset, size_t length);
+
+    CURVEFS_ERROR AlignWrite(const char* buf, off_t offset, size_t length);
+
+    bool IsAligned(off_t offset, size_t length);
+
+    Range CalcAlignRange(off_t start, off_t end);
+
+ private:
     uint64_t fd_;
```
xu-chaojie, line 161: Can fd be equal to 0?,  if deed, maybe use int64_t as fd will be better, thus  can use -1 as invalid fd.

curvefs/src/client/block_device_client.cpp
```diff
@@ -21,22 +21,199 @@
...
+    isOpen_ = true;
+    filename_ = filename;
+    owner_ = owner;
+    return CURVEFS_ERROR::OK;
+}
+
+CURVEFS_ERROR BlockDeviceClientImpl::Close() {
+    if (!isOpen_ || fd_ < 0) {
+        LOG(WARNING) << "File already closed";
+        return CURVEFS_ERROR::BAD_FD;
```
xu-chaojie, line 67: just return CURVEFS_ERROR::OK, no need to left error to up layer.

Wine93: > just return CURVEFS_ERROR::OK, no need to left error to up layer.

done

curvefs/src/client/block_device_client.h
Wine93: > I need a constructor as follows, please see my pr: https://github.com/opencurve/curve/pull/402/files#diff-97023361921fe6f18ece5ed622166ee6036bab57b9553e4cdebfa094dfe7e898
> 
> And the "fd_" & "isOpen_" shoule be initailize in initailize list !
> 
> BlockDeviceClientImpl()
> : fileClient_(std::make_shared()),
> fd_(-1),
> isOpen_(false) {}

done

curvefs/src/client/block_device_client.h
Wine93: > Can fd be equal to 0?, if deed, maybe use int as fd will be better, thus can use -1 as invalid fd.

done

curvefs/src/client/block_device_client.h
Wine93: > using "const std::shared_ptr &fileClient" as parameter will be better.

done

wu-hanqing (approved):

xu-chaojie (approved):

PR #415, opened 2021-07-08
curve_ops_tool: fix concurrent check copysets on server

SeanHai:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/src/tools/copyset_check_core.cpp
+++ b/src/tools/copyset_check_core.cpp
@@ -148,12 +148,21 @@ int CopysetCheckCore::CheckCopysetsOnChunkServer(
 }
 
 ChunkServerHealthStatus CopysetCheckCore::CheckCopysetsOnChunkServer(
-                    const std::string& chunkserverAddr,
-                    const std::set<std::string>& groupIds,
-                    bool queryLeader) {
+                                const std::string& chunkserverAddr,
+                                const std::set<std::string>& groupIds,
+                                bool queryLeader,
+                                std::pair<int, butil::IOBuf> *record,
+                                bool queryCs) {
     bool isHealthy = true;
+    int res = 0;
     butil::IOBuf iobuf;
-    int res = QueryChunkServer(chunkserverAddr, &iobuf);
+    if (queryCs) {
+        res = QueryChunkServer(chunkserverAddr, &iobuf);
+    } else {
+        res = record->first;
+        iobuf = record->second;
+    }
+
     if (res != 0) {
         // 如果查询chunkserver失败，认为不在线，把它上面所有的
         // copyset都添加到peerNotOnlineCopysets_里面
@@ -244,16 +253,18 @@ ChunkServerHealthStatus CopysetCheckCore::CheckCopysetsOnChunkServer(
             isHealthy = false;
         }
     }
+
     // 遍历没有leader的copyset
     bool health = CheckCopysetsNoLeader(chunkserverAddr,
                                         noLeaderCopysetsPeers);
     if (!health) {
         isHealthy = false;
     }
+
     // 遍历chunkserver发送请求
     for (const auto& item : csAddrMap) {
         ChunkServerHealthStatus res = CheckCopysetsOnChunkServer(item.first,
-                                                           item.second);
+                                                                 item.second);
         if (res != ChunkServerHealthStatus::kHealthy) {
             isHealthy = false;
         }
@@ -266,9 +277,9 @@ ChunkServerHealthStatus CopysetCheckCore::CheckCopysetsOnChunkServer(
 }
 
 bool CopysetCheckCore::CheckCopysetsNoLeader(const std::string& csAddr,
-                            const std::map<std::string,
-                                           std::vector<std::string>>&
-                                                copysetsPeers) {
+                                             const std::map<std::string,
+                                             std::vector<std::string>>&
+                                             copysetsPeers) {
     if (copysetsPeers.empty()) {
         return true;
     }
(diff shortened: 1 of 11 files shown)


src/tools/copyset_check_core.cpp
```diff
@@ -154,12 +154,14 @@ ChunkServerHealthStatus CopysetCheckCore::CheckCopysetsOnChunkServer(
     bool isHealthy = true;
     butil::IOBuf iobuf;
     int res = QueryChunkServer(chunkserverAddr, &iobuf);
+    statisticsMutest.lock();
```
wu-hanqing, line 157: This lock will bring some difficulties to maintenance.
The real problem that causes a lot of times is sending RPC to chunkserver, so I think you can change schemes here. You can concurrently send RPC to chunkserver and save RPC response. And if all RPC returned or failed, gather all responses and process in a single thread. So, this will not bring data race to processing RPC responses.

src/tools/copyset_check_core.h
```diff
@@ -105,9 +108,11 @@ const char kThreeCopiesInconsistent[] = "Three copies inconsistent";
 
 class CopysetCheckCore {
  public:
+    explicit CopysetCheckCore(std::shared_ptr<MDSClient> mdsClient) :
```
wu-hanqing, line 111: This overloaded constructor is duplicated, you can set the default value of csClient to nullptr in the blow constructor.

src/tools/copyset_check_core.cpp
```diff
@@ -638,13 +674,15 @@ void CopysetCheckCore::UpdateChunkServerCopysets(
 // 通过发送RPC检查chunkserver是否在线
 bool CopysetCheckCore::CheckChunkServerOnline(
                     const std::string& chunkserverAddr) {
-    int res = csClient_->Init(chunkserverAddr);
+    auto csClient = (csClient_ == nullptr) ?
```
wu-hanqing, line 677: create it in constructor

src/tools/copyset_check_core.h
SeanHai: > This overloaded constructor is duplicated, you can set the default value of csClient to nullptr in the blow constructor.

Change  csClient_ to csClient as a local var for concurrently send rpc, but unit test need to set mock csClient, so add this constructor.

src/tools/copyset_check_core.cpp
SeanHai: > create it in constructor

The reason same as above.

src/tools/copyset_check_core.cpp
Wine93: > This lock will bring some difficulties to maintenance.
> The real problem that causes a lot of times is sending RPC to chunkserver, so I think you can change schemes here. You can concurrently send RPC to chunkserver and save RPC response. And if all RPC returned or failed, gather all responses and process in a single thread. So, this will not bring data race to processing RPC responses.

I agree this suggestion. I think it's better handle RPC request by concurrent and handle the RPC responses together. It will make the logic more clearly and you can reuse a lot of code.

Wine93: > This lock will bring some difficulties to maintenance.
> The real problem that causes a lot of times is sending RPC to chunkserver, so I think you can change schemes here. You can concurrently send RPC to chunkserver and save RPC response. And if all RPC returned or failed, gather all responses and process in a single thread. So, this will not bring data race to processing RPC responses.

BTW：You only need one lock to save RPC responses in parallel.

src/tools/copyset_check_core.h
SeanHai: > This overloaded constructor is duplicated, you can set the default value of csClient to nullptr in the blow constructor.

done

src/tools/copyset_check_core.cpp
SeanHai: fixed

src/tools/copyset_check_core.cpp
```diff
@@ -351,9 +357,31 @@ int CopysetCheckCore::CheckCopysetsOnServer(const std::string& serverIp,
...
+        indexMutex.lock();
+        if (*index + 1 > chunkservers.size()) {
+            indexMutex.unlock();
+            break;
+        }
+        auto info = chunkservers[*index];
+        (*index)++;
+        indexMutex.unlock();
+        std::string csAddr = info.hostip() + ":" + std::to_string(info.port());
+        std::shared_ptr<butil::IOBuf> iobuf = std::make_shared<butil::IOBuf>();
```
wu-hanqing, line 374: IOBuf just holds some pointers which store real data address, it cheep to copy it.
so, it's not necessary to use shared_ptr to wrap it.

src/tools/copyset_check_core.cpp
```diff
@@ -148,12 +148,16 @@ int CopysetCheckCore::CheckCopysetsOnChunkServer(
...
-                    const std::set<std::string>& groupIds,
-                    bool queryLeader) {
+                            const std::string& chunkserverAddr,
+                            const std::set<std::string>& groupIds,
+                            bool queryLeader, butil::IOBuf *iobuf, int res) {
     bool isHealthy = true;
-    butil::IOBuf iobuf;
-    int res = QueryChunkServer(chunkserverAddr, &iobuf);
+    butil::IOBuf tmpBuf;
+    if (nullptr == iobuf) {
```
wu-hanqing, line 156: event it don't have some problems, but it's just looks wired.
I think is better to encapsulate last two parameters into a struct/class, and add another state which indicates whether we should `QueryChunkServer` or not.

src/tools/copyset_check_core.cpp
SeanHai: > IOBuf just holds some pointers which store real data address, it cheep to copy it.
> so, it's not necessary to use shared_ptr to wrap it.

fixed

src/tools/copyset_check_core.cpp
SeanHai: > event it don't have some problems, but it's just looks wired.
> I think is better to encapsulate last two parameters into a struct/class, and add another state which indicates whether we should `QueryChunkServer` or not.

fixed

wu-hanqing (approved):

Wine93 (approved):

ilixiaocui (approved):

PR #416, opened 2021-07-08
set unstable pendding empty chunkserver status to retired

cw123:

Problem Summary: set unstable pendding empty chunkserver status to retired

What's Changed:

How it Works:

### Check List

### DIFF
--- a/src/mds/heartbeat/chunkserver_healthy_checker.cpp
+++ b/src/mds/heartbeat/chunkserver_healthy_checker.cpp
@@ -134,8 +134,8 @@ void ChunkserverHealthyChecker::UpdateChunkServerOnlineState(
 
 bool ChunkserverHealthyChecker::TrySetChunkServerRetiredIfNeed(
     const HeartbeatInfo &info) {
-    // Disregard when chunkserver isn't in OFFLINE status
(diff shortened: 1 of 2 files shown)


src/mds/heartbeat/chunkserver_healthy_checker.cpp
```diff
@@ -134,8 +134,8 @@ void ChunkserverHealthyChecker::UpdateChunkServerOnlineState(
 
 bool ChunkserverHealthyChecker::TrySetChunkServerRetiredIfNeed(
     const HeartbeatInfo &info) {
-    // Disregard when chunkserver isn't in OFFLINE status
-    if (OnlineState::OFFLINE != info.state) {
+    // Disregard when chunkserver is in ONLINE status
+    if (OnlineState::ONLINE == info.state) {
```
xu-chaojie, line 138: Why pendding empty chunkserver is unstale ？

ilixiaocui (approved):

src/mds/heartbeat/chunkserver_healthy_checker.cpp
cw123: when you stop this chunkserver, this chunkserver state change from online to unstable.

PR #418, opened 2021-07-08
fix out of range of the ChunkServerInfo, maybe chunkserver status is …

dheart-joe:
…'RETIRED'

Issue Number: close #xxx 

Problem Summary: when chunkserver status was 'RETIRED', mdsclient will skip add chunkserverinfo into ChunkServerInfo list. 

What's Changed:

How it Works:

### Check List


dheart-joe:
recheck

PR #424, opened 2021-07-09
glog: fix chunkserver exit with core

### LINKED ISSUE #425: chunkserver will exit with core dump when receive SIGINT
**curve user feedback** 

**Describe the bug (描述bug)**
1. kill -2 $chunkserverid in normal enviroment
2. generate chunkserver core file

**To Reproduce (复现方法)**
kill -2 $chunkserverid 

**Expected behavior (期望行为)**
chunkserver exits normally

**Versions (各种版本)**
OS: debain9
Compiler:  use curve release version
curve-mds: v1.3.0-beta
curve-chunkserver: v1.3.0-beta
curve-snapshotcloneserver: v1.3.0-beta
curve-sdk: v1.3.0-beta
nebd: v1.3.0-beta
curve-nbd: v1.3.0-beta

image

image

ilixiaocui:
chunkserver will exit with core dump when receive SIGINT 

Issue Number: close #425  

Problem Summary:
1. we compile a debug version, and  locate abort() as follwing:
image

2. abort happens when destruct (static bthread::TimerThread timer_) in curve::common::LeakyBucket.
   The  lifecycle of static member variable in class  is the same as process
image

3.  bthread::TimerThread destruct with run() exit. run() need log print before exit.
image

4. log print need get vmodule_lock, which is a globle static variable
image

5.  globle static variable and  class static member variable  are both detruct before process exit. But the order of their destruction is determined by the compiler. If  vmodule_lock destruct before bthread::TimerThread, the process will core dump

What's Changed:

change static Mutex vmodule_lock -> static Mutex* vmodule_lock = new Mutex();

How it Works:

### Check List

### DIFF
--- a/thirdparties/glog/glog.patch
+++ b/thirdparties/glog/glog.patch
@@ -97,7 +97,7 @@ index 9968b96..f491f9b 100644
  #undef MUST_UNDEF_GFLAGS_DECLARE_MACROS
  #undef DECLARE_VARIABLE
 diff --git a/src/logging.cc b/src/logging.cc
-index 0c86cf6..669c12d 100644
+index 0c86cf6..b77a4d8 100644
 --- a/src/logging.cc
 +++ b/src/logging.cc
 @@ -27,6 +27,39 @@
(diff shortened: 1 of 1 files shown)


thirdparties/glog/glog.patch
```diff
@@ -1006,3 +1006,45 @@ index 9554718..e02bd02 100644
...
  static bool kFailureSignalHandlerInstalled = false;
+diff --git a/src/vlog_is_on.cc b/src/vlog_is_on.cc
+index e8fdbae..eb3a39e 100644
+--- a/src/vlog_is_on.cc
++++ b/src/vlog_is_on.cc
+@@ -123,7 +123,9 @@ struct VModuleInfo {
+ };
+ 
+ // This protects the following global variables.
+-static Mutex vmodule_lock;
```
wu-hanqing, line 1017: there's exists more than one global `Mutex`, you should check whether the same problem will happen on other global `Mutex`.

ilixiaocui:
recheck

thirdparties/glog/glog.patch
ilixiaocui: done

wu-hanqing (approved):

PR #428, opened 2021-07-12
nebd/curve-nbd: support multiple nebd

wu-hanqing:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curve-ansible/roles/install_package/templates/nebd-daemon.j2
+++ b/curve-ansible/roles/install_package/templates/nebd-daemon.j2
@@ -10,260 +10,304 @@
 # Description:       Start the nebd-daemon service and associated helpers
 ### END INIT INFO
 
-# 检查脚本执行是否通过root权限执行
-if [[ $(id -u) -ne 0 ]]
-then
-    echo "Please run with sudo"
-    exit 1
+NEBD_SERVER_BIN={{ curve_bin_dir }}/nebd-server
+STARTED_AS_ROOT=false
+USERS_FILE=/etc/nebd/users
+USERS=()
+
+if [[ $(id -u) -eq 0 ]]; then
+    STARTED_AS_ROOT=true
 fi
 
-# nebd-server路径
-bin={{ curve_bin_dir }}/nebd-server
-
-# 默认配置文件
-confPath={{ nebd_server_config_path }}
-
-# 日志文件路径
-baseLogPath={{ nebd_log_dir }}
-logPath=${baseLogPath}/server
-
-# pidfile
-pidFile=${baseLogPath}/nebd-server.pid
-
-# daemon log
-daemonLog=${baseLogPath}/nebd-server-daemon.log
-
-# console output
-consoleLog=${baseLogPath}/nebd-server-console.log
-
 function get_nebd_server_pid() {
-    local tmp=`pidof nebd-server`
+    local tmp=$(pgrep -u ${1} nebd-server)
     echo $tmp
 }
 
-# 启动nebd-server
-function start() {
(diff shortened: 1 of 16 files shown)


nebd/nebd-package/usr/bin/nebd-daemon
```diff
@@ -10,261 +10,305 @@
...
+    line2=$(cat ${DAEMON_LOG} | wc -l)
+    if [ $line1 != $line2 ]; then
+        echo "$1: start nebd-server met error!"
+        stop_one $1
     fi
 }
 
-# 停止daemon进程，且停止nebd-server
-function stop() {
+function stop_one() {
```
SeanHai, line 128: An administrator or maintener can only  start  most two nebd-server(user of root and himself) ?

curve-ansible/roles/install_package/templates/nebd-daemon.j2
```diff
@@ -10,261 +10,305 @@
...
-# 默认配置文件
-confPath={{ nebd_server_config_path }}
-
-# 日志文件路径
-baseLogPath={{ nebd_log_dir }}
-logPath=${baseLogPath}/server
-
-# pidfile
-pidFile=${baseLogPath}/nebd-server.pid
+NEBD_SERVER_BIN=/usr/bin/nebd-server
```
SeanHai, line 13: Before the nebd-daemon.j2 will use client.ini to generate the nebd-daemon.sh. If it has difficult to generate from jinja template, can the field like 'nebd_data_dir'  be delete ?

nbd/src/ImageInstance.cpp
```diff
@@ -31,7 +31,12 @@ namespace nbd {
 bool ImageInstance::Open() {
     int ret = 0;
 
-    ret = nebd_lib_init();
+    if (config_ && !config_->nebd_conf.empty()) {
+        ret = nebd_lib_init_with_conf(config_->nebd_conf.c_str());
+    } else {
+        ret = nebd_lib_init();
```
ilixiaocui, line 37: why not use same function: nebd_lib_init(conf)? Handle conf within nebd_lib_init.

nebd/src/part1/libnebd.cpp
```diff
@@ -43,6 +43,21 @@ int nebd_lib_init() {
     return ret;
 }
 
+int nebd_lib_init_with_conf(const char* confPath) {
```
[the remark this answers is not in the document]
wu-hanqing: functions in here are all C functions because thay are all decorated by `extern "C"`, so thay aren't support default parameters and overloading.

nebd/nebd-package/usr/bin/nebd-daemon
wu-hanqing: if start/stop/restart/status executed by root or with sudo, it will execute for every users that recoreded in `/etc/nebd/users` and root user which it default included.
operations with `_one` suffix only for current user, event run by root or with sudo.

curve-ansible/roles/install_package/templates/nebd-daemon.j2
wu-hanqing: `curve_bin_dir` was added backup, and another fileds now have default value.

nbd/src/ImageInstance.cpp
wu-hanqing: functions in here are all C functions because thay are all decorated by extern "C", so thay aren't support default parameters and overloading.

ilixiaocui (approved):

PR #447, opened 2021-07-20
nbd: fix concurrent nbd map

wu-hanqing:

What's Changed:

cherry-pick #302 to release1.2

### Check List


wu-hanqing:
recheck

PR #464, opened 2021-07-22
scripts: fix potential overflow when cnovert ip to value

wu-hanqing:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curve-ansible/roles/install_package/templates/mds-daemon.sh.j2
+++ b/curve-ansible/roles/install_package/templates/mds-daemon.sh.j2
@@ -21,6 +21,13 @@ daemonLog=${logPath}/curve-mds-daemon.log
 # console output
 consoleLog=${logPath}/curve-mds-console.log
 
+function ip_value() {
+  echo $1 | awk -F '[:/.]' '{
+    mask = (2 ^ 8)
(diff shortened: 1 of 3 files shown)


curve-ansible/roles/install_package/templates/mds-daemon.sh.j2
```diff
@@ -21,6 +21,18 @@ daemonLog=${logPath}/curve-mds-daemon.log
 # console output
 consoleLog=${logPath}/curve-mds-console.log
 
+function ip_value() {
+  ip=$(echo $1 | awk -F':' '{ printf $1 }')
```
Wine93, line 25: Maybe the function can be more simple:
```
echo $1 | awk -F '[/.]' '{ 
    mask = (2 ** 8) 
    print (($1 * mask + $2)  * mask + $3) * mask + $4 
}'
```

SeanHai (approved):

curve-ansible/roles/install_package/templates/mds-daemon.sh.j2
wu-hanqing: fix

Wine93 (approved):

ilixiaocui (approved):

PR #466, opened 2021-07-22
ansible: fix mds etcd daemon script

SeanHai:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curve-ansible/rolling_update_curve.yml
+++ b/curve-ansible/rolling_update_curve.yml
@@ -33,6 +33,7 @@
   roles:
     - { role: install_package, package_name: etcd, tags: install_etcd }
     - { role: generate_config, template_name: etcd.conf.yml, conf_path: "{{ etcd_config_path }}", tags: update_etcd_config }
+    - { role: install_package, package_name: etcd-daemon, tags: install_etcd_daemon }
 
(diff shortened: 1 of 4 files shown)


curve-ansible/rolling_update_curve.yml
```diff
@@ -33,6 +33,7 @@
   roles:
     - { role: install_package, package_name: etcd, tags: install_etcd }
     - { role: generate_config, template_name: etcd.conf.yml, conf_path: "{{ etcd_config_path }}", tags: update_etcd_config }
+    - { role: install_package, package_name: etcd-daemon, tags: install_etcd_daemon }
```
ilixiaocui, line 36: ags:
    - etcd
    - never
    
should never be deleted?

wu-hanqing (approved):

curve-ansible/rolling_update_curve.yml
SeanHai: > ags:
> - etcd
> - never
> 
> should never be deleted?

Don't need, run this yml use '--tags install_etcd_daemon' is ok.

ilixiaocui (approved):

PR #497, opened 2021-08-10
curvefs: add default permission check

SeanHai:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


curvefs/src/client/main.c
```diff
@@ -43,6 +43,12 @@ static const struct fuse_lowlevel_ops curve_ll_oper = {
 };
 
 int main(int argc, char *argv[]) {
+    // add default_permissions and allow_other
```
xu-chaojie, line 46: 这个地方不要这么改，改到启动脚本里面去，或者在下面的opts中直接设置默认值比较好

xu-chaojie (changes requested):

curvefs/src/client/main.c
SeanHai: > It will be better to add "default_permissions" and "allow_other" in the startup script or add these default option in the "opts" below.

fixed

xu-chaojie (approved):

wu-hanqing (approved):

PR #503, opened 2021-08-19
curvefs: support open flags

SeanHai:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


curvefs/src/client/fuse_s3_client.cpp
```diff
@@ -149,21 +181,31 @@ CURVEFS_ERROR FuseS3Client::FuseOpWrite(fuse_req_t req, fuse_ino_t ino,
...
     Inode inode;
     CURVEFS_ERROR ret = inodeManager_->GetInode(ino, &inode);
     if (ret != CURVEFS_ERROR::OK) {
         LOG(ERROR) << "inodeManager get inode fail, ret = " << ret
                   << ", inodeid = " << ino;
         return ret;
     }
+
+    if (fi->flags & O_DIRECT) {  // check align
+        if (off % DefaultBlksize != 0 || size % DefaultBlksize !=0) {
```
wu-hanqing, line 204: you can use this function to check whether a value is aligned to another value which is power of 2

https://github.com/opencurve/curve/blob/a89a10d02d389de1457d7b527701fffb38904802/curvefs/src/space/fast_align.h#L71

curvefs/src/client/fuse_volume_client.cpp
```diff
@@ -142,6 +142,22 @@ void FuseVolumeClient::FuseOpDestroy(void *userdata) {
     return;
 }
 
+CURVEFS_ERROR FuseVolumeClient::FuseOpOpen(fuse_req_t req, fuse_ino_t ino,
```
wu-hanqing, line 145: Does `FuseS3Client::FuseOpOpen` and `FuseVolumeClient::FuseOpen` are only different in truncate ?
If so, you can leave base class's `FuseOpOpen`, add a private virtual member function, e.g., `Truncate`, and call this function if open flags with O_TRUNC.
For the `Trunate` function, the derived class must override it, and do it's own work.

curvefs/src/client/fuse_client.cpp
```diff
@@ -347,7 +334,7 @@ CURVEFS_ERROR FuseClient::FuseOpSetAttr(
     if (to_set & FUSE_SET_ATTR_MTIME) {
         inode.set_mtime(attr->st_mtime);
     }
-    uint64_t nowTime = TimeUtility::GetTimeofDayMs();
+    uint64_t nowTime = TimeUtility::GetTimeofDaySec();
```
wu-hanqing, line 337: why change this percision ?

SeanHai: > why change this percision ?

The time unit of atime/ctime/mtime in inode need to fix later.

curvefs/src/client/fuse_client.h
```diff
@@ -40,6 +40,8 @@
 #include "curvefs/src/client/s3/client_s3_adaptor.h"
 #include "curvefs/proto/common.pb.h"
 
+#define DefaultBlksize 512
```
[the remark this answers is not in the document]
SeanHai: > `DirectIOAlignemnt` may be better, because backend storage can be objest storage serivce.

fixed.

curvefs/src/client/fuse_volume_client.cpp
SeanHai: > Does `FuseS3Client::FuseOpOpen` and `FuseVolumeClient::FuseOpen` are only different in truncate ?
> If so, you can leave base class's `FuseOpOpen`, add a private virtual member function, e.g., `Truncate`, and call this function if open flags with O_TRUNC.
> For the `Trunate` function, the derived class must override it, and do it's own work.

fixed

curvefs/src/client/fuse_s3_client.cpp
SeanHai: > you can use this function to check whether a value is aligned to another value which is power of 2
> 
> https://github.com/opencurve/curve/blob/a89a10d02d389de1457d7b527701fffb38904802/curvefs/src/space/fast_align.h#L71

done

curvefs/proto/metaserver.proto
```diff
@@ -114,6 +114,7 @@ message S3ChunkInfoList {
     repeated S3ChunkInfo s3Chunks = 1;
 };
 
+// Todo: time unit of atime/ctime/mtime need fix
```
xu-chaojie, line 117: then fix it

wu-hanqing (approved):

PR #545, opened 2021-09-18
curvefs/mds: fix bugs and add retry policy for metaserverclient

SeanHai:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/mds/fs_manager.cpp
+++ b/curvefs/src/mds/fs_manager.cpp
@@ -36,7 +36,6 @@ using NameLockGuard = ::curve::common::GenericNameLockGuard<Mutex>;
 using curvefs::mds::topology::TopoStatusCode;
 
 bool FsManager::Init() {
-    LOG_IF(FATAL, !metaserverClient_->Init()) << "metaserverClient Init fail";
     LOG_IF(FATAL, !spaceClient_->Init()) << "spaceClient Init fail";
     LOG_IF(FATAL, !fsStorage_->Init()) << "fsStorage Init fail";
 
@@ -46,7 +45,6 @@ bool FsManager::Init() {
 void FsManager::Uninit() {
     fsStorage_->Uninit();
     spaceClient_->Uninit();
-    metaserverClient_->Uninit();
 }
 
 FSStatusCode FsManager::CreateFs(const std::string& fsName,
@@ -76,7 +74,7 @@ FSStatusCode FsManager::CreateFs(const std::string& fsName,
     }
 
     if (!skipCreateNewFs) {
-        uint64_t fsId = fsStorage_->NextFsId();
+        uint32_t fsId = fsStorage_->NextFsId();
         if (fsId == INVALID_FS_ID) {
             LOG(ERROR) << "Generator fs id failed, fsName = " << fsName;
             return FSStatusCode::INTERNAL_ERROR;
@@ -101,22 +99,16 @@ FSStatusCode FsManager::CreateFs(const std::string& fsName,
 
     // create partition
     auto ret = FSStatusCode::OK;
-    curvefs::mds::topology::CreatePartitionRequest request;
-    curvefs::mds::topology::CreatePartitionResponse response;
-    request.set_fsid(wrapper.GetFsId());
-    request.set_count(1);
-    topoManager_->CreatePartition(&request, &response);
-    if (TopoStatusCode::TOPO_OK != response.statuscode() ||
-        response.partitioninfolist_size() < 1) {
+    PartitionInfo partition;
+    TopoStatusCode topoRet = topoManager_->CreatePartitionsAndGetMinPartition(
+        wrapper.GetFsId(), &partition);
+    if (TopoStatusCode::TOPO_OK != topoRet) {
         LOG(ERROR) << "CreateFs fail, create partition fail"
-                   << ", request = " << request.ShortDebugString()
-                   << ", error code = " << response.statuscode();
+                   << ", fsId = " << wrapper.GetFsId();
         ret = FSStatusCode::CREATE_PARTITION_ERROR;
     } else {
-        PartitionInfo partition = response.partitioninfolist(0);
-        // get leader
+        // get copyset members
         std::set<std::string> addrs;
-        std::string leader;
         if (TopoStatusCode::TOPO_OK != topoManager_->GetCopysetMembers(
                 partition.poolid(), partition.copysetid(), &addrs)) {
             LOG(ERROR) << "CreateFs fail, get copyset members fail,"
@@ -124,17 +116,9 @@ FSStatusCode FsManager::CreateFs(const std::string& fsName,
                        << ", copysetId = " << partition.copysetid();
             ret = FSStatusCode::UNKNOWN_ERROR;
         } else {
-            ret = metaserverClient_->GetLeader(partition.poolid(),
-                                    partition.copysetid(), addrs, &leader);
-            if (ret != FSStatusCode::OK) {
-                LOG(ERROR) << "Createfs fail, get leader fail"
-                        << ", fsName = " << fsName
-                        << ", partition = " << partition.ShortDebugString();
-            } else {
-                ret = metaserverClient_->CreateRootInode(wrapper.GetFsId(),
-                        partition.poolid(), partition.copysetid(),
-                        partition.partitionid(), uid, gid, mode, leader);
-            }
+            ret = metaserverClient_->CreateRootInode(wrapper.GetFsId(),
+                    partition.poolid(), partition.copysetid(),
+                    partition.partitionid(), uid, gid, mode, addrs);
         }
     }
     if (ret != FSStatusCode::OK && ret != FSStatusCode::INODE_EXIST) {
@@ -289,7 +273,8 @@ FSStatusCode FsManager::MountFs(const std::string& fsName,
     }
 
     // 4. If this is the first mountpoint, init space,
-    if (wrapper.IsMountPointEmpty()) {
+    if (wrapper.GetFsType() == FSType::TYPE_VOLUME &&
+        wrapper.IsMountPointEmpty()) {
         FsInfo tempFsInfo = wrapper.ProtoFsInfo();
         ret = spaceClient_->InitSpace(tempFsInfo);
         if (ret != FSStatusCode::OK) {
@@ -350,7 +335,8 @@ FSStatusCode FsManager::UmountFs(const std::string& fsName,
     }
 
     // 3. if no mount point exist, uninit space
-    if (wrapper.IsMountPointEmpty()) {
+    if (wrapper.GetFsType() == FSType::TYPE_VOLUME &&
+        wrapper.IsMountPointEmpty()) {
         ret = spaceClient_->UnInitSpace(wrapper.GetFsId());
         if (ret != FSStatusCode::OK) {
             LOG(ERROR) << "UmountFs fail, uninit space fail, fsName = "
--- a/curvefs/src/mds/metaserverclient/metaserver_client.cpp
+++ b/curvefs/src/mds/metaserverclient/metaserver_client.cpp
@@ -21,6 +21,10 @@
  */
 
 #include "curvefs/src/mds/metaserverclient/metaserver_client.h"
+#include <bthread/bthread.h>
+
+namespace curvefs {
+namespace mds {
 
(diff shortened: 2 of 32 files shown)


curvefs/src/mds/common/mds_define.h
```diff
@@ -39,6 +44,26 @@ typedef uint32_t UserIdType;
...
+    return ip + ":" + std::to_string(port) + ":" + std::to_string(idx);
+}
+
+inline bool SplitPeerId(const std::string &peerId, std::string *ip,
+                        uint32_t *port, uint32_t *idx = nullptr) {
+    std::vector<std::string> items;
+    SplitString(peerId, ":", &items);
+    if (3 == items.size()) {
+        *ip = items[0];
+        *port = std::stoul(items[1]);
```
wu-hanqing, line 58: careful about it, it can throw exceptions if the string doesn't store a valid value or value is out of range.
use https://github.com/opencurve/curve/blob/a0b92d3680c5125bb66f8ad1a6deb33624446288/src/common/string_util.h#L72 or define another `StringToUl`

curvefs/src/mds/common/mds_define.h
```diff
@@ -39,6 +44,26 @@ typedef uint32_t UserIdType;
...
+        if (idx != nullptr) {
+            *idx = std::stoul(items[2]);
```
wu-hanqing, line 60: ditto

curvefs/src/mds/common/mds_define.h
```diff
@@ -39,6 +44,26 @@ typedef uint32_t UserIdType;
 
 const uint32_t UNINTIALIZE_ID = 0u;
 
+inline std::string BuildPeerId(const std::string &ip, uint32_t port,
```
wu-hanqing, line 47: add some annotations to explain why we need build peer id, and its format

curvefs/src/mds/metaserverclient/metaserver_client.cpp
```diff
@@ -221,7 +253,7 @@ FSStatusCode MetaserverClient::CreateCopySet(uint32_t poolId,
         copyset->set_poolid(poolId);
         copyset->set_copysetid(id);
         for (auto item : addrs) {
-            copyset->add_peers()->set_address(item + std::to_string(0));
+            copyset->add_peers()->set_address(item + ":0");
```
wu-hanqing, line 256: hard code ":0" is obscure, it's better to add another `BuildPeerId` for this case

curvefs/src/mds/metaserverclient/metaserver_client.cpp
```diff
@@ -69,11 +72,18 @@ FSStatusCode MetaserverClient::CreateRootInode(uint32_t fsId, uint32_t poolId,
     request.set_mode(mode);
 
     stub.CreateRootInode(&cntl, &request, &response, nullptr);
-
+    uint32_t maxRetry = options_.rpcRetryTimes;
+    while (cntl.Failed() && maxRetry > 0) {
```
wu-hanqing, line 76: missing check response is ok, and another problem is leader may changed, you should deal with this case.

curvefs/src/mds/topology/topology.cpp
```diff
@@ -938,6 +938,17 @@ TopoStatusCode TopologyImpl::ChooseSingleMetaServerInZone(ZoneIdType zoneId,
     return TopoStatusCode::TOPO_OK;
 }
 
+uint32_t TopologyImpl::GetPartitionNumberOfFs(FsIdType fsId) {
+    ReadLockGuard rlockPartition(partitionMutex_);
+    uint32_t pNumber = 0;
+    for (auto it : partitionMap_) {
```
wu-hanqing, line 944: `for (const auto& it ...` to avoid unnecessary copy

curvefs/src/mds/topology/topology_manager.cpp
```diff
@@ -596,6 +596,37 @@ void TopologyManager::ListPool(const ListPoolRequest *request,
     }
 }
 
+TopoStatusCode TopologyManager::CreatePartition(FsIdType fsId,
```
wu-hanqing, line 599: `CreatePartitionsAndGetMinPartition`

curvefs/src/mds/fs_manager.cpp
```diff
@@ -101,29 +101,25 @@ FSStatusCode FsManager::CreateFs(const std::string& fsName,
...
-        std::string leader;
         if (TopoStatusCode::TOPO_OK != topoManager_->GetCopysetMembers(
                 partition.poolid(), partition.copysetid(), &addrs)) {
             LOG(ERROR) << "CreateFs fail, get copyset members fail,"
                        << " poolId = " << partition.poolid()
                        << ", copysetId = " << partition.copysetid();
             ret = FSStatusCode::UNKNOWN_ERROR;
         } else {
+            // get leader
+            std::string leader;
```
wu-hanqing, line 122: the leader can be changed, I think you should not let the caller know about who is the leader, just pass all replicas' addresses, and do get leader procedure in metaserverclient.

curvefs/src/mds/topology/topology_manager.cpp
```diff
@@ -596,6 +596,37 @@ void TopologyManager::ListPool(const ListPoolRequest *request,
...
+    } else {
+        LOG(WARNING) << "CreatePartition but empty response, "
+                     << "request = " << request.ShortDebugString()
+                     << "response = " << response.ShortDebugString();
+        return TopoStatusCode::TOPO_CREATE_PARTITION_FAIL;
+    }
+    return TopoStatusCode::TOPO_OK;
+}
+
 void TopologyManager::CreatePartition(const CreatePartitionRequest *request,
```
wu-hanqing, line 630: `CreatePartitions` is more suitable

curvefs/src/mds/topology/topology_manager.cpp
SeanHai: > CreatePartitionsAndGetMinPartition

fixed

curvefs/src/mds/topology/topology_manager.cpp
SeanHai: > `CreatePartitions` is more suitable

fixed

curvefs/src/mds/topology/topology.cpp
SeanHai: > `for (const auto& it ...` to avoid unnecessary copy

fixed

curvefs/src/mds/metaserverclient/metaserver_client.cpp
SeanHai: > hard code ":0" is obscure, it's better to add another `BuildPeerId` for this case

fixed

curvefs/src/mds/common/mds_define.h
SeanHai: > add some annotations to explain why we need build peer id, and its format

fixed

curvefs/src/mds/common/mds_define.h
SeanHai: > careful about it, it can throw exceptions if the string doesn't store a valid value or value is out of range.
> use
> 
> https://github.com/opencurve/curve/blob/a0b92d3680c5125bb66f8ad1a6deb33624446288/src/common/string_util.h#L72
> 
> or define another `StringToUl`

Add StringToUl.

curvefs/src/mds/common/mds_define.h
SeanHai: > ditto

fixed

curvefs/src/mds/fs_manager.cpp
SeanHai: > the leader can be changed, I think you should not let the caller know about who is the leader, just pass all replicas' addresses, and do get leader procedure in metaserverclient.

fixed

curvefs/src/mds/topology/topology_manager.cpp
```diff
@@ -644,14 +664,27 @@ void TopologyManager::CreatePartition(const CreatePartitionRequest *request,
...
                                                     copysetMemberAddr, &leader);
         if (FSStatusCode::NOT_FOUND == retcode) {
-            LOG(WARNING) << "Get leader failed when create partition.";
+            LOG(INFO) << "Get leader failed when create partition.";
             continue;
         }
 
+        // calculate partition number of this fs
+        uint32_t pNumber = topology_->GetPartitionNumberOfFs(fsId);
+        uint64_t idStart = pNumber * option_.idNumberInPartition;
```
xu-chaojie, line 679: root inode id is 1， so the first partition is start from 1

curvefs/src/mds/metaserverclient/metaserver_client.cpp
```diff
@@ -159,6 +169,15 @@ FSStatusCode MetaserverClient::GetLeader(uint32_t poolId, uint32_t copysetId,
         cntl.set_timeout_ms(options_.rpcTimeoutMs);
         CliService2_Stub stub(&channel_);
         stub.GetLeader(&cntl, &request, &response, nullptr);
+
+        uint32_t maxRetry = options_.rpcRetryTimes;
```
xu-chaojie, line 173: The logic of retry is mostly similar, so you can consider reusing it

curvefs/src/mds/topology/topology_manager.cpp
SeanHai: > root inode id is 1， so the first partition is start from 1

This is checked and dealed in metaserver

curvefs/src/mds/metaserverclient/metaserver_client.cpp
SeanHai: > The logic of retry is mostly similar, so you can consider reusing it

Improved the retry logic, but it is hard to reuse.

curvefs/src/mds/metaserverclient/metaserver_client.cpp
```diff
@@ -189,27 +227,62 @@ FSStatusCode MetaserverClient::CreatePartition(uint32_t fsId, uint32_t poolId,
...
+            if (channel_.Init(leader.c_str(), nullptr) != 0) {
+                LOG(ERROR) << "Init channel to metaserver: " << leader
+                           << " failed!";
+                return FSStatusCode::RPC_ERROR;
+            }
+        }
+        cntl.Reset();
+        cntl.set_timeout_ms(options_.rpcTimeoutMs);
+        MetaServerService_Stub stub(&channel_);
+        stub.CreatePartition(&cntl, &request, &response, nullptr);
```
wu-hanqing, line 248: almost same with previous CreateRootInode, you can follow https://github.com/opencurve/curve/blob/a0b92d3680c5125bb66f8ad1a6deb33624446288/src/tools/mds_client.cpp#L1063 to do some improvement.

curvefs/src/mds/metaserverclient/metaserver_client.cpp
SeanHai: > The logic of retry is mostly similar, so you can consider reusing it

Reuse it with template.

curvefs/src/mds/metaserverclient/metaserver_client.cpp
```diff
@@ -30,36 +34,67 @@ using curvefs::metaserver::MetaServerService_Stub;
...
-}
+FSStatusCode MetaserverClient::GetLeader(uint32_t poolId, uint32_t copysetId,
+                                         const std::set<std::string> &addrs,
+                                         std::string *leader) {
+    GetLeaderRequest2 request;
+    GetLeaderResponse2 response;
+    request.set_poolid(poolId);
+    request.set_copysetid(copysetId);
+
+    for (std::string item : addrs) {
```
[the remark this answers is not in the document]
SeanHai: > const std::string& item : addrs

fixed

curvefs/src/mds/metaserverclient/metaserver_client.cpp
SeanHai: > almost same with previous CreateRootInode, you can follow
> 
> https://github.com/opencurve/curve/blob/a0b92d3680c5125bb66f8ad1a6deb33624446288/src/tools/mds_client.cpp#L1063
> 
> to do some improvement.

fixed

SeanHai:
recheck

xu-chaojie (approved):

wu-hanqing (approved):

PR #548, opened 2021-09-23
curvefs client : add releaseDir,fsync,background sync thread & fix pjdtest problem

xu-chaojie:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/conf/curvefs_client.conf
+++ b/curvefs/conf/curvefs_client.conf
@@ -40,6 +40,8 @@ defer.close.second=600
 fuseClient.attrTimeOut=1.0
 fuseClient.entryTimeOut=1.0
 fuseClient.listDentryLimit=100
+fuseClient.flushPeriodMs=5
+fuseClient.maxNameLength=255
 
 #volume
 volume.bigFileSize=1048576
@@ -49,17 +51,20 @@ volume.fsBlockSize=4096
 #s3
 s3.blocksize=4194304
 s3.chunksize=67108864
-s3.ak=
-s3.sk=
 s3.nos_address=
 s3.snapshot_bucket_name=
-s3.loglevel=4
+s3.ak=
+s3.sk=
+# http = 0, https = 1
 s3.http_scheme=0
 s3.verify_SSL=False
 s3.max_connections=32
 s3.connect_timeout=60000
 s3.request_timeout=10000
+# Off = 0,Fatal = 1,Error = 2,Warn = 3,Info = 4,Debug = 5,Trace = 6
+s3.loglevel=4
 s3.async_thread_num=10
+# throttle
 s3.throttle.iopsTotalLimit=0
 s3.throttle.iopsReadLimit=0
 s3.throttle.iopsWriteLimit=0
--- a/curvefs/src/client/fuse_client.cpp
+++ b/curvefs/src/client/fuse_client.cpp
@@ -93,6 +93,31 @@ void FuseClient::UnInit() {
     spaceBase_ = nullptr;
 }
 
+CURVEFS_ERROR FuseClient::Run() {
+    if (isStop_.exchange(false)) {
+        flushThread_ =
+            Thread(&FuseClient::FlushInodeLoop, this);
+        LOG(INFO) << "Start fuse client flush thread ok.";
+        return CURVEFS_ERROR::OK;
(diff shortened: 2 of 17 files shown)


curvefs/src/client/fuse_client.cpp
```diff
@@ -180,6 +209,9 @@ CURVEFS_ERROR FuseClient::FuseOpOpen(fuse_req_t req, fuse_ino_t ino,
                            << ", inodeid = " << ino;
                 return CURVEFS_ERROR::INTERNAL;
             }
+            inode.set_length(0);
+            inode.set_ctime(nowTime);
+            inode.set_mtime(nowTime);
```
Wine93, line 214: should we set `access time` to current time?

curvefs/src/client/fuse_client.cpp
```diff
@@ -537,6 +595,9 @@ CURVEFS_ERROR FuseClient::FuseOpSymlink(fuse_req_t req, const char *link,
 CURVEFS_ERROR FuseClient::FuseOpLink(fuse_req_t req, fuse_ino_t ino,
     fuse_ino_t newparent, const char *newname,
     fuse_entry_param *e) {
+    if (strlen(newname) > option_.maxNameLength) {
```
Wine93, line 598: There are many same logics, i think we can define a function to do some precheck, like `max name length` and others.

curvefs/src/client/fuse_s3_client.cpp
```diff
@@ -230,9 +247,42 @@ CURVEFS_ERROR FuseS3Client::FuseOpMkNod(fuse_req_t req, fuse_ino_t parent,
...
+}
+
 CURVEFS_ERROR FuseS3Client::Truncate(Inode *inode, uint64_t length) {
     return s3Adaptor_->Truncate(inode, 0);
 }
 
+void FuseS3Client::FlushData() {
+    CURVEFS_ERROR ret = CURVEFS_ERROR::UNKNOWN;
+    do {
+        ret = s3Adaptor_->FsSync();
```
Wine93, line 283: I think we need some protect strategy to prevent run `FsSync` forever.

curvefs/src/client/fuse_client.cpp
xu-chaojie: no need

curvefs/src/client/fuse_client.cpp
xu-chaojie: Currently, there are only two line is doing precheck, maybe we can  do reusing  after adding more logical in the future.

curvefs/src/client/fuse_s3_client.cpp
xu-chaojie: Break infinite loop here will lose data.

xu-chaojie:
recheck

curvefs/conf/curvefs_client.conf
```diff
@@ -49,17 +51,21 @@ volume.fsBlockSize=4096
...
-s3.sk=
 s3.nos_address=
 s3.snapshot_bucket_name=
-s3.loglevel=4
+s3.ak=
+s3.sk=
+# http = 0, https = 1
 s3.http_scheme=0
 s3.verify_SSL=False
+s3.user_agent_conf=S3 Browser
```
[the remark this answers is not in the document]
xu-chaojie: delete it

curvefs/conf/curvefs_client.conf
```diff
@@ -49,17 +51,21 @@ volume.fsBlockSize=4096
...
 s3.blocksize=4194304
 s3.chunksize=67108864
-s3.ak=
-s3.sk=
 s3.nos_address=
 s3.snapshot_bucket_name=
-s3.loglevel=4
+s3.ak=
+s3.sk=
+# http = 0, https = 1
```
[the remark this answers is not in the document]
xu-chaojie: using http or https to connect s3,  see https://sdk.amazonaws.com/cpp/api/LATEST/_scheme_8h_source.html

xu-chaojie:
recheck

Wine93 (approved):

ilixiaocui (approved):

PR #559, opened 2021-09-27
curvefs/metaserver: fix update nlink, check partition fsid and inode id

cw123:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/proto/metaserver.proto
+++ b/curvefs/proto/metaserver.proto
@@ -40,6 +40,7 @@ enum MetaStatusCode {
     PARTITION_ALLOC_ID_FAIL = 17;
     SAVE_META_FAIL = 18;
     S3_DELETE_ERR = 19;
+    PARTITION_ID_MISSMATCH = 20;
 }
 
 // dentry interface
@@ -165,7 +166,7 @@ message S3ChunkInfo {
     required uint64 offset = 3;
     required uint64 len = 4;  // file logic length
     required uint64 size = 5; // file size in object storage
-    required bool zero = 6; // 
+    required bool zero = 6; //
 };
 
 message S3ChunkInfoList {
--- a/curvefs/src/metaserver/partition.cpp
+++ b/curvefs/src/metaserver/partition.cpp
@@ -54,24 +54,41 @@ Partition::Partition(const PartitionInfo& paritionInfo) {
 
 // dentry
 MetaStatusCode Partition::CreateDentry(const Dentry& dentry) {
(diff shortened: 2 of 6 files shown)


curvefs/src/metaserver/partition.cpp
```diff
@@ -54,24 +54,41 @@ Partition::Partition(const PartitionInfo& paritionInfo) {
...
                                     std::vector<Dentry>* dentrys,
                                     uint32_t limit) {
+    if (!IsInodeBelongs(dentry.fsid(), dentry.parentinodeid())) {
+        return MetaStatusCode::PARTITION_ID_MISSMATCH;
+    }
     return dentryManager_->ListDentry(dentry, dentrys, limit);
 }
 
 MetaStatusCode Partition::HandleRenameTx(const std::vector<Dentry>& dentrys) {
+    for (auto it : dentrys) {
```
wu-hanqing, line 87: `for (const auto& it : dentrys)`

curvefs/src/metaserver/partition.cpp
```diff
@@ -80,6 +97,11 @@ bool Partition::InsertPendingTx(const PrepareRenameTxRequest& pendingTx) {
         pendingTx.dentrys().begin(),
         pendingTx.dentrys().end()
     };
+    for (auto it : dentrys) {
```
wu-hanqing, line 100: ditto

curvefs/proto/metaserver.proto
```diff
@@ -40,6 +40,7 @@ enum MetaStatusCode {
     PARTITION_ALLOC_ID_FAIL = 17;
     SAVE_META_FAIL = 18;
     S3_DELETE_ERR = 19;
+    PARTITION_ID_MISSMATCH = 20;
```
wu-hanqing, line 43: should the client handle this error?

xu-chaojie (approved):

curvefs/proto/metaserver.proto
cw123: no special handle needed

curvefs/src/metaserver/partition.cpp
cw123: done

curvefs/src/metaserver/partition.cpp
cw123: done

wu-hanqing (approved):

PR #568, opened 2021-10-08
curvefs client : fix core dump when FuseOpDestroy

xu-chaojie:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/fuse_s3_client.cpp
+++ b/curvefs/src/client/fuse_s3_client.cpp
@@ -64,92 +64,23 @@ void FuseS3Client::UnInit() {
     FuseClient::UnInit();
 }
 
-void FuseS3Client::FuseOpInit(void *userdata, struct fuse_conn_info *conn) {
-    struct MountOption *mOpts = (struct MountOption *) userdata;
-    std::string mountPointStr =
-        (mOpts->mountPoint == nullptr) ? "" : mOpts->mountPoint;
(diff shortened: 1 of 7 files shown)


curvefs/src/client/fuse_s3_client.cpp
```diff
@@ -64,92 +64,22 @@ void FuseS3Client::UnInit() {
...
-                      << ", fsName = " << fsName;
-
-            ::curvefs::common::S3Info s3Info;
-            s3Info.set_ak(option_.s3Opt.s3AdaptrOpt.ak);
-            s3Info.set_sk(option_.s3Opt.s3AdaptrOpt.sk);
-            s3Info.set_endpoint(option_.s3Opt.s3AdaptrOpt.s3Address);
-            s3Info.set_bucketname(option_.s3Opt.s3AdaptrOpt.bucketName);
-            s3Info.set_blocksize(option_.s3Opt.blocksize);
-            s3Info.set_chunksize(option_.s3Opt.chunksize);
-            // TODO(xuchaojie) : where to get 4096?
```
wu-hanqing, line 94: TODO is been removed, but you didn't solve the problem

xu-chaojie: fsBlockSize means min allocsize, for s3, we do not need this

SeanHai (approved):

PR #569, opened 2021-10-08
curvefs/client: init glog

SeanHai:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -43,14 +43,25 @@ using ::curvefs::client::CURVEFS_ERROR;
...
 
+    if (FLAGS_log_dir.empty()) {
+        if (!conf.GetStringValue("client.common.logDir", &FLAGS_log_dir)) {
+            LOG(WARNING) << "no client.common.logDir in " << confPath
+                         << ", will log to /tmp";
+        }
+    }
+
+    // initialize logging module
+    google::InitGoogleLogging(argv0);
```
xu-chaojie, line 63: Implement the above logic in another function

SeanHai: > Implement the above logic in another function

fixed

wu-hanqing (approved):

PR #576, opened 2021-10-12
curve-tools:  add build-topology; read FlagInfo from configure; adjust CurvefsToolRpc

Cyber-SiKu:
…t CurvefsToolRpc

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


Cyber-SiKu:
recheck

curvefs/src/tools/curvefs_tool_define.h
```diff
@@ -80,4 +85,22 @@ const char kZoneNum[] = "zonenum";
...
+namespace curvefs {
+namespace tools {
+
+void SetStringFlagInfo(curve::common::Configuration* conf,
+                       google::CommandLineFlagInfo* info,
+                       const std::string& key, fLS::clstring* flag);
+
+void SetUInt64FlagInfo(curve::common::Configuration* conf,
+                       google::CommandLineFlagInfo* info,
+                       const std::string& key, uint64_t* flag);
```
wu-hanqing, line 97: use template, and curve::common::Configuration has `GetValueFatalIfFail` for almost all types

curvefs/src/tools/curvefs_tool.h
```diff
@@ -102,16 +140,82 @@ class CurvefsToolRpc : public CurvefsTool {
...
         return 0;
     }
 
+    virtual void AddUpdateFlagsFunc(
+        const std::function<void(curve::common::Configuration*,
+                                 google::CommandLineFlagInfo*)>& func) {
+        updateFlagsFunc_.push_back(func);
+    }
+
+    virtual void UpdateFlagsFromConf() {
```
wu-hanqing, line 154: it seems no need to be virtual for these two functions.

curvefs/src/tools/curvefs_tool.h
```diff
@@ -92,8 +107,31 @@ class CurvefsToolRpc : public CurvefsTool {
...
+            if (channel_->Init(host.c_str(), nullptr) != 0) {
+                std::cerr << "Fail init channel to host: " << host << std::endl;
+                continue;
+            }
+            // if service_stub_func_ does not assign a value
+            // it will crash in there
+            service_stub_func_(controller_.get(), request_.get(),
+                               response_.get());
+            if (AfterSendRequestToService(host) == true) {
+                return true;
```
wu-hanqing, line 129: if RPC succeed, it's better to save current working host.
for example, another tool needs to communicate with mds in different steps, the first time you should try all MDS until success, but next time, you can directly send RPC to current working one.

curvefs/src/tools/topology/curvefs_build_topology_tool.cpp
```diff
@@ -0,0 +1,714 @@
...
+DECLARE_string(confPath);
+DECLARE_string(op);
+
+using ::curve::common::SplitString;
+
+namespace curvefs {
+namespace mds {
+namespace topology {
+
+void UpdateFlagsFromConf(curve::common::Configuration* conf) {
```
wu-hanqing, line 37: can this union with `CurvefsToolRpc::UpdateFlagsFromConf`?

curvefs/src/tools/curvefs_tool.h
```diff
@@ -102,16 +140,82 @@ class CurvefsToolRpc : public CurvefsTool {
...
+        std::string confPath = FLAGS_confPath.c_str();
+        curve::common::Configuration conf;
+        conf.SetConfigPath(confPath);
+        if (!conf.LoadConfig()) {
+            std::cerr << "load configure file " << confPath << " failed!"
+                      << std::endl;
+        }
+        google::CommandLineFlagInfo info;
+
+        for (auto i : updateFlagsFunc_) {
```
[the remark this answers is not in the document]
Cyber-SiKu: > for (auto& f :

fix

curvefs/src/tools/curvefs_tool.h
Cyber-SiKu: > it seems no need to be virtual for these two functions.

UpdateFlagsFromConf sets as a virtual function to prevent some subclasses from needing to manually manage FlagsInfo by themselves. It is indeed not necessary.

curvefs/src/tools/curvefs_tool.h
Cyber-SiKu: > if RPC succeed, it's better to save current working host. for example, another tool needs to communicate with mds in different steps, the first time you should try all MDS until success, but next time, you can directly send RPC to current working one.

It's really a good idea, but the command will over after running, and loop input is not accepted.

curvefs/src/tools/topology/curvefs_build_topology_tool.cpp
Cyber-SiKu: > can this union with `CurvefsToolRpc::UpdateFlagsFromConf`?

for less work, it inherited from curvefs::tools::CurvefsTool.

Cyber-SiKu:
recheck

curvefs/src/tools/topology/curvefs_build_topology_tool.h
```diff
@@ -101,6 +91,29 @@ class CurvefsTopologyTool {
 
     int TryAnotherMdsAddress();
 
+    void PrintHelp() override {
```
SeanHai, line 94: If needn't, delete it

Cyber-SiKu: > If needn't, delete it

this function usde for -example to get help hiw to use this command

Cyber-SiKu:
recheck

Cyber-SiKu:
recheck

PR #579, opened 2021-10-13
curvefs/tools: delete register metaserver in tools

SeanHai:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/devops/filter_plugins/filter.py
+++ b/curvefs/devops/filter_plugins/filter.py
@@ -101,7 +101,7 @@ def smart_topology(metaserver_groups, hostvars, nzone, ncopyset):
         idx += 1
         servers.append(server)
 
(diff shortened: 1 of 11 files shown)


curvefs/devops/filter_plugins/filter.py
```diff
@@ -69,22 +69,6 @@ def smart_topology(metaserver_groups, hostvars, nzone, ncopyset):
         'zonenum': nzone
     })
 
-    # metaservers
-    idx = 1
-    metaservers = []
```
Wine93, line 74: The `server` list can't generate when you delete the `metaserver` list, maybe you need delete the metaserver fileld of topology only.

SeanHai: > The `server` list can't generate when you delete the `metaserver` list, maybe you need delete the metaserver fileld of topology only.

fixed

PR #584, opened 2021-10-15
curvefs client : use lru cache for inode and dentry cache & fix umount lost flush inode

xu-chaojie:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


src/common/lru_cache.h
```diff
@@ -0,0 +1,372 @@
...
+};
+
+template<class T>
+struct CacheTraits {
+    static uint64_t CountBytes(const T &v) {
+        return sizeof(T);
+    }
+};
+
+template<>
```
ilixiaocui, line 82: template empty?

PR #585, opened 2021-10-15
curvefs/client: fix read data is inconsistent

baijiaruo:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -726,7 +726,7 @@ void ChunkCacheManager::ReadByWriteCache(uint64_t chunkPos, uint64_t readLen,
                     ------           DataCache
             */
             if (chunkPos + readLen <= dcChunkPos + dcLen) {
-                memcpy(dataBuf + request.len, cacheData,
+                memcpy(dataBuf + request.len + dataBufOffset, cacheData,
```
xu-chaojie, line 729: add unit test for this situation

baijiaruo:
recheck

xu-chaojie (approved):

PR #587, opened 2021-10-18
 curvefs/client: fix datacache flush get inode failed cause core dump

baijiaruo:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/s3/client_s3_cache_manager.cpp
+++ b/curvefs/src/client/s3/client_s3_cache_manager.cpp
@@ -1012,6 +1012,27 @@ void ChunkCacheManager::ReleaseCache(S3ClientAdaptorImpl* s3ClientAdaptor) {
     dataRCacheMap_.clear();
 }
 
+void ChunkCacheManager::ReleaseWriteDataCache(uint64_t key,
+     uint64_t dataCacheLen) {
+    rwLockWrite_.WRLock();
+    if (dataWCacheMap_.count(key)) {
+        dataWCacheMap_.erase(key);
+        rwLockWrite_.Unlock();
+        s3ClientAdaptor_->GetFsCacheManager()->DataCacheNumFetchSub(1);
+        LOG(INFO) << "chunk flush DataCacheByteDec1 len:"
+                  << dataCacheLen;
+        s3ClientAdaptor_->GetFsCacheManager()->
+            DataCacheByteDec(dataCacheLen);
+        if (!s3ClientAdaptor_->GetFsCacheManager()
+            ->WriteCacheIsFull()) {
+            LOG(INFO) << "write cache is not full, signal wait.";
+            s3ClientAdaptor_->GetFsCacheManager()->FlushSignal();
+        }
+    } else {
+        rwLockWrite_.Unlock();
+    }
+}
+
 CURVEFS_ERROR ChunkCacheManager::Flush(uint64_t inodeId, bool force) {
     std::map<uint64_t, DataCachePtr> tmp;
     curve::common::LockGuard lg(flushMtx_);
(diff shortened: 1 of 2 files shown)


curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1012,6 +1012,24 @@ void ChunkCacheManager::ReleaseCache(S3ClientAdaptorImpl* s3ClientAdaptor) {
...
+        dataWCacheMap_.erase(key);
+        s3ClientAdaptor_->GetFsCacheManager()->DataCacheNumFetchSub(1);
+        LOG(INFO) << "chunk flush DataCacheByteDec1 len:"
+                  << dataCacheLen;
+        s3ClientAdaptor_->GetFsCacheManager()->
+            DataCacheByteDec(dataCacheLen);
+        if (!s3ClientAdaptor_->GetFsCacheManager()
+            ->WriteCacheIsFull()) {
+            LOG(INFO) << "write cache is not full, signal wait.";
+            s3ClientAdaptor_->GetFsCacheManager()->FlushSignal();
```
wu-hanqing, line 1028: it seems these code are not necessary protected by lock

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1227,15 +1234,14 @@ CURVEFS_ERROR DataCache::Flush(uint64_t inodeId, bool force) {
...
-            dirty_.exchange(false, std::memory_order_acq_rel);
             ret = s3ClientAdaptor_->AllocS3ChunkId(fsId, &chunkId);
             if (ret != FSStatusCode::OK) {
                 LOG(ERROR) << "alloc s3 chunkid fail. ret:" << ret;
-                delete data;
                 return CURVEFS_ERROR::INTERNAL;
             }
+            data = new char[len_];
+            memcpy(data, data_, len_);
+            dirty_.exchange(false, std::memory_order_acq_rel);
```
wu-hanqing, line 1244: dirty_.store is enough

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1259,6 +1265,8 @@ CURVEFS_ERROR DataCache::Flush(uint64_t inodeId, bool force) {
             }
             if (ret < 0) {
                 LOG(ERROR) << "write object fail. object: " << objectName;
+                delete data;
+                dirty_.exchange(true, std::memory_order_acq_rel);
```
wu-hanqing, line 1269: ditto

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1277,6 +1285,7 @@ CURVEFS_ERROR DataCache::Flush(uint64_t inodeId, bool force) {
                     inodeId, inodeWrapper);
             if (ret != CURVEFS_ERROR::OK) {
                 LOG(ERROR) << "get inode fail, ret:" << ret;
+                dirty_.exchange(true, std::memory_order_acq_rel);
```
wu-hanqing, line 1288: ditto

curvefs/src/client/s3/client_s3_cache_manager.cpp
wu-hanqing: `len_` in line 1219, and `createTime_` in line 1221 are not within lock

curvefs/src/client/s3/client_s3_cache_manager.cpp
baijiaruo: fixed

PR #591, opened 2021-10-19
curvefs/client: make io smooth

wuhongsong:
make io smooth after mem cache will be full.


wuhongsong:
recheck

curvefs/src/client/s3/client_s3_adaptor.cpp
```diff
@@ -84,9 +87,22 @@ int S3ClientAdaptorImpl::Write(uint64_t inodeId, uint64_t offset,
...
+        // upload to s3 derectly or cache disk full
+        if (!enableDiskCache_ || (enableDiskCache_ &&
+             diskCacheManagerImpl_->IsDiskCacheFull())) {
+            uint32_t exponent = pow(2, (exceedRatio)/10);
+            bthread_usleep(throttleBaseSleepUs_*exceedRatio*exponent);
+            LOG(INFO) << "write cache is nearfull and use ratio is: "
+                      << memCacheRatio
+                      << ", exponent is: " << exponent;
+        }
+        // if localdisk(not full)
```
ilixiaocui, line 102: why not handle?

curvefs/src/client/s3/client_s3_adaptor.cpp
```diff
@@ -84,9 +87,21 @@ int S3ClientAdaptorImpl::Write(uint64_t inodeId, uint64_t offset,
...
+        if (!enableDiskCache_ || (enableDiskCache_ &&
+             diskCacheManagerImpl_->IsDiskCacheFull())) {
+            uint32_t exponent = pow(2, (exceedRatio)/10);
+            bthread_usleep(throttleBaseSleepUs_*exceedRatio*exponent);
+            LOG(INFO) << "write cache is nearfull and use ratio is: "
+                      << memCacheRatio
+                      << ", exponent is: " << exponent;
+        }
+    }
     if (fsCacheManager_->WriteCacheIsFull()) {
```
baijiaruo, line 103: I think this code should put in front of line 92

PR #628, opened 2021-11-03
curve-fuse: enable splice

### LINKED ISSUE #866: enable splice to reduce memory copy between user space and kernel space

wu-hanqing:

Issue Number: close #866

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/curve_fuse_op.cpp
+++ b/curvefs/src/client/curve_fuse_op.cpp
@@ -46,6 +46,25 @@ static FuseClientOption *fuseClientOption = nullptr;
 
 DECLARE_int32(v);
 
+namespace {
+
+void EnableSplice(struct fuse_conn_info* conn) {
+    if (conn->capable & FUSE_CAP_SPLICE_MOVE) {
+        conn->want |= FUSE_CAP_SPLICE_MOVE;
+        LOG(INFO) << "FUSE_CAP_SPLICE_MOVE enabled";
+    }
+    if (conn->capable & FUSE_CAP_SPLICE_READ) {
+        conn->want |= FUSE_CAP_SPLICE_READ;
+        LOG(INFO) << "FUSE_CAP_SPLICE_READ enabled";
+    }
+    if (conn->capable & FUSE_CAP_SPLICE_WRITE) {
+        conn->want |= FUSE_CAP_SPLICE_WRITE;
+        LOG(INFO) << "FUSE_CAP_SPLICE_WRITE enabled";
+    }
+}
+
+}  // namespace
+
 int InitGlog(const char *confPath, const char *argv0) {
     Configuration conf;
     conf.SetConfigPath(confPath);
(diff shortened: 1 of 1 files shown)


curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -46,6 +46,21 @@ static FuseClientOption *fuseClientOption = nullptr;
 
 DECLARE_int32(v);
 
+void EnableSplice(struct fuse_conn_info* conn) {
```
ilixiaocui, line 49: Splicing has requirements for the kernel version. You need add enable/disable flag and remarks

curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -116,6 +131,8 @@ void FuseOpInit(void *userdata, struct fuse_conn_info *conn) {
     if (ret != CURVEFS_ERROR::OK) {
         LOG(FATAL) << "FuseOpInit failed, ret = " << ret;
     }
+
+    EnableSplice(conn);
```
ilixiaocui, line 135: EnableSplice can be set in g_ClientInstance->FuseOpInit()

curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -199,15 +216,22 @@ void FuseOpOpen(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi) {
...
         return;
     }
-    fuse_reply_buf(req, buffer.get(), rSize);
+
+    struct fuse_bufvec bufvec;
+    bufvec.count = 1;
+    bufvec.off = 0;
+    bufvec.idx = 0;
+    bufvec.buf[0].size = rSize;
+    bufvec.buf[0].mem = buffer.get();
```
ilixiaocui, line 232: Here data in buffer is in user space, should we use bufvec instead of std::unique_ptr<char[]> buffer(new char[size]);?

curvefs/src/client/curve_fuse_op.cpp
wu-hanqing: `capable` in fuse_conn_info indicates capability flags that the kernel supports, and below if conditions are used to do the right thing, like, if kernel support `FUSE_CAP_SPLICE_MOVE` then I want this feature.

curvefs/src/client/curve_fuse_op.cpp
wu-hanqing: it ok to put it in here

curvefs/src/client/curve_fuse_op.cpp
wu-hanqing: no, they are two different things. bufvec is a vector of fuse_buf, and fuse_buf has a pointer `mem` that points the data that we want to send to kernel space. so, we have to allocate a buffer to store read data and make `mem` points to it.

wu-hanqing:
recheck

ilixiaocui (approved):

SeanHai (approved):

PR #629, opened 2021-11-03
makefile: support make image

Wine93:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/docker/Dockerfile
+++ b/curvefs/docker/Dockerfile
@@ -1,0 +1,7 @@
+FROM opencurvedocker/curve-base:debian9
+RUN mkdir -p /usr/local/curvefs
+COPY curvefs /usr/local/curvefs
+COPY entrypoint.sh /
+COPY curvefs/tools/sbin/curvefs_tool /usr/bin
+RUN chmod a+x /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
--- a/curvefs/docker/entrypoint.sh
+++ b/curvefs/docker/entrypoint.sh
@@ -1,0 +1,119 @@
+#!/usr/bin/env bash
+
+# Copyright (C) 2021 Jingli Chen (Wine93), NetEase Inc.
+
+############################  GLOBAL VARIABLES
+g_role=""
+g_args=""
+g_prefix=""
+g_binary=""
+g_start_args=""
+
+############################  BASIC FUNCTIONS
+function msg() {
+    printf '%b' "$1" >&2
+}
+
+function success() {
+    msg "\33[32m[✔]\33[0m ${1}${2}"
+}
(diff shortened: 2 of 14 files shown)


curvefs/docker/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM opencurvedocker/curve-base:debian9
+RUN mkdir -p /usr/local/curvefs
+COPY curvefs /usr/local/curvefs
+COPY entrypoint.sh /
+COPY curvefs/tools/sbin/curvefs_tool /usr/bin
+RUN chmod a+x /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
+CMD [""]
```
wu-hanqing, line 8: CMD may be unnecessary

curvefs/util/image.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+# tmpl.sh = /usr/local/metaserver.conf /tmp/metaserver.con
```
wu-hanqing, line 3: missing `f` in the end?

curvefs/docker/entrypoint.sh
```diff
@@ -0,0 +1,119 @@
...
+            g_start_args="--confPath $conf_path"
+            ;;
+        metaserver)
+            g_binary="$g_prefix/sbin/curvefs_metaserver_main"
+            g_start_args="--confPath $conf_path"
+            ;;
+        client)
+            g_binary="$g_prefix/sbin/curve-fuse"
+            g_start_args="--confPath $conf_path"
+            ;;
```
wu-hanqing, line 89: I think curve-fuse should not start within a docker container, it's mainly used to mount a filesystem in hosts and provide persist storage for other docker containers

curvefs/util/image.sh
Wine93: > missing `f` in the end?

fixed.

curvefs/docker/Dockerfile
Wine93: > CMD may be unnecessary

fixed.

curvefs/docker/entrypoint.sh
Wine93: > I think curve-fuse should not start within a docker container, it's mainly used to mount a filesystem in hosts and provide persist storage for other docker containers

yep, we support multi-way of mount, docker is one of them, actually it will mount host path if we use docker now.

SeanHai (approved):

Wine93:
recheck

Wine93:
recheck

wu-hanqing (approved):

PR #655, opened 2021-11-10
change log level of get inode

xu-chaojie:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1458,7 +1458,7 @@ CURVEFS_ERROR DataCache::Flush(uint64_t inodeId, bool force) {
                 s3ClientAdaptor_->GetInodeCacheManager()->GetInode(
                     inodeId, inodeWrapper);
             if (ret != CURVEFS_ERROR::OK) {
-                LOG(ERROR) << "get inode fail, ret:" << ret;
+                LOG(WARNING) << "get inode fail, ret:" << ret;
```
ilixiaocui, line 1461: ret need more readable

wu-hanqing (approved):

ilixiaocui (approved):

PR #656, opened 2021-11-10
curvefs/client: aync upload write disk file

wuhongsong:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/s3/disk_cache_write.cpp
+++ b/curvefs/src/client/s3/disk_cache_write.cpp
@@ -26,6 +26,8 @@
 #include <errno.h>
 #include <dirent.h>
 
+#include <vector>
+
 #include "curvefs/src/client/s3/disk_cache_write.h"
 
 namespace curvefs {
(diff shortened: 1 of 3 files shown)


curvefs/src/client/s3/disk_cache_write.cpp
```diff
@@ -100,13 +102,18 @@ int DiskCacheWrite::UploadFile(const std::string name) {
...
         return -1;
     }
-    ret = client_->Upload(name, buffer, fileSize);
-    if (ret < 0) {
-        LOG(ERROR) << "upload object fail. object: " << name;
-        posixWrapper_->free(buffer);
-        posixWrapper_->close(fd);
-        return -1;
-    }
+    PutObjectAsyncCallBack cb =
```
ilixiaocui, line 105: This definition in many places，make it public

curvefs/src/client/s3/disk_cache_write.cpp
```diff
@@ -139,7 +146,7 @@ int DiskCacheWrite::AsyncUploadFunc() {
         std::list<std::string>::iterator iter;
         int ret;
         for (iter = toUpload.begin(); iter != toUpload.end(); iter++) {
-            ret = UploadAndRemove(*iter);
+            ret = UploadFile(*iter);
```
ilixiaocui, line 149: Why not remove? When will it be deleted？

wuhongsong: > Why not remove? When will it be deleted？

delete in aync callback.

curvefs/src/client/s3/disk_cache_write.cpp
```diff
@@ -100,13 +102,18 @@ int DiskCacheWrite::UploadFile(const std::string name) {
...
-    ret = client_->Upload(name, buffer, fileSize);
-    if (ret < 0) {
-        LOG(ERROR) << "upload object fail. object: " << name;
-        posixWrapper_->free(buffer);
-        posixWrapper_->close(fd);
-        return -1;
-    }
+    PutObjectAsyncCallBack cb =
+        [&](const std::shared_ptr<PutObjectAsyncContext> &context) {
+            RemoveFile(context->key);
```
wu-hanqing, line 107: `RemoveFiles`'s result is ignored?

wu-hanqing (approved):

ilixiaocui (approved):

curvefs/src/client/s3/disk_cache_write.cpp
wuhongsong: The callback content is not the same

Issue #684, opened 2021-11-16
curvefs:  read lru cache coredump


wu-hanqing:
**Describe the bug (描述bug)**

curve-fuse core dump when testing

![image](https://user-images.githubusercontent.com/10327590/141993504-77a7db61-18c0-4be9-8893-0062e8b64f09.png)

**To Reproduce (复现方法)**

vdbench profile
```bash
fsd=fsd1,anchor=/mnt/vdbench3,depth=1,width=10,files=10,sizes=(100m,0),shared=yes,openflags=o_direct
fwd=fwd1,fsd=fsd1,threads=10,xfersize=(512,20,4k,20,64k,20,512k,20,1024k,20),fileio=random,fileselect=random,rdpct=50
rd=rd1,fwd=fwd*,fwdrate=max,format=restart,elapsed=7200,interval=1
```

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS: debian buster
Compiler: gcc (Debian 8.3.0-6) 8.3.0
fs: 087c8a3dd8c983ce3e7cb2fa6ac301af4a20fe3f

**Additional context/screenshots (更多上下文/截图)**

![image](https://user-images.githubusercontent.com/10327590/141993867-1b7b453c-d698-426e-a8c7-51d7ccdb960e.png)

wu-hanqing:
solved https://github.com/opencurve/curve/pull/690

PR #685, opened 2021-11-16
curvefs/client: temporary disable read lru cache

wu-hanqing:

Issue Number: #684 

Problem Summary: Currently, the solution is not so clear, so just disable it now in order not to affect other tests

What's Changed:

*temporarily disable read lru cache*

*read performance may lower down, but it was not obvious in the test*

### Check List


baijiaruo (approved):

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1162,6 +1162,8 @@ ChunkCacheManager::CreateWriteDataCache(S3ClientAdaptorImpl *s3ClientAdaptor,
 }
 
 void ChunkCacheManager::AddReadDataCache(DataCachePtr dataCache) {
+    return;
```
ilixiaocui, line 1165: add TODO here

YunhuiChen:
recheck

PR #690, opened 2021-11-18
curvefs/client: fix read lru bug

### LINKED ISSUE #684: curvefs:  read lru cache coredump

wu-hanqing:

Issue Number: close #684 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/s3/client_s3_cache_manager.cpp
+++ b/curvefs/src/client/s3/client_s3_cache_manager.cpp
@@ -67,43 +67,64 @@ void FsCacheManager::ReleaseFileCacheManager(uint64_t inodeId) {
     return;
 }
 
-std::pair<std::list<DataCachePtr>::iterator, WeakDataCachePtr>
-FsCacheManager::Set(DataCachePtr dataCache) {
-    WriteLockGuard writeLockGuard(rwLockLru_);
+std::list<DataCachePtr>::iterator FsCacheManager::Set(DataCachePtr dataCache) {
+    std::lock_guard<std::mutex> lk(lruMtx_);
 
     VLOG(3) << "lru current byte:" << lruByte_
             << ",lru max byte:" << readCacheMaxByte_;
 
     // trim cache without consider dataCache's size, because its size is
     // expected to be very smaller than `readCacheMaxByte_`
-    while (lruByte_ >= readCacheMaxByte_) {
-        auto &trimDataCache = lruReadDataCacheList_.back();
-        VLOG(3) << "lru release data cache, size: " << trimDataCache->GetLen();
-        lruByte_ -= trimDataCache->GetLen();
-        trimDataCache->Release();
-        lruReadDataCacheList_.pop_back();
-    }
-    lruByte_ += dataCache->GetLen();
+    if (lruByte_ >= readCacheMaxByte_) {
+        uint64_t retiredBytes = 0;
+        auto iter = lruReadDataCacheList_.end();
 
-    WeakDataCachePtr weak = dataCache;
+        while (lruByte_ >= readCacheMaxByte_) {
+            --iter;
+            auto& trim = *iter;
+            trim->SetReadCacheState(false);
+            lruByte_ -= trim->GetLen();
+            retiredBytes += trim->GetLen();
+        }
+
+        std::list<DataCachePtr> retired;
+        retired.splice(retired.end(), lruReadDataCacheList_, iter,
+                       lruReadDataCacheList_.end());
+
+        VLOG(3) << "lru release " << retiredBytes << " bytes, retired "
+                << retired.size() << " data cache";
+
+        releaseReadCache_.Release(&retired);
+    }
+
+    lruByte_ += dataCache->GetLen();
+    dataCache->SetReadCacheState(true);
     lruReadDataCacheList_.push_front(std::move(dataCache));
-    return {lruReadDataCacheList_.begin(), weak};
+
+    return lruReadDataCacheList_.begin();
 }
 
 void FsCacheManager::Get(std::list<DataCachePtr>::iterator iter) {
-    WriteLockGuard writeLockGuard(rwLockLru_);
+    std::lock_guard<std::mutex> lk(lruMtx_);
+
+    if (!(*iter)->InReadCache()) {
+        return;
+    }
 
     lruReadDataCacheList_.splice(lruReadDataCacheList_.begin(),
                                  lruReadDataCacheList_, iter);
-    return;
 }
 
 void FsCacheManager::Delete(std::list<DataCachePtr>::iterator iter) {
-    WriteLockGuard writeLockGuard(rwLockLru_);
+    std::lock_guard<std::mutex> lk(lruMtx_);
 
+    if (!(*iter)->InReadCache()) {
+        return;
+    }
+
+    (*iter)->SetReadCacheState(false);
     lruByte_ -= (*iter)->GetLen();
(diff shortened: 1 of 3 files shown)


curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -67,44 +67,64 @@ void FsCacheManager::ReleaseFileCacheManager(uint64_t inodeId) {
...
-        lruByte_ -= trimDataCache->GetLen();
-        trimDataCache->Release();
-        lruReadDataCacheList_.pop_back();
+    if (lruByte_ >= readCacheMaxByte_) {
+        uint64_t retiredBytes = 0;
+        auto iter = lruReadDataCacheList_.end();
+
+        while (lruByte_ >= readCacheMaxByte_) {
+            --iter;
+            auto& trim = *iter;
```
ilixiaocui, line 84: Why not use iter directly？

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -67,44 +67,64 @@ void FsCacheManager::ReleaseFileCacheManager(uint64_t inodeId) {
...
     // trim cache without consider dataCache's size, because its size is
     // expected to be very smaller than `readCacheMaxByte_`
-    while (lruByte_ >= readCacheMaxByte_) {
-        auto& trimDataCache = lruReadDataCacheList_.back();
-        VLOG(3) << "lru release data cache, size: "
-        << trimDataCache->GetLen();
-        lruByte_ -= trimDataCache->GetLen();
-        trimDataCache->Release();
-        lruReadDataCacheList_.pop_back();
+    if (lruByte_ >= readCacheMaxByte_) {
```
ilixiaocui, line 78: if is redundant？
{
      uint64_t retiredBytes = 0;
      auto iter = lruReadDataCacheList_.end();
      while(){}
}

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1584,5 +1588,51 @@ void DataCache::UpdateInodeChunkInfo(S3ChunkInfoList *s3ChunkInfoList,
...
+}
+
+void FsCacheManager::ReadCacheReleaseExecutor::Stop() {
+    std::lock_guard<std::mutex> lk(mtx_);
+    running_ = false;
+    cond_.notify_one();
+}
+
+FsCacheManager::ReadCacheReleaseExecutor::~ReadCacheReleaseExecutor() {
+    Stop();
```
ilixiaocui, line 1626: print log

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1405,9 +1408,10 @@ void DataCache::Write(uint64_t chunkPos, uint64_t len, const char *data,
 }
 
 void DataCache::Release() {
-    chunkCacheManager_->ReleaseReadDataCache(chunkPos_);
-
-    return;
+    auto chunkCacheManager = chunkCacheManager_.lock();
```
ilixiaocui, line 1411: do not need. 
Release()  is now only called by ReadCacheReleaseExecutor. We can get chunkCacheManager in FsCacheManager.
Datacahe does not require reverse dependency on chunkCacheManager.

curvefs/src/client/s3/client_s3_cache_manager.h
```diff
@@ -91,25 +91,32 @@ struct ObjectChunkInfo {
...
+          inReadCache_(false) {
         data_ = new char[len];
         memcpy(data_, data, len);
         createTime_ = ::curve::common::TimeUtility::GetTimeofDaySec();
-        //  dirty_.exchange(true, std::memory_order_acq_rel);
     }
     DataCache(S3ClientAdaptorImpl *s3ClientAdaptor,
-              ChunkCacheManager *chunkCacheManager, uint64_t chunkPos,
-              uint64_t len)
+              const std::shared_ptr<ChunkCacheManager>& chunkCacheManager,
```
baijiaruo, line 109: & is near chunkCacheManager，The format needs to be uniform

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1166,43 +1175,38 @@ void ChunkCacheManager::AddReadDataCache(DataCachePtr dataCache) {
...
         auto iter = dataRCacheMap_.find(key);
-        std::list<DataCachePtr>::iterator dcpIter = iter->second.first;
+        std::list<DataCachePtr>::iterator dcpIter = iter->second;
         s3ClientAdaptor_->GetFsCacheManager()->Delete(dcpIter);
         dataRCacheMap_.erase(iter);
     }
 
-    dataRCacheMap_.emplace(
-        chunkPos, s3ClientAdaptor_->GetFsCacheManager()->Set(dataCache));
+    std::list<DataCachePtr>::iterator newIter =
```
baijiaruo, line 1203: Don't see the difference

curvefs/src/client/s3/client_s3_cache_manager.cpp
wu-hanqing: it's more readable and concise than
```cpp
(*iter)->SetReadCacheState(false);
lruByte_ -= (*iter)->GetLen();
retiredBytes += (*iter)->GetLen();
```

curvefs/src/client/s3/client_s3_cache_manager.cpp
wu-hanqing: not really, if do what you say, another if should be added to check whether I have to cut some data chches from lruReadDataCacheList_ or not, because while loop does not pop any data cache.

curvefs/src/client/s3/client_s3_cache_manager.cpp
wu-hanqing: fix

curvefs/src/client/s3/client_s3_cache_manager.cpp
wu-hanqing: this change has been removed because the case that can cause chunkCacheManager_ expired has been implicitly fix by @baijiaruo

curvefs/src/client/s3/client_s3_cache_manager.h
wu-hanqing: fix

curvefs/src/client/s3/client_s3_cache_manager.cpp
wu-hanqing: fix

ilixiaocui (approved):

baijiaruo (approved):

PR #736, opened 2021-11-26
curvefs/rpcclient: fix io hang when mds host power down

### LINKED ISSUE #732: shutdown one etcd\mds\metaserver,  io hang
**Describe the bug (描述bug)**

shutdown one etcd\mds\metaserver,  io hang
**To Reproduce (复现方法)**
1、write vdbench
2、reboot one host, have (etcd \ mds \ metaserver)
3、io hange 4-5 minutes

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
curve-mds:
curve-chunkserver:
curve-snapshotcloneserver:
curve-sdk:
nebd:
curve-nbd:

image
image

wu-hanqing:

Issue Number: close #732 

Problem Summary: https://github.com/opencurve/curve/issues/732#issuecomment-979659922

What's Changed: lower down some configurations

How it Works:

### Check List

### DIFF
--- a/curvefs/conf/client.conf
+++ b/curvefs/conf/client.conf
@@ -4,14 +4,13 @@ mdsOpt.mdsMaxRetryMS=16000
 # The maximum timeout of RPC communicating with MDS.
 # The timeout of exponential backoff cannot exceed this value
 mdsOpt.rpcRetryOpt.maxRPCTimeoutMS=2000
-# TODO(chengyi): after curvefs-tools support create fs, this value can lower down to 500
 # RPC timeout for once communication with MDS
-mdsOpt.rpcRetryOpt.rpcTimeoutMs=5000
+mdsOpt.rpcRetryOpt.rpcTimeoutMs=500
 # RPC with mds needs to sleep for a period of time before each retry
 mdsOpt.rpcRetryOpt.rpcRetryIntervalUS=50000
 # Switch if the number of consecutive retries on the current MDS exceeds the limit.
 # The number of failures includes the number of timeout retries
-mdsOpt.rpcRetryOpt.maxFailedTimesBeforeChangeAddr=3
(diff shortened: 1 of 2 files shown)


curvefs/src/client/rpcclient/mds_client.cpp
```diff
@@ -94,7 +94,10 @@ FSStatusCode MdsClientImpl::CreateFsS3(const std::string &fsName,
 
         return ret;
     };
-    return ReturnError(rpcexcutor_.DoRPCTask(task, mdsOpt_.mdsMaxRetryMS));
+
+    // manually set to a large timeout, because create fs may very slow
```
ilixiaocui, line 98: // TODO
createfs will remove in the feature.

curvefs/conf/client.conf
```diff
@@ -57,7 +56,7 @@ extentManager.preAllocSize=65536
 
 #### brpc
 # close socket after defer.close.second
-rpc.defer.close.second=600
+rpc.defer.close.second=1
```
ilixiaocui, line 59: defer set to a long time for reuse, Will this be disconnected under abnormal conditions?

curvefs/src/client/rpcclient/mds_client.cpp
wu-hanqing: line 78 has a TODO about this

curvefs/conf/client.conf
wu-hanqing: it's not, there is an atomic variable that stores currentWorkingMds, and only when RPC succeeded atomic get updated.

so, next `AllocS3ChunkId` will send RPC to power down MDS firstly, because no one succeeds sending RPC to MDS and currentWorkingMds still stores the power down MDS

SeanHai (approved):

ilixiaocui (approved):

PR #761, opened 2021-12-02
glog: now we can logging normally in child process even if async enabled

### LINKED ISSUE #743: copyset raft snapshot hang
**Describe the bug (描述bug)**

image

image

snapshot_status stays SAVING for several hours and exists lots of metaserver processes

**To Reproduce (复现方法)**

uncertain

**Expected behavior (期望行为)**

complete save raft snapshot

**Versions (各种版本)**
curve-fuse/metaserver: 38cf877ab597f2014a7c1a0e36da0e4ab15dd3fe

metaserver pid 160 
image

metaserver pid 747
image

Wine93:

Issue Number: close #743  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/thirdparties/glog/glog.patch
+++ b/thirdparties/glog/glog.patch
@@ -97,7 +97,7 @@ index 9968b96..f491f9b 100644
  #undef MUST_UNDEF_GFLAGS_DECLARE_MACROS
  #undef DECLARE_VARIABLE
 diff --git a/src/logging.cc b/src/logging.cc
-index 0c86cf6..b77a4d8 100644
+index 0c86cf6..601bdda 100644
 --- a/src/logging.cc
 +++ b/src/logging.cc
 @@ -27,6 +27,39 @@
@@ -371,6 +371,15 @@ index 0c86cf6..b77a4d8 100644
    delete sinks_;
    sinks_ = NULL;
  }
+@@ -906,7 +984,7 @@ bool LogFileObject::CreateLogfile(const string& time_pid_string) {
+   string string_filename = base_filename_+filename_extension_+
+                            time_pid_string;
+   const char* filename = string_filename.c_str();
+-  int fd = open(filename, O_WRONLY | O_CREAT | O_EXCL, FLAGS_logfile_mode);
++  int fd = open(filename, O_WRONLY | O_CREAT | O_APPEND | O_EXCL, FLAGS_logfile_mode);
+   if (fd == -1) return false;
+ #ifdef HAVE_FCNTL
+   // Mark the file close-on-exec. We don't really care if this fails
 @@ -1015,9 +1093,9 @@ void LogFileObject::Write(bool force_flush,
      } else {
        // If no base filename for logs of this severity has been set, use a
@@ -678,10 +687,10 @@ index 0c86cf6..b77a4d8 100644
  _END_GOOGLE_NAMESPACE_
 diff --git a/src/logging_async.cc b/src/logging_async.cc
 new file mode 100644
-index 0000000..1f3c8c1
+index 0000000..8d68ebe
 --- /dev/null
 +++ b/src/logging_async.cc
-@@ -0,0 +1,176 @@
+@@ -0,0 +1,206 @@
 +/*
 + *  Copyright (c) 2020 NetEase Inc.
 + *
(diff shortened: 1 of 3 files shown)


thirdparties/glog/glog.patch
```diff
@@ -741,7 +742,30 @@ index 0000000..1f3c8c1
...
++    flushBuffer_.reset(new Buffer());
++    running_ = false;
++
++    if (start) {
++        Start();
++    }
++}
++
 +void AsyncLogger::Start() {
++    ResetInChildProcess(false);
```
ilixiaocui, line 767: AsyncLogger::Start() will called by child process after fork?

Wine93: > 

yep, because in the child process, there is no SyncLogger thread.

thirdparties/glog/glog.patch
```diff
@@ -741,7 +742,30 @@ index 0000000..1f3c8c1
...
++    if (!PidHasChanged()) {
++        return;
++    }
++
++    // After fork(2) and in the child process we should:
++    //   (1) reset all status for AyncLogger
++    //   (2) restart AyncLogger when necessary
++    appThreadsBlockedCount_ = 0;
++    flushCount_ = 0;
++    mutex_.unlock();
```
wu-hanqing, line 756: according to the description in here https://en.cppreference.com/w/cpp/thread/mutex/unlock, 

```
The mutex must be locked by the current thread of execution, otherwise, the behavior is undefined.
```

and, destroy a mutex locked by another thread is also UB.
https://en.cppreference.com/w/cpp/thread/mutex/%7Emutex

so, current solution may not work well.

you can check whether `pthread_mutexattr_setpshared` is helpful, or find another way.

wu-hanqing (commented):
BTW, I remember you close all fd just after fork, and what's the log destination of child process?

Wine93:
> BTW, I remember you close all fd just after fork, and what's the log destination of child process?

Actully, I only close all `socket`, the regular file will be retained.

thirdparties/glog/glog.patch
Wine93: > 

yep, it's a bad solution, i have changed a way to solve it.

thirdparties/glog/glog.patch
```diff
@@ -716,63 +725,91 @@ index 0000000..1f3c8c1
...
++      running_(false),
++      current_(0) {}
 +
 +AsyncLogger::~AsyncLogger() {
 +    if (running_) {
 +        Stop();
 +    }
 +}
 +
++// NOTE: this function only run once in child process
```
wu-hanqing, line 753: the whole thing between line 763 and line 775 isn't atomic, it's better to encapsulate them into std::call_once

thirdparties/glog/glog.patch
```diff
@@ -371,6 +371,15 @@ index 0c86cf6..b77a4d8 100644
    delete sinks_;
    sinks_ = NULL;
  }
+@@ -906,7 +984,7 @@ bool LogFileObject::CreateLogfile(const string& time_pid_string) {
+   string string_filename = base_filename_+filename_extension_+
+                            time_pid_string;
+   const char* filename = string_filename.c_str();
+-  int fd = open(filename, O_WRONLY | O_CREAT | O_EXCL, FLAGS_logfile_mode);
++  int fd = open(filename, O_WRONLY | O_CREAT | O_APPEND | O_EXCL, FLAGS_logfile_mode);
```
wu-hanqing, line 379: this line code https://github.com/google/glog/blob/503e3dec8d1fe071376befc62119a837c26612a3/src/logging.cc#L1123 may break your expectation that child process can write logs into the parent process's log file.

but, anyway, it's far more difficult to fix all cases, we can leave it behind, and fix this latter.

thirdparties/glog/glog.patch
Wine93: > 

yep, i will improve it.

thirdparties/glog/glog.patch
Wine93: > 

I had invoke `PidHasChanged()` before create log file, so the logger will never create a new file when pid changed.

Wine93:
recheck

ilixiaocui (approved):

wu-hanqing (approved):

Wine93:
recheck

aspirer:
need to push to fs and 0.1.0 branch.

PR #815, opened 2021-12-09
curvefs-mds/curvefs-metaserver: install SIGTERM handler

### LINKED ISSUE #804: curvefs-mds process doesn't exit when observe leader fail
**Describe the bug (描述bug)**

```
E 2021-12-08T06:08:35.706418+0000    33 leader_election.cpp:79] 10.182.26.16:6700 leader observe for prefix:07leader occur error, errcode: 23
I 2021-12-08T06:08:35.706470+0000    33 leader_election.cpp:86] mds is existing due to the error of leader observation
```

**To Reproduce (复现方法)**

uncertain

**Expected behavior (期望行为)**

when observe leader fail, curvefs-mds process should exit

**Versions (各种版本)**
curvefs-mds: b1f805dda17d6e85fe44ffa6b376b4d89bb7410f

wu-hanqing:

Issue Number: close #804 

Problem Summary:

Leader observation is handled in another detached thread,
and when observe leader failed, it raises a SIGTERM in the detached thread.

By default, we didn't register handler for SIGTERM, so the process should exit directly.
But, in current scenario, it seems the SIGTERM signal is ignored.

What's Changed:

To avoid this problem, register handler for SIGTERM through brpc's flags graceful_quit_on_sigterm

### Check List

### DIFF
--- a/curvefs/src/mds/mds.cpp
+++ b/curvefs/src/mds/mds.cpp
@@ -30,6 +30,10 @@
 
 #include "curvefs/src/mds/mds_service.h"
 
+namespace brpc {
+DECLARE_bool(graceful_quit_on_sigterm);
+}  // namespace brpc
+
 namespace curvefs {
 namespace mds {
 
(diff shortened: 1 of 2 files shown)


wu-hanqing:
recheck

Wine93 (approved):

cw123 (approved):

Wine93:
LGTM！

PR #834, opened 2021-12-13
curve: support block storage open multiple times

### LINKED ISSUE #856: does curve block storage support open multiple times ?

wu-hanqing:

Issue Number: close #856

What's Changed:

support block storage open multiple times
e.g., a single file/image can be mapped multiple times on different hosts by
curve-nbd with extra option `--no-exclusive'

limitations:
1. provided as a shared disk/device WITHOUT data consistent guarantee,
   so it's users' responsibility to deal with this problem.
2. each file/image can only open once on each host.

### Check List

### DIFF
--- a/curve-ansible/roles/install_package/tasks/include/install_nebd.yml
+++ b/curve-ansible/roles/install_package/tasks/include/install_nebd.yml
@@ -64,6 +64,12 @@
         local_file_path: "{{ local_nebd_package_path }}/bin/"
         file_mode: 0755
       include_tasks: copy_file_to_remote.yml
+    - name: install nebd include
+      vars:
+        remote_dir_name: "{{ curve_include_dir }}"
+        local_file_path: "{{ local_nebd_package_path }}/include/"
+        file_mode: 0644
+      include_tasks: copy_file_to_remote.yml
     - name: install nebd lib
       vars:
         remote_dir_name: "{{ curve_lib_dir }}"
--- a/include/client/libcurve.h
+++ b/include/client/libcurve.h
@@ -450,6 +450,12 @@ typedef struct UserInfo {
     }
 } UserInfo_t;
 
+struct OpenFlags {
+    bool exclusive;
+
+    OpenFlags() : exclusive(true) {}
+};
+
 class CurveClient {
  public:
     CurveClient();
@@ -474,7 +480,7 @@ class CurveClient {
      * @return 成功返回fd，失败返回-1
      */
     virtual int Open(const std::string& filename,
-                     std::string* sessionId);
+                     const OpenFlags& openflags);
 
     /**
      * 重新打开文件
@@ -484,8 +490,7 @@ class CurveClient {
      * @return 成功返回fd，失败返回-1
      */
     virtual int ReOpen(const std::string& filename,
-                       const std::string& sessionId,
-                       std::string* newSessionId);
+                       const OpenFlags& openflags);
 
     /**
      * 关闭文件
--- a/nbd/src/define.h
+++ b/nbd/src/define.h
@@ -50,14 +50,6 @@
 #include <sstream>
 #include <string>
 
-namespace std {
-
-inline std::string to_string(const std::string& val) {
-    return val;
-}
-
-}  // namespace std
-
 namespace curve {
 namespace nbd {
 
@@ -87,6 +79,8 @@ struct NBDConfig {
     bool try_netlink = false;
     // 需要映射的后端文件名称
     std::string imgname;
+    // exclusive open image or not
+    bool exclusive = true;
     // 指定需要映射的nbd设备路径
     std::string devpath;
     // force unmap even if the device is mounted
@@ -142,9 +136,16 @@ inline std::string BoolOption(const std::string& name, bool value,
     return opts;
 }
 
+inline std::string to_string(const std::string& val) {
+    return val;
+}
+
 template <typename T>
 inline std::string KeyValueOption(const std::string& optName, const T& value,
                                   const T& defaultValue, bool* firstOpt) {
+    using curve::nbd::to_string;
+    using std::to_string;
+
     std::string opts;
 
     if (value != defaultValue) {
@@ -152,7 +153,7 @@ inline std::string KeyValueOption(const std::string& optName, const T& value,
             opts += ",";
         }
 
-        opts += std::string(optName + "=") + std::to_string(value);
+        opts += std::string(optName + "=") + to_string(value);
         *firstOpt = false;
     }
 
@@ -171,6 +172,7 @@ inline std::string NBDConfig::MapOptions() const {
     opts.append(KeyValueOption("timeout", timeout, 3600, &firstOpt));
     opts.append(KeyValueOption("block-size", block_size, 4096, &firstOpt));
     opts.append(KeyValueOption("nebd-conf", nebd_conf, {}, &firstOpt));
+    opts.append(BoolOption("no-exclusive", !exclusive, &firstOpt));
 
     return opts.empty() ? "defaults" : opts;
 }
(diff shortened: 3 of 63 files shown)


wu-hanqing:
recheck

wu-hanqing:
recheck

wu-hanqing:
recheck

wu-hanqing:
recheck

curve-ansible/roles/install_package/tasks/include/install_nebd.yml
```diff
@@ -64,6 +64,12 @@
         local_file_path: "{{ local_nebd_package_path }}/bin/"
         file_mode: 0755
       include_tasks: copy_file_to_remote.yml
+    - name: install nebd include
+      vars:
+        remote_dir_name: "{{ curve_include_dir }}"
+        local_file_path: "{{ local_nebd_package_path }}/include/"
```
SeanHai, line 70: The 'include' directory doesn't exist in nebd-package.

nbd/src/define.h
```diff
@@ -142,17 +136,24 @@ inline std::string BoolOption(const std::string& name, bool value,
...
+
     std::string opts;
 
     if (value != defaultValue) {
         if (!*firstOpt) {
             opts += ",";
         }
 
-        opts += std::string(optName + "=") + std::to_string(value);
+        opts += std::string(optName + "=") + to_string(value);
```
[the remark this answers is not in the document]
wu-hanqing: because previous method is illegal, even it works.

nbd/src/define.h
```diff
@@ -171,6 +172,7 @@ inline std::string NBDConfig::MapOptions() const {
     opts.append(KeyValueOption("timeout", timeout, 3600, &firstOpt));
     opts.append(KeyValueOption("block-size", block_size, 4096, &firstOpt));
     opts.append(KeyValueOption("nebd-conf", nebd_conf, {}, &firstOpt));
+    opts.append(BoolOption("no-exclusive", !exclusive, &firstOpt));
```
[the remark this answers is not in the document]
wu-hanqing: firstly, it should be exclusive by default. another reason is current argument parser doesn't support bool option with a value, so if a bool option presents it means it's true.

an alternative is define this option as exclusive 1 or 0, but it's also strange.

nebd/src/part2/file_entity.cpp
```diff
@@ -57,18 +81,26 @@ int NebdFileEntity::Init(const NebdFileEntityOption& option) {
...
+int NebdFileEntity::Open(const OpenFlags* openflags) {
     CHECK(executor_ != nullptr) << "file entity is not inited. "
                                 << "filename: " << fileName_;
     std::unique_lock<bthread::Mutex> lock(fileStatusMtx_);
     if (status_ == NebdFileStatus::OPENED) {
-        LOG(WARNING) << "File is already opened. "
-                     << "filename: " << fileName_
-                     << "fd: " << fd_;
-        return fd_;
+        if (IsOpenFlagsExactlySame(openFlags_.get(), openflags)) {
```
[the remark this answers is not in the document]
wu-hanqing: yep, nebd-server is just a proxy, it shouldn't concerned about openflags.

nebd/src/part2/request_executor.cpp
```diff
@@ -36,6 +36,7 @@ NebdRequestExecutorFactory::GetExecutor(NebdFileType type) {
             executor = &CurveRequestExecutor::GetInstance();
             break;
         case NebdFileType::TEST:
+            LOG(ERROR) << "here";
```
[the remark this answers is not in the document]
wu-hanqing: fix

curve-ansible/roles/install_package/tasks/include/install_nebd.yml
wu-hanqing: it's been created in mk-tar.sh

https://github.com/opencurve/curve/blob/32ee88c706dacca40d2118bad41310de0e92cf93/mk-tar.sh#L457

src/client/splitor.cpp
```diff
@@ -469,5 +467,19 @@ RequestSourceInfo Splitor::CalcRequestSourceInfo(IOTracker* ioTracker,
     return {};
 }
 
+bool Splitor::NeedGetOrAllocateSegment(MetaCacheErrorType error, OpType opType,
+                                       const ChunkIDInfo& chunkInfo,
+                                       const MetaCache* metaCache) {
+    if (error == MetaCacheErrorType::CHUNKINFO_NOT_FOUND) {
+        return true;
+    } else if (error == MetaCacheErrorType::OK && !chunkInfo.chunkExist) {
+        return (opType == OpType::WRITE) ||
```
ilixiaocui, line 476: If a read-only node becomes a read-write node, we need to restart nedb? Here write chunk not exist does not getorallocate segment.

src/mds/nameserver2/curvefs.cpp
```diff
@@ -566,7 +574,7 @@ StatusCode CurveFS::IsSnapshotAllowed(const std::string &fileName) {
 
     // the client version satisfies the conditions
     std::string clientVersion;
-    bool exist = fileRecordManager_->GetFileClientVersion(
+    bool exist = fileRecordManager_->GetMinimumFileClientVersion(
```
ilixiaocui, line 577: Can it be guaranteed that min is a write node?

include/client/libcurve.h
```diff
@@ -474,7 +480,7 @@ class CurveClient {
      * @return 成功返回fd，失败返回-1
      */
     virtual int Open(const std::string& filename,
-                     std::string* sessionId);
+                     const OpenFlags& openflags);
```
ilixiaocui, line 483: Why is sessionId not needed here？

nbd/src/define.h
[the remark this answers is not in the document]
wu-hanqing: firstly, it should be exclusive by default. another reason is current argument parser doesn't support bool option with a value, so if a bool option presents it means it's true.

an alternative is define this option as exclusive 1 or 0, but it's also strange.

nbd/src/util.cpp
```diff
@@ -328,7 +334,8 @@ int check_dev_can_unmap(const NBDConfig *cfg) {
     return -EINVAL;
 }
 
-int check_size_from_file(const std::string& path, uint64_t expected_size) {
+int check_size_from_file(const std::string &path, uint64_t expected_size,
+                         bool sizeInSector = false) {
```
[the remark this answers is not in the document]
wu-hanqing: value in `/sys/block/nbdX/size` is the number of sectors of current device, so we have to multiple sector size before check.

nbd/test/nbd_tool_test.cpp
```diff
@@ -59,6 +59,7 @@ class NBDToolTest : public ::testing::Test {
 
     void TearDown() override {
         nbdThread_.join();
+        sleep(5);
```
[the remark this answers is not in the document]
wu-hanqing: fix

src/client/splitor.cpp
wu-hanqing: > If a read-only node becomes a read-write node, we need to restart nebd? 

reopen or re-map is needed, but this code is nothing about read-only

> Here write chunk not exist does not getorallocate segment.

nope, if it's a write operation this function returns true.

src/mds/nameserver2/curvefs.cpp
wu-hanqing: nope

include/client/libcurve.h
wu-hanqing: session has been removed, so it became useless.

wu-hanqing:
recheck

ilixiaocui (approved):

SeanHai (approved):

PR #898, opened 2021-12-27
curvefs/metaserver: fix s3compact crash when partition have been deleted

h0hmj:
we need erase s3compact from vector when partition is deleted

### DIFF
--- a/curvefs/src/metaserver/s3compact.h
+++ b/curvefs/src/metaserver/s3compact.h
@@ -24,10 +24,9 @@
 #define CURVEFS_SRC_METASERVER_S3COMPACT_H_
 
 #include <memory>
+
 #include "curvefs/proto/common.pb.h"
 #include "curvefs/src/metaserver/inode_storage.h"
-
-
 
 namespace curvefs {
 namespace metaserver {
@@ -37,20 +36,21 @@ using curvefs::common::PartitionInfo;
 class S3Compact {
  public:
     explicit S3Compact(std::shared_ptr<InodeStorage> inodeStorage,
-                       PartitionInfo* partitionInfo)
+                       const PartitionInfo& partitionInfo)
         : inodeStorage_(inodeStorage), partitionInfo_(partitionInfo) {}
 
     std::shared_ptr<InodeStorage> GetMutableInodeStorage() {
         return inodeStorage_;
     }
 
-    PartitionInfo GetPartition() {
-        return *partitionInfo_;
+    PartitionInfo GetPartition() const {
+        return partitionInfo_;
     }
 
  private:
     std::shared_ptr<InodeStorage> inodeStorage_;
-    PartitionInfo* partitionInfo_;
+    // we only need data that will not be changed after being created.
+    const PartitionInfo partitionInfo_;
 };
 
 }  // namespace metaserver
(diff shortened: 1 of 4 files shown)


curvefs/src/metaserver/s3compact.h
```diff
@@ -37,20 +36,21 @@ using curvefs::common::PartitionInfo;
...
 
     PartitionInfo GetPartition() {
-        return *partitionInfo_;
+        return partitionInfo_;
     }
 
  private:
     std::shared_ptr<InodeStorage> inodeStorage_;
-    PartitionInfo* partitionInfo_;
+    // we only need data that will not be changed after being created.
```
wu-hanqing, line 52: if so, make it const

curvefs/src/metaserver/s3compact.h
```diff
@@ -37,20 +36,21 @@ using curvefs::common::PartitionInfo;
 class S3Compact {
  public:
     explicit S3Compact(std::shared_ptr<InodeStorage> inodeStorage,
-                       PartitionInfo* partitionInfo)
+                       PartitionInfo partitionInfo)
```
wu-hanqing, line 39: const PartitionInfo& partitionInfo

curvefs/src/metaserver/s3compact.h
h0hmj: done

curvefs/src/metaserver/s3compact.h
h0hmj: done

wu-hanqing (approved):

h0hmj:
recheck

curvefs/src/metaserver/s3compact_manager.cpp
```diff
@@ -183,16 +183,26 @@ void S3CompactManager::Stop() {
...
     if (s3compacts.empty()) {
+        // fastpath
         sleeper_.wait_for(std::chrono::milliseconds(opts_.enqueueSleepMS));
+        return;
     }
-    for (const auto& s3compact : s3compacts) {
+    std::vector<int> toErase;
+    for (int i = 0; i < s3compacts.size(); i++) {
+        auto s3compact(s3compacts[i].lock());
+        if (!s3compact) {
```
cw123, line 200: how to judge partition is deleted

cw123: when partition is under deleting, the compact record should be deleted

cw123: when the partition is deleting, it will reject update inode request.

h0hmj: only partition hold the shared_ptr of s3compact, if partition is deleted, s3compact will be deleted.
weak_ptr lock() is thread safe.
if s3compact in partition has been deconstructed. lock() will return nullptr.
there is no deleting state, just deleted or not deleted.

cw123 (approved):

h0hmj:
recheck

YunhuiChen:
recheck

h0hmj:
recheck

h0hmj:
recheck

wu-hanqing:
recheck

wu-hanqing:
recheck

Issue #902, opened 2021-12-29
I/O error when curve-fuse read after metaserver compaction


h0hmj:
**Describe the bug (描述bug)**

After metaserver compaction finished, inode metadata has been updated.
But client curve-fuse use the stale inode cache, and try to read deleted s3 objects.
And there is no retry mechanism when client read, It will result in Input/Output error.

**To Reproduce (复现方法)**

1. mount
2. using fio do randrw to trigger compaction
fio -directory=/mountpoint -ioengine=psync -rw=randrw -bs=4k -direct=1 -group_reporting=1 -fallocate=none -time_based=1  -name=test_file -numjobs=1 -iodepth=128 -nrfiles=1 -size=100M -verify=md5 --runtime=600 --time_based
3. i/o error

**Expected behavior (期望行为)**

It should be success.

PR #907, opened 2021-12-30
curvefs/client: retry ReadFromS3 when needed

h0hmj:
curvefs/client: update inode cache and retry ReadFromS3 when obj does not
    exist. because objs may be deleted by s3compact, client inode cache is
    stale.

### DIFF
--- a/curvefs/src/client/s3/client_s3.cpp
+++ b/curvefs/src/client/s3/client_s3.cpp
@@ -63,6 +63,11 @@ int S3ClientImpl::Download(const std::string &name, char *buf, uint64_t offset,
             << ",length:" << length;
     ret = s3Adapter_->GetObject(name, buf, offset, length);
     if (ret < 0) {
+        const Aws::String aws_key(name.c_str(), name.size());
+        if (!s3Adapter_->ObjectExist(aws_key)) {
+            LOG(INFO) << "obj " << name << " seems not exist";
+            ret = -2;
+        }
         LOG(ERROR) << "download error:" << ret;
     }
 
--- a/curvefs/src/client/s3/client_s3_cache_manager.cpp
+++ b/curvefs/src/client/s3/client_s3_cache_manager.cpp
@@ -275,67 +275,95 @@ int FileCacheManager::Read(uint64_t inodeId, uint64_t offset, uint64_t length,
         return readOffset;
     }
 
-    std::vector<S3ReadRequest> totalS3Requests;
-    auto iter = totalRequests.begin();
-    uint64_t fileLen;
     {
+        unsigned int maxRetry = 3;  // hardcode, fixme
+        unsigned int retry = 0;
         std::shared_ptr<InodeWrapper> inodeWrapper;
-        CURVEFS_ERROR ret = s3ClientAdaptor_->GetInodeCacheManager()->GetInode(
-            inodeId, inodeWrapper);
-        if (ret != CURVEFS_ERROR::OK) {
+        auto inodeManager = s3ClientAdaptor_->GetInodeCacheManager();
+        CURVEFS_ERROR r = inodeManager->GetInode(inodeId, inodeWrapper);
+        if (r != CURVEFS_ERROR::OK) {
             LOG(WARNING) << "get inode fail, ret:" << ret;
             return -1;
         }
-        ::curve::common::UniqueLock lgGuard = inodeWrapper->GetUniqueLock();
-        Inode *inode = inodeWrapper->GetMutableInodeUnlocked();
-        fileLen = inode->length();
-        for (; iter != totalRequests.end(); iter++) {
-            VLOG(6) << "ReadRequest index:" << iter->index
-                    << ",chunkPos:" << iter->chunkPos << ",len:" << iter->len
-                    << ",bufOffset:" << iter->bufOffset;
-            auto s3InfoListIter = inode->s3chunkinfomap().find(iter->index);
-            if (s3InfoListIter == inode->s3chunkinfomap().end()) {
-                VLOG(6) << "s3infolist is not found.index:" << iter->index;
-                memset(dataBuf + iter->bufOffset, 0, iter->len);
-                continue;
+        std::vector<S3ReadResponse> responses;
+        while (retry < maxRetry) {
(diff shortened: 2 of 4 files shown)


h0hmj:
related to #902

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -275,69 +275,97 @@ int FileCacheManager::Read(uint64_t inodeId, uint64_t offset, uint64_t length,
         return readOffset;
     }
 
-    std::vector<S3ReadRequest> totalS3Requests;
-    auto iter = totalRequests.begin();
-    uint64_t fileLen;
     {
+        unsigned int maxRetry = 3;  // hardcode, fixme
```
xu-chaojie, line 279: Retry 3 times is not enough maybe

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -275,69 +275,97 @@ int FileCacheManager::Read(uint64_t inodeId, uint64_t offset, uint64_t length,
...
-    for (; repIter != responses.end(); repIter++) {
-        VLOG(6) << "readOffset:" << repIter->GetReadOffset()
-                << ",bufLen:" << repIter->GetBufLen();
-        memcpy(dataBuf + repIter->GetReadOffset(), repIter->GetDataBuf(),
-               repIter->GetBufLen());
+            auto repIter = responses.begin();
+            for (; repIter != responses.end(); repIter++) {
+                VLOG(6) << "readOffset:" << repIter->GetReadOffset()
+                        << ",bufLen:" << repIter->GetBufLen();
+                memcpy(dataBuf + repIter->GetReadOffset(),
```
xu-chaojie, line 363: Should we move this memcpy outside the while loop?

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -275,69 +275,97 @@ int FileCacheManager::Read(uint64_t inodeId, uint64_t offset, uint64_t length,
...
 
-    std::vector<S3ReadResponse> responses;
+            std::vector<S3ReadResponse> responses;
 
-    ret = ReadFromS3(totalS3Requests, &responses, fileLen);
-    if (ret < 0) {
-        LOG(ERROR) << "read from s3 failed. ret:" << ret;
-        return ret;
-    }
+            ret = ReadFromS3(totalS3Requests, &responses, fileLen);
```
xu-chaojie, line 339: When ret = -2,  all of the totalS3Requests will be retry? even some one read succeed?

curvefs/src/client/s3/client_s3_cache_manager.cpp
h0hmj: if success, loop will break, it's ok.

curvefs/src/client/s3/client_s3_cache_manager.cpp
h0hmj: if obj does not exit, we assume the s3infomap cache is stale.
totalS3Requests is generated by stale inode s3infomap cache.
we need retry the whole process, including generating totalS3Request and send request.

curvefs/src/client/s3/client_s3_cache_manager.cpp
h0hmj: we only retry when cache is stale, retry once(clear cache and get inode) and success is expected.
i think 3 is enough. will make it a config later, but later.

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -275,69 +275,97 @@ int FileCacheManager::Read(uint64_t inodeId, uint64_t offset, uint64_t length,
...
+            if (ret < 0) {
```
baijiaruo, line 340: If one request fails, retry everything?

curvefs/src/client/s3/client_s3.cpp
```diff
@@ -63,6 +63,11 @@ int S3ClientImpl::Download(const std::string &name, char *buf, uint64_t offset,
             << ",length:" << length;
     ret = s3Adapter_->GetObject(name, buf, offset, length);
     if (ret < 0) {
+        const Aws::String aws_key(name.c_str(), name.size());
```
baijiaruo, line 66: Why not use a separate interface, but put it in getobject

curvefs/src/client/s3/client_s3_cache_manager.cpp
h0hmj: if obj does not exit, we assume the s3infomap cache is stale.
totalS3Requests is generated by the stale inode s3infomap cache.
we need retry the whole process, including generating totalS3Request and send request.

curvefs/src/client/s3/client_s3.cpp
h0hmj: won't fix. s3 sdk has been wrapped so many times...

curvefs/src/client/s3/client_s3_cache_manager.cpp
h0hmj: fixed

xu-chaojie (approved):

h0hmj:
recheck

h0hmj:
recheck

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -275,67 +275,95 @@ int FileCacheManager::Read(uint64_t inodeId, uint64_t offset, uint64_t length,
         return readOffset;
     }
 
-    std::vector<S3ReadRequest> totalS3Requests;
-    auto iter = totalRequests.begin();
-    uint64_t fileLen;
     {
+        unsigned int maxRetry = 3;  // hardcode, fixme
```
wu-hanqing, line 279: did you add test case or modify existing test case about this?

h0hmj: maybe later

wu-hanqing (approved):

Issue #883, opened 2021-12-23
curvefs delete fs in backgroud


cw123:
curvefs增加后台删除fs功能，可通过curvefs tool工具删除fs。mds会标记对fs重命名并在后台进行删除。当fs的所有数据和元数据全部清除之后，删除记录fs。
deletefs之后，用户可立即创建同名fs。

---------------------------------------------------------------------------

Curvefs adds the function of deleting fs in the background, and fs can be deleted through the curvefs tool. mds will mark the fs rename and delete it in the background. When all the data and metadata of the fs are cleared, the record fs is deleted.
After deletefs, users can immediately create fs with the same name.

cw123:
如果S3 sdk支持deleteObjects，能够批量删除记录，设置metasever.conf的s3.enableDeleteObjects为true，并通过s3.batchsize配置项限制每次deleteObjects请求删除的objects的个数。
如果s3.enableDeleteObjects配置为False，s3.batchsize配置项不生效。

=====

If S3 sdk supports deleteObjects, you can delete records in batches. Set s3.enableDeleteObjects in metasever.conf to true, and limit the number of objects deleted each time by the s3.batchsize configuration item.
If s3.enableDeleteObjects is configured as False, the s3.batchsize configuration item will not take effect.

cw123:
deleteFs不支持多bucket删除S3数据

===

deleteFs does not support delete s3 from multi bucket.

PR #947, opened 2022-01-07
curvefs/mds:fix delete fs

cw123:

Issue Number: #883 

Problem Summary: fix some bug when delete fs

What's Changed:
1. Mds mark partition to DELETING after call metaserver DeletePartition.
2. When GetInode, delete check Partition status.
3. Only leader copyset can should delete inode and partition from raft.
4. s3 param conf in metaserver.conf and client.conf is mismatch. Change theas param same.

How it Works:

### Check List

### DIFF
--- a/curvefs/src/metaserver/partition.cpp
+++ b/curvefs/src/metaserver/partition.cpp
@@ -220,10 +220,6 @@ MetaStatusCode Partition::GetInode(uint32_t fsId, uint64_t inodeId,
         return MetaStatusCode::PARTITION_ID_MISSMATCH;
     }
 
-    if (GetStatus() == PartitionStatus::DELETING) {
-        return MetaStatusCode::PARTITION_DELETING;
-    }
-
     return inodeManager_->GetInode(fsId, inodeId, inode);
 }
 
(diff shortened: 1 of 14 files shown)


cw123:
recheck

cw123:
recheck

cw123:
recheck

curvefs/src/metaserver/partition_cleaner.cpp
```diff
@@ -28,18 +28,6 @@ namespace curvefs {
...
-        MetaStatusCode ret = DeletePartition();
-        if (ret == MetaStatusCode::OK) {
-            return true;
-        } else {
-            LOG(WARNING) << "delete partition from copyset fail, ret = "
-                         << MetaStatusCode_Name(ret);
-            return false;
-        }
-    }
-
```
wu-hanqing, line 42: you should put below if condition into for-loop, because the leader may be transferred.

curvefs/src/metaserver/partition.cpp
```diff
@@ -220,10 +220,6 @@ MetaStatusCode Partition::GetInode(uint32_t fsId, uint64_t inodeId,
         return MetaStatusCode::PARTITION_ID_MISSMATCH;
     }
 
-    if (GetStatus() == PartitionStatus::DELETING) {
```
wu-hanqing, line 223: why remove this check?

curvefs/src/metaserver/partition_cleaner.cpp
cw123: the new leader will cover this situation

curvefs/src/metaserver/partition.cpp
cw123: a deleting partition can read, not write

ilixiaocui, line 223: Why a under-deleting partition is readable?

cw123: When delete inode, the partition will get the inode first.

curvefs/src/metaserver/partition.cpp
```diff
@@ -296,10 +292,18 @@ bool Partition::IsDeletable() {
...
 bool Partition::IsInodeBelongs(uint32_t fsId, uint64_t inodeId) {
     if (fsId != partitionInfo_.fsid()) {
+        LOG(WARNING) << "partition fsid mismatch, fsId = " << fsId
+                     << ", inodeId = " << inodeId
+                     << ", partition fsId = " << partitionInfo_.fsid();
         return false;
     }
 
     if (inodeId < partitionInfo_.start() || inodeId > partitionInfo_.end()) {
+        LOG(WARNING) << "partition inode mismatch, fsId = " << fsId
```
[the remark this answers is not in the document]
cw123: Under normal circumstances, the code will not come here.

ilixiaocui (approved):

wu-hanqing (approved):

PR #954, opened 2022-01-07
curvefs: support replace a servev which donen't works anymore.

SeanHai:
If a server with some metaservers on it, and it crashed. Then will need
add a new server into the cluster.

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


SeanHai:
recheck

curvefs/proto/topology.proto
```diff
@@ -38,7 +38,7 @@ enum TopoStatusCode {
     TOPO_IP_PORT_DUPLICATED = 14;
     TOPO_NAME_DUPLICATED = 15;
     TOPO_CREATE_COPYSET_ON_METASERVER_FAIL = 16;
-    TOPO_CANNOT_REMOVE_NOT_RETIRED = 17;
+    TOPO_CANNOT_REMOVE_NOT_OFFLINE = 17;
```
wu-hanqing, line 41: do we have `RETIRED` state for metaserver now?

curvefs/src/tools/create/curvefs_create_topology_tool.h
```diff
@@ -133,6 +133,9 @@ class CurvefsBuildTopologyTool : public curvefs::tools::CurvefsTool {
     int InitPoolData();
     int ScanCluster();
     int ScanPool();
+    int ClearPool();
+    int ClearZone();
+    int ClearServer();
```
wu-hanqing, line 138: `Clear` is miss leading

curvefs/src/tools/create/curvefs_create_topology_tool.cpp
```diff
@@ -415,6 +436,118 @@ int CurvefsBuildTopologyTool::GetServersInZone(
...
+int CurvefsBuildTopologyTool::ClearPool() {
+    TopologyService_Stub stub(&channel_);
+    for (auto it : poolToDel) {
+        DeletePoolRequest request;
+        DeletePoolResponse response;
+        request.set_poolid(it);
+
+        brpc::Controller cntl;
+        cntl.set_timeout_ms(FLAGS_rpcTimeoutMs);
+        cntl.set_log_id(1);
```
wu-hanqing, line 448: if you don't want to use `log_id` to trace RPC calls, there's no need to set it. otherwise, you have to set a different ID for each RPC call.

curvefs/proto/topology.proto
SeanHai: > do we have `RETIRED` state for metaserver now?

Metaserver does't have this state.

curvefs/src/tools/create/curvefs_create_topology_tool.h
SeanHai: > `Clear` is miss leading

fix

curvefs/src/tools/create/curvefs_create_topology_tool.cpp
SeanHai: > if you don't want to use `log_id` to trace RPC calls, there's no need to set it. otherwise, you have to set a different ID for each RPC call.

fix

aspirer:
Please associate the pull request with an issue.

wu-hanqing (approved):

Wine93:
LGTM!

Wine93 (approved):

PR #976, opened 2022-01-11
curvefs/tool: list-topology output json file

### LINKED ISSUE #975: curvefs/tool: list-topology output jsonfile
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

list-topology output jsonfile wich can be used to create-topology

**Describe the solution you'd like (描述你期望的解决方法)**

use the output of  list-topology to build a json file

**Describe alternatives you've considered (描述你想到的折衷方案)**

Cyber-SiKu:
add output json file which can be used for create-topology

Issue Number: close #975  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/tools/curvefs_tool_define.h
+++ b/curvefs/src/tools/curvefs_tool_define.h
@@ -133,6 +133,14 @@ const char kHostFollowerValue[] = "follower";
 const char kEtcdLeaderValue[] = "StateLeader";
 const char kEtcdFollowerValue[] = "StateFollower";
 
+/* fs type */
+const char kFsTypeS3[] = "s3";
+const char kFsTypeVolume[] = "volume";
+
+/* json type */
+const char kJsonTypeBuild[] = "build";
+const char kJsonTypeTree[] = "tree";
+
 }  // namespace tools
 }  // namespace curvefs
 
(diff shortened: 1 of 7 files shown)


Cyber-SiKu:
recheck

curvefs/src/tools/curvefs_tool_define.h
```diff
@@ -133,6 +133,14 @@ const char kHostFollowerValue[] = "follower";
 const char kEtcdLeaderValue[] = "StateLeader";
 const char kEtcdFollowerValue[] = "StateFollower";
 
+/* fs type */
+const char kFsTypeS3[] = "s3";
+const char kFsTypeVolume[] = "volume";
+
+/* json type */
+const char kJsonTypeBuild[] = "build";
```
SeanHai, line 141: What's the difference of json type of 'build' and 'tree'?

Cyber-SiKu: > What's the difference of json type of 'build' and 'tree'?

build jsonfile can be used to build topology,
tree json file will form a tree of topology according to the structure of cluser, pool, zone, server, metaserver

SeanHai (approved):

curvefs/src/tools/list/curvefs_topology_list.h
```diff
@@ -133,36 +137,64 @@ class TopologyListTool
...
+
+    bool GetClusterTree(Json::Value* cluster, const std::string& clusterId);
+
+    bool GetPoolTree(Json::Value* pool, uint64_t poolId);
+
+    bool GetZoneTree(Json::Value* zone, uint64_t zoneId);
+
+    bool GetServerTree(Json::Value* server, uint64_t serverId);
+
+    bool GetMetaserverTree(Json::Value* metaserver, uint64_t metaserverId);
```
wu-hanqing, line 154: keep a class small, so this format logic should not be put in here.
condier define tree builder / json builder into single class

curvefs/src/tools/list/curvefs_topology_list.h
```diff
@@ -133,36 +137,64 @@ class TopologyListTool
...
     std::map<mds::topology::MetaServerIdType, MetaserverInfoType>
-        metaserverId2MetaserverInfo;
+        metaserverId2MetaserverInfo_;
+
+    /**
+     * @brief cluster map, like conf/topology.json
+     *
+     * @details
+     */
+    Json::StreamWriterBuilder clusterMap_;
```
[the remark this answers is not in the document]
Cyber-SiKu: > there's no need to define it as a data member

It is reserved to add a return function if other commands need to use it in the future, but it is also a good idea to modify it at that time.

curvefs/src/tools/list/curvefs_topology_list.h
Cyber-SiKu: > keep a class small, so this format logic should not be put in here. condier define tree builder / json builder into single class

fix

wu-hanqing (approved):

Issue #995, opened 2022-01-14
Copyset migration fails when there is a file system deletion task


YunhuiChen:
**Describe the bug (描述bug)**
Copyset migration fails when there is a file system deletion task

**To Reproduce (复现方法)**
![image](https://user-images.githubusercontent.com/67582932/149464942-c900790d-c622-4a5a-950d-9342c231b336.png)

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
curve-mds:
curve-chunkserver:
curve-snapshotcloneserver:
curve-sdk:
nebd:
curve-nbd:

**Additional context/screenshots (更多上下文/截图)**

cw123:
fix merge to master #997 
fix merge to fs-release0.2.0 #1016

PR #997, opened 2022-01-14
fix insert inode fail when load deleting partition

cw123:

Issue Number:  #995

Problem Summary: fix insert inode fail when load deleting partition

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/metaserver/metastore.cpp
+++ b/curvefs/src/metaserver/metastore.cpp
@@ -49,15 +49,6 @@ bool MetaStoreImpl::LoadPartition(uint32_t partitionId, void* entry) {
     partitionId = partitionInfo->partitionid();
     auto partition = std::make_shared<Partition>(*partitionInfo);
     partitionMap_.emplace(partitionId, partition);
-    if (partitionInfo->status() == PartitionStatus::DELETING) {
-        std::shared_ptr<PartitionCleaner> partitionCleaner =
-            std::make_shared<PartitionCleaner>(GetPartition(partitionId));
-        copyset::CopysetNode *copysetNode =
-            copyset::CopysetNodeManager::GetInstance().GetCopysetNode(
-                partition->GetPoolId(), partition->GetCopySetId());
-        PartitionCleanManager::GetInstance().Add(partitionId, partitionCleaner,
-                                                 copysetNode);
-    }
     return true;
 }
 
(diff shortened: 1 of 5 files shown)


xu-chaojie (approved):

curvefs/src/metaserver/metastore.cpp
```diff
@@ -69,9 +60,10 @@ bool MetaStoreImpl::LoadInode(uint32_t partitionId, void* entry) {
     }
 
     auto inode = reinterpret_cast<Inode*>(entry);
-    auto rc = partition->InsertInode(*inode);
+    MetaStatusCode rc = partition->InsertInode(*inode);
     if (rc != MetaStatusCode::OK) {
-        LOG(ERROR) << "InsertInode failed, retCode = " << rc;
+        LOG(ERROR) << "InsertInode failed, retCode = "
```
xu-chaojie, line 65: using “：” instead of “=” in log

curvefs/src/metaserver/metastore.cpp
```diff
@@ -137,6 +130,20 @@ bool MetaStoreImpl::Load(const std::string& pathname) {
         partitionMap_.clear();
         LOG(ERROR) << "Load metadata failed.";
     }
+
+    for (auto it = partitionMap_.begin(); it != partitionMap_.end(); it++) {
```
wu-hanqing, line 134: use range-based for loop

wu-hanqing (commented):
you should add tests to verify this patch.

curvefs/src/metaserver/metastore.cpp
cw123: The two methods are equivalent

cw123:
> you should add tests to verify this patch.

done, add test case

curvefs/test/metaserver/metastore_test.cpp
```diff
@@ -94,7 +94,10 @@ class MetastoreTest : public ::testing::Test {
                    first.partitionid() == second.partitionid() &&
                    first.start() == second.start() &&
                    first.end() == second.end() &&
-                   first.nextid() == second.nextid();
+                   first.nextid() == second.nextid() &&
```
wu-hanqing, line 97: you can use `google::protobuf::util::MessageDifferencer::Equals` to simplify this compare

wu-hanqing (approved):

cw123:
recheck

cw123:
recheck

cw123:
recheck

Issue #998, opened 2022-01-14
curvefs_tool create-fs failed, error is CREATE_PARTITION_ERROR


wu-hanqing:
**Describe the bug (描述bug)**

![image](https://user-images.githubusercontent.com/10327590/149468878-978abdca-0cb5-423b-bd0a-0743c3d9c0df.png)

**To Reproduce (复现方法)**

use CurveAdm deploy a cluster with three etcd/mds/metaserver, and after deploy succeed, use curvefs_tool create a filesystem.

**Expected behavior (期望行为)**

**Versions (各种版本)**
fs: v0.2.0-beta

**Additional context/screenshots (更多上下文/截图)**

wu-hanqing:
master mds logs

```
I 2022-01-14T14:54:29.571087+0800    48 metaserver_healthy_checker.cpp:63] metaServer 1 is online
I 2022-01-14T14:54:29.571146+0800    48 metaserver_healthy_checker.cpp:63] metaServer 2 is online
I 2022-01-14T14:54:29.571152+0800    48 metaserver_healthy_checker.cpp:63] metaServer 3 is online
I 2022-01-14T14:54:29.571719+0800    47 recoverScheduler.cpp:34] recoverScheduler begin.
I 2022-01-14T14:54:29.571755+0800    47 recoverScheduler.cpp:123] recoverScheduler generate 0 operators at this round
I 2022-01-14T14:54:34.571838+0800    47 recoverScheduler.cpp:34] recoverScheduler begin.
I 2022-01-14T14:54:34.571890+0800    47 recoverScheduler.cpp:123] recoverScheduler generate 0 operators at this round
I 2022-01-14T14:54:37.933858+0800    26 mds_service.cpp:71] CreateFs request, fsName = curvefs, blockSize = 1048576, s3Info.bucketname = wuhanqing
I 2022-01-14T14:54:37.935137+0800    26 etcd_client.cpp:89] key 'fs_02' not exist
I 2022-01-14T14:54:39.571966+0800    47 recoverScheduler.cpp:34] recoverScheduler begin.
I 2022-01-14T14:54:39.572010+0800    47 recoverScheduler.cpp:123] recoverScheduler generate 0 operators at this round
I 2022-01-14T14:54:40.596598+0800    26 topology_manager.cpp:620] Have no available copyset when create partition.
I 2022-01-14T14:54:44.572091+0800    47 recoverScheduler.cpp:34] recoverScheduler begin.
[...]

PR #1002, opened 2022-01-17
curvefs: fix create copyset when exist already.

SeanHai:
curvefs: fix create copyset when exist already. If peers are all the same will success, otherwise return exist.

Issue Number: close #xxx 
#998 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/metaserver/copyset/copyset_node_manager.cpp
+++ b/curvefs/src/metaserver/copyset/copyset_node_manager.cpp
@@ -164,10 +164,33 @@ CopysetNode* CopysetNodeManager::GetCopysetNode(PoolId poolId,
     return nullptr;
 }
 
-bool CopysetNodeManager::IsCopysetNodeExist(PoolId poolId,
-                                            CopysetId copysetId) {
+int CopysetNodeManager::IsCopysetNodeExist(
+    const CreateCopysetRequest::Copyset& copyset) {
     ReadLockGuard lock(lock_);
-    return copysets_.count(ToGroupId(poolId, copysetId)) != 0;
+    auto iter = copysets_.find(ToGroupId(copyset.poolid(),
+                                         copyset.copysetid()));
+    if (iter == copysets_.end()) {
+        return 0;
+    } else {
+        auto copysetNode = iter->second.get();
+        std::vector<Peer> peers;
+        copysetNode->ListPeers(&peers);
+        if (peers.size() != copyset.peers_size()) {
+            return -1;
+        }
+
+        for (int i = 0; i < copyset.peers_size(); i++) {
+            auto cspeer = copyset.peers(i);
+            auto iter = std::find_if(
+            peers.begin(), peers.end(),
+            [&cspeer](const Peer& p) { return
+                cspeer.address() == p.address();});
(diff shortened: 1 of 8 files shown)


SeanHai:
recheck

curvefs/src/metaserver/copyset/copyset_node_manager.h
```diff
@@ -57,7 +57,7 @@ class CopysetNodeManager {
 
     virtual CopysetNode* GetCopysetNode(PoolId poolId, CopysetId copysetId);
 
-    bool IsCopysetNodeExist(PoolId poolId, CopysetId copysetId);
+    int IsCopysetNodeExist(const CreateCopysetRequest::Copyset& copyset);
```
cw123, line 60: add meaning of return value

curvefs/src/metaserver/copyset/copyset_node_manager.cpp
```diff
@@ -164,10 +164,29 @@ CopysetNode* CopysetNodeManager::GetCopysetNode(PoolId poolId,
...
     ReadLockGuard lock(lock_);
-    return copysets_.count(ToGroupId(poolId, copysetId)) != 0;
+    auto iter = copysets_.find(ToGroupId(copyset.poolid(),
+                                         copyset.copysetid()));
+    if (iter == copysets_.end()) {
+        return 0;
+    } else {
+        auto copysetNode = iter->second.get();
+        std::vector<Peer> peers;
+        copysetNode->ListPeers(&peers);
```
cw123, line 177: compare peer num

peers [1, 2, 3]  peers[1, 2, 3, 4] is not same

curvefs/src/metaserver/copyset/copyset_node_manager.h
SeanHai: > add meaning of return value

fixed

curvefs/src/metaserver/copyset/copyset_node_manager.cpp
SeanHai: > compare peer num
> 
> peers [1, 2, 3] peers[1, 2, 3, 4] is not same

fixed

cw123 (approved):

curvefs/src/metaserver/copyset/copyset_service.cpp
```diff
@@ -79,13 +79,17 @@ void CopysetServiceImpl::GetCopysetsStatus(
 
 COPYSET_OP_STATUS CopysetServiceImpl::CreateOneCopyset(
     const CreateCopysetRequest::Copyset& copyset) {
-    bool exists =
-        manager_->IsCopysetNodeExist(copyset.poolid(), copyset.copysetid());
-    if (exists) {
+    int exists = manager_->IsCopysetNodeExist(copyset);
```
wu-hanqing, line 82: another problem is above `CreateCopysetNode` will stop directly when failed to create one copyset, so you can find whether there's a good reason to continue creating all copysets and return status for each copyset.

wu-hanqing (approved):

Issue #999, opened 2022-01-14
copyset is been removed just after created


wu-hanqing:
**Describe the bug (描述bug)**

copyset is been removed just after created

**To Reproduce (复现方法)**

use CurveAdm deploy a cluster with three etcd/mds/metaserver, and after deploy succeed, use curvefs_tool create a filesystem.

**Expected behavior (期望行为)**

**Versions (各种版本)**
fs: v0.2.0-beta

**Additional context/screenshots (更多上下文/截图)**

metaserver logs

```
[...]

PR #1011, opened 2022-01-18
curvefs mds heartbeat doesn't delete copyset creating.

SeanHai:
curvefs mds heartbeat doesn't delete copyset creating.
The empty conf will send to metaserver to delete copysets which don't exist in mds, but when the copyset just created in metaserver and have not add to topology, the metaserver heartbeat will contain these copysets' infomation.

Issue Number: close #xxx 
#999 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/mds/topology/topology.cpp
+++ b/curvefs/src/mds/topology/topology.cpp
@@ -991,6 +991,13 @@ TopoStatusCode TopologyImpl::AddCopySet(const CopySetInfo &data) {
     }
 }
 
+TopoStatusCode TopologyImpl::AddCopySetCreating(const CopySetKey &key) {
+    WriteLockGuard wlockCopySetCreating(copySetCreatingMutex_);
+    auto iter = copySetCreating_.insert(key);
+    return iter.second ? TopoStatusCode::TOPO_OK :
+        TopoStatusCode::TOPO_ID_DUPLICATED;
+}
+
 TopoStatusCode TopologyImpl::RemoveCopySet(CopySetKey key) {
     WriteLockGuard wlockCopySetMap(copySetMutex_);
     auto it = copySetMap_.find(key);
@@ -1003,6 +1010,11 @@ TopoStatusCode TopologyImpl::RemoveCopySet(CopySetKey key) {
     } else {
         return TopoStatusCode::TOPO_COPYSET_NOT_FOUND;
     }
+}
+
+void TopologyImpl::RemoveCopySetCreating(CopySetKey key) {
+    WriteLockGuard wlockCopySetCreating(copySetCreatingMutex_);
+    copySetCreating_.erase(key);
 }
 
 TopoStatusCode TopologyImpl::UpdateCopySetTopo(const CopySetInfo &data) {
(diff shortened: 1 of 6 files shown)


SeanHai:
recheck

SeanHai:
recheck

curvefs/src/mds/topology/topology.h
```diff
@@ -445,6 +451,8 @@ class TopologyImpl : public Topology {
 
     std::string GetHostNameAndPortById(MetaServerIdType msId) override;
 
+    bool IsCopysetCreating(const CopySetKey &key) override;
```
wu-hanqing, line 454: declare it as const member functions

curvefs/src/mds/topology/topology.cpp
```diff
@@ -1358,6 +1370,13 @@ std::string TopologyImpl::GetHostNameAndPortById(MetaServerIdType msId) {
     // get hostName of the metaserver
     return server.GetHostName() + ":" + std::to_string(ms.GetInternalPort());
 }
+
+bool TopologyImpl::IsCopysetCreating(const CopySetKey &key) {
+    ReadLockGuard rlockCopySetCreating(copySetCreatingMutex_);
+    return copySetCreating_.find(key) == copySetCreating_.end() ? false : true;
```
wu-hanqing, line 1376:
```cpp
return copysetCreating_.find(key) != copysetCreating_.end();
// or
return coypsetCreating_.count(key) != 0;
```

curvefs/src/mds/topology/topology_manager.cpp
```diff
@@ -721,6 +721,13 @@ bool TopologyManager::CreateCopysetNodeOnMetaServer(
...
 }
 
+void TopologyManager::ClearCopysetCreating(PoolIdType poolId,
+    const std::set<CopySetIdType> &copysets) {
+    for (const auto &id : copysets) {
+        topology_->RemoveCopySetCreating(CopySetKey(poolId, id));
+    }
+}
+
 TopoStatusCode TopologyManager::CreateCopyset() {
```
cw123, line 731: Does CreateCopyset have concurrency?

curvefs/src/mds/topology/topology.h
SeanHai: > declare it as const member functions

fixed.

curvefs/src/mds/topology/topology.cpp
SeanHai: > ```c++
> return copysetCreating_.find(key) != copysetCreating_.end();
> // or
> return coypsetCreating_.count(key) != 0;
> ```

fixed

wu-hanqing (approved):

curvefs/src/mds/topology/topology_manager.cpp
SeanHai: > Does CreateCopyset have concurrency?

Concurrency occurs in extreme cases, but the copysets' id must be different.

cw123 (approved):

Issue #810, opened 2021-12-09
curvefs: The performance of MDTest does not improve after concurrency is added


YunhuiChen:
**Describe the bug (描述bug)**
The performance of MDTest does not improve after concurrency is added

**To Reproduce (复现方法)**

![image](https://user-images.githubusercontent.com/67582932/145318065-1ad7cf9e-4fa5-4b0c-8799-ef15cd137afe.png)

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
curve-mds:
curve-chunkserver:
curve-snapshotcloneserver:
curve-sdk:
nebd:
curve-nbd:

**Additional context/screenshots (更多上下文/截图)**

SeanHai:
pr: https://github.com/opencurve/curve/pull/1052
![截屏2022-01-24 上午10 58 08](https://user-images.githubusercontent.com/13496900/150715035-865dc421-3d3a-4de0-b116-ff4c688ebebb.png)

PR #1052, opened 2022-01-24
curvefs improve the inode and dentry cache lock

SeanHai:

Issue Number: close #xxx 
#810 
Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/dentry_cache_manager.cpp
+++ b/curvefs/src/client/dentry_cache_manager.cpp
@@ -34,19 +34,18 @@ namespace curvefs {
 namespace client {
 
 using curve::common::WriteLockGuard;
+using NameLockGuard = ::curve::common::GenericNameLockGuard<Mutex>;
 
 void DentryCacheManagerImpl::InsertOrReplaceCache(const Dentry &dentry) {
     std::string key = GetDentryCacheKey(dentry.parentinodeid(), dentry.name());
-
-    WriteLockGuard lg(mtx_);
+    NameLockGuard lock(nameLock_, key);
     dCache_->Put(key, dentry);
 }
 
 void DentryCacheManagerImpl::DeleteCache(uint64_t parentId,
                                          const std::string &name) {
     std::string key = GetDentryCacheKey(parentId, name);
-
-    WriteLockGuard lg(mtx_);
+    NameLockGuard lock(nameLock_, key);
     dCache_->Remove(key);
 }
 
@@ -54,14 +53,7 @@ CURVEFS_ERROR DentryCacheManagerImpl::GetDentry(uint64_t parent,
                                                 const std::string &name,
                                                 Dentry *out) {
     std::string key = GetDentryCacheKey(parent, name);
-    {
-        curve::common::ReadLockGuard lg(mtx_);
-        bool ok = dCache_->Get(key, out);
-        if (ok) {
-            return CURVEFS_ERROR::OK;
-        }
-    }
-    curve::common::WriteLockGuard lg(mtx_);
+    NameLockGuard lock(nameLock_, key);
     bool ok = dCache_->Get(key, out);
     if (ok) {
         return CURVEFS_ERROR::OK;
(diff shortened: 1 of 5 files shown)


SeanHai:
recheck

curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -44,12 +44,6 @@ CURVEFS_ERROR InodeCacheManagerImpl::GetInode(uint64_t inodeid,
...
     }
 
-    curve::common::WriteLockGuard lg(mtx_);
-    bool ok = iCache_->Get(inodeid, &out);
-    if (ok) {
-        return CURVEFS_ERROR::OK;
-    }
-
     Inode inode;
     MetaStatusCode ret2 = metaClient_->GetInode(fsId_, inodeid, &inode);
```
xu-chaojie, line 48: if don't lock GetInode here, when update inode between two GetInode was called at the same time,  maybe second GetInode will return the old version inode？

xu-chaojie, line 48: If not lock GetInode here， when an update inode and read inode concuently call， the read inode maybe return old version of inode

curvefs/src/client/dentry_cache_manager.cpp
```diff
@@ -61,11 +59,6 @@ CURVEFS_ERROR DentryCacheManagerImpl::GetDentry(uint64_t parent,
             return CURVEFS_ERROR::OK;
         }
     }
-    curve::common::WriteLockGuard lg(mtx_);
-    bool ok = dCache_->Get(key, out);
-    if (ok) {
-        return CURVEFS_ERROR::OK;
-    }
 
     MetaStatusCode ret = metaClient_->GetDentry(fsId_, parent, name, out);
```
xu-chaojie, line 63: if don‘t lock GetDentry here， when concurrent with InsertOrReplaceCache above， this fuction maybe return old version of dentry？

curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -64,7 +58,12 @@ CURVEFS_ERROR InodeCacheManagerImpl::GetInode(uint64_t inodeid,
         std::move(inode), metaClient_);
 
     std::shared_ptr<InodeWrapper> eliminatedOne;
-    bool eliminated = iCache_->Put(inodeid, out, &eliminatedOne);
+    bool eliminated;
```
ilixiaocui, line 61: eliminated need to be initialized

curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -91,7 +89,11 @@ CURVEFS_ERROR InodeCacheManagerImpl::CreateInode(
...
```
ilixiaocui, line 92: ditto

curvefs/src/client/dentry_cache_manager.cpp
ilixiaocui: we need get from cache first, if cache do not have this one, then  get from metaserver.

may be readlock can be use by get.

SeanHai: Using namelock to lock the specific one. If lock entire icache will block requests to other inode.

SeanHai: > if don‘t lock GetDentry here， when concurrent with InsertOrReplaceCache above， this fuction maybe return old version of dentry？

Using namelock to lock the specific one. If lock entire icache will block requests to other inode.

SeanHai: > we need get from cache first, if cache do not have this one, then get from metaserver.
> 
> may be readlock can be use by get.

The key issue is the lock for dCache will block requests on other dentry. Using namelock to lock the specific one.

curvefs/src/client/inode_cache_manager.cpp
SeanHai: > eliminated need to be initialized

fixed

curvefs/src/client/inode_cache_manager.cpp
SeanHai: > ditto

fixed

SeanHai:
recheck

ilixiaocui (approved):

SeanHai:
recheck

SeanHai:
recheck

curvefs/src/client/dentry_cache_manager.cpp
```diff
@@ -61,10 +60,14 @@ CURVEFS_ERROR DentryCacheManagerImpl::GetDentry(uint64_t parent,
...
     }
-    curve::common::WriteLockGuard lg(mtx_);
-    bool ok = dCache_->Get(key, out);
-    if (ok) {
-        return CURVEFS_ERROR::OK;
+
+    NameLockGuard lock(nameLock_, key);
+    {
+        curve::common::ReadLockGuard lg(mtx_);
+        bool ok = dCache_->Get(key, out);
```
xu-chaojie, line 67: LRU cache already lock put & get in its own logical，maybe mtx_ here is not needed？

SeanHai: > LRU cache already lock put & get in its own logical，maybe mtx_ here is not needed？

Yes, the lock into LRU can protect operations on cache, the namelock can sequence the operations on the same one.

xu-chaojie (approved):

Issue #1061, opened 2022-01-25
The tool shows that some copysets were not migrated successfully


YunhuiChen:
**Describe the bug (描述bug)**

The tool shows that some copies were not migrated successfully

**To Reproduce (复现方法)**
1、delete fs 
2、kill one metaserver 
3、copysets meigrated
![image](https://user-images.githubusercontent.com/67582932/150934252-5c207b24-4a6e-4b58-9ae8-3326e7050a4c.png)
![image](https://user-images.githubusercontent.com/67582932/150934260-940be9d4-fd1b-4880-8651-f28db983f37d.png)

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
curve-mds:
curve-chunkserver:
curve-snapshotcloneserver:
curve-sdk:
nebd:
curve-nbd:

**Additional context/screenshots (更多上下文/截图)**

cw123:
Because the copyset did not update its candidate from heartbeat. So it will fail when change peer copyset.

PR #1085, opened 2022-02-16
curvefs update copyset condidate

cw123:

Issue Number: #1061

Problem Summary: 

The tool shows that some copysets were not migrated successfully. Because the copyset did not update its candidate from heartbeat. So it will fail when change peer copyset.

What's Changed:  Update copyset cadidate when heartbeat

Side effects(Breaking backward compatibility? Performance regression?):  no

### Check List


cw123:
recheck

wu-hanqing (approved):

Issue #1148, opened 2022-03-08
curvefs support listxattr


SeanHai:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**
The curvefs sholud support listxattr to show all the xattr at a time.

**Describe the solution you'd like (描述你期望的解决方法)**

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

SeanHai:
<img width="548" alt="截屏2022-03-08 下午1 55 00" src="https://user-images.githubusercontent.com/13496900/157175845-8e3d3e09-06c8-4301-85a0-212454229fdf.png">

Cyber-SiKu:
recheck

PR #1149, opened 2022-03-08
curvefs: support listxattr

SeanHai:
#1148 
Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/curve_fuse_op.cpp
+++ b/curvefs/src/client/curve_fuse_op.cpp
@@ -218,6 +218,24 @@ void FuseOpGetXattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     }
 }
 
+void FuseOpListXattr(fuse_req_t req, fuse_ino_t ino, size_t size) {
+    std::unique_ptr<char[]> buf(new char[size]);
+    std::memset(buf.get(), 0, size);
+    size_t xattrSize = 0;
+    CURVEFS_ERROR ret = g_ClientInstance->FuseOpListXattr(req, ino, buf.get(),
+                                                          size, &xattrSize);
+    if (ret != CURVEFS_ERROR::OK) {
+        FuseReplyErrByErrCode(req, ret);
+        return;
+    }
+
+    if (size == 0) {
+        fuse_reply_xattr(req, xattrSize);
+    } else {
+        fuse_reply_buf(req, buf.get(), xattrSize);
+    }
+}
+
 void FuseOpReadDir(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
                    struct fuse_file_info *fi) {
     char *buffer = nullptr;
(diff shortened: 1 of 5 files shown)


SeanHai:
recheck

xu-chaojie:
recheck

Wine93:
recheck

Wine93:
recheck

SeanHai:
recheck

curvefs/src/client/fuse_client.cpp
```diff
@@ -1119,6 +1119,56 @@ CURVEFS_ERROR FuseClient::FuseOpGetXattr(fuse_req_t req, fuse_ino_t ino,
...
+    Inode inode = inodeWrapper->GetInodeLocked();
+    // get xattr key
+    for (const auto &it : inode.xattr()) {
+        // +1 because, the format is key\0key\0
+        *realSize += it.first.length() + 1;
+    }
+
+    // add summary xattr key
+    if (inode.type() == FsFileType::TYPE_DIRECTORY) {
+        *realSize += strlen(XATTRRFILES) + 1;
```
xu-chaojie, line 1143: if enable SummaryInfo, will it duplicated with xattr  store in inode.xattr?

SeanHai: > if enable SummaryInfo, will it duplicated with xattr stored in inode.xattr?

That will not happen. The curve.dir.r* don't stored in inode xattr and here only return the xattr key.

SeanHai:
recheck

xu-chaojie (approved):

SeanHai:
recheck

SeanHai:
recheck

SeanHai:
recheck

SeanHai:
recheck

curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -218,6 +218,24 @@ void FuseOpGetXattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
+    std::memset(buf.get(), 0, size);
+    size_t xattrSize = 0;
+    CURVEFS_ERROR ret = g_ClientInstance->FuseOpListXattr(req, ino, buf.get(),
+                                                          size, &xattrSize);
+    if (ret != CURVEFS_ERROR::OK) {
+        FuseReplyErrByErrCode(req, ret);
+        return;
+    }
+
+    if (size == 0) {
```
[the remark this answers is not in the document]
SeanHai: > Why not check if size==0 in the beginning？

The logic of getxattr and listxattr of fuse is:
1. first call the func and the size = 0, you need return the true size.
2. Then call the func with a bigger size the true size again when the fuse get the true size.

so you need get the content size first.

ilixiaocui (approved):

PR #1150, opened 2022-03-08
curvefs/client: fix bug of support summary info in dir xattr

SeanHai:

Issue Number: close #xxx 

Problem Summary:
1. Fix the summary info of hardlink when enableSumInDIr if false. Because batch get inodeAttr is divided by batch limit, the struct std::set will not deal with hardlink.
2. Fix UpdateParentInodeXattr dirty flag, the update of xattr doesn't based on dirty flag.
3. Disable hardlink when  enableSumInDIr is true.
4. batchGetInodeAttr and batchGetXattr from metaserver after find in inodeCache.

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/fuse_client.cpp
+++ b/curvefs/src/client/fuse_client.cpp
@@ -30,6 +30,7 @@
 #include <vector>
 #include <stack>
 #include <set>
+#include <unordered_map>
 
 #include "curvefs/proto/mds.pb.h"
 #include "curvefs/src/client/fuse_common.h"
@@ -308,9 +309,9 @@ CURVEFS_ERROR FuseClient::FuseOpOpen(fuse_req_t req, fuse_ino_t ino,
                 XAttr xattr;
                 xattr.mutable_xattrinfos()->insert({XATTRFBYTES,
                     std::to_string(length)});
-                std::list<uint64_t> parentIds;
-                if (inodeManager_->GetParent(inode->inodeid(), &parentIds)) {
-                    ret = UpdateParentInodeXattr(parentIds, xattr, false);
+                uint64_t parentId;
+                if (inodeManager_->GetParent(inode->inodeid(), &parentId)) {
+                    ret = UpdateParentInodeXattr(parentId, xattr, false);
                 } else {
                     LOG(ERROR) << "inodeManager getParent failed, inodeId = "
                                << inode;
@@ -417,8 +418,7 @@ CURVEFS_ERROR FuseClient::MakeNode(fuse_req_t req, fuse_ino_t parent,
         }
         xattr.mutable_xattrinfos()->insert({XATTRFBYTES,
             std::to_string(inodeWrapper->GetLength())});
-        ret = UpdateParentInodeXattr(
-            std::list<uint64_t>({parent}), xattr, true);
+        ret = UpdateParentInodeXattr(parent, xattr, true);
     }
 
     GetDentryParamFromInode(inodeWrapper, e);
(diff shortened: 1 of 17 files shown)


Wine93:
recheck

SeanHai:
recheck

SeanHai:
recheck

SeanHai:
recheck

curvefs/src/client/inode_cache_manager.h
```diff
@@ -168,8 +161,8 @@ class InodeCacheManagerImpl : public InodeCacheManager {
     std::map<uint64_t, std::shared_ptr<InodeWrapper>> dirtyMap_;
     curve::common::Mutex dirtyMapMutex_;
 
-    // inodeid to parent inodeid, may have more parent at hard link
-    std::map<uint64_t, std::list<uint64_t>> parentIdMap_;
+    // inodeid to parent inodeid
+    std::map<uint64_t, uint64_t> parentIdMap_;
```
baijiaruo, line 165: Does this only work if you turn it on?

baijiaruo (approved):

curvefs/src/client/inode_cache_manager.h
SeanHai: > Does this only work if you turn it on?

Yes, the relationship is needed when turn on the enableSumInDir flag.

curvefs/src/client/fuse_client.cpp
```diff
@@ -1297,10 +1301,6 @@ CURVEFS_ERROR FuseClient::FuseOpRelease(fuse_req_t req, fuse_ino_t ino,
         return ret;
     }
 
-    if (enableSumInDir_) {
```
cw123, line 1300: why not delete

curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -183,47 +213,36 @@ void InodeCacheManagerImpl::FlushInodeOnce() {
...
-    curve::common::LockGuard lg2(parentIdMapMutex_);
-    if (parentIdMap_.count(inodeId)) {
-        auto iter = std::find(parentIdMap_[inodeId].begin(),
-            parentIdMap_[inodeId].end(), parentId);
-        if (iter != parentIdMap_[inodeId].end()) {
-            parentIdMap_[inodeId].erase(iter);
-        }
+        LOG(WARNING) << "AddParent but exist, inodeId = " << inodeId
+                     << ", parentId = " << parentIdMap_[inodeId]
+                     << ", need2AddParentId = " << parentId;
```
cw123, line 218: check old parent and new parent are equal?

curvefs/src/client/fuse_client.cpp
SeanHai: > why not delete

Oplookup will not called when vfs cached the info of inode, so if delete the relationship when release maybe lost the parent.

curvefs/src/client/inode_cache_manager.cpp
SeanHai: > check old parent and new parent are equal?

fixed

SeanHai:
recheck

curvefs/src/client/inode_cache_manager.cpp
cw123: if (parentIdMap_.count(inodeId) && parentIdMap_[inodeId] != parentId) 
what happens here?

SeanHai: > if (parentIdMap_.count(inodeId) && parentIdMap_[inodeId] != parentId) what happens here?

It won't happen normally, just do anomaly detection.

cw123 (approved):

PR #1154, opened 2022-03-09
fix_fs_manager_ut

Cyber-SiKu:

 fs_mageger ut error:
C++ exception with description "Resource temporarily unavailable"

Issue Number: close #xxx 

Problem Summary:

due to not set asyncThreadNum and not mock void init(const
S3AdapterOption &), ut will call init function and create thread by uninit
value.

What's Changed:

How it Works:

### Check List


Cyber-SiKu:
recheck

ilixiaocui (approved):

xu-chaojie (approved):

Issue #1210, opened 2022-03-25
[request help] Curve v2.1 read error for release modle


Wine93:
# Problem

If we build project with relase modle for Curve `v2.1`, it will report error when reading data.

# Reproduce

* build project with release modle
```
$ make build only=//curvefs/src release=1
```

* deploy cluster and mount filesystem

```
$ curveadm deploy
...
```

* enter mounpoint
```
$ cd /path/to/mountpoint
```

* generate file
```
$ seq 1 100000000 > file
```

* calculate MD5
```
$ md5sum file
```

It will report error which is `Bad File Description`.

Cyber-SiKu:
https://github.com/opencurve/curve/pull/1103

Cyber-SiKu:
[pr](https://github.com/opencurve/curve/pull/1103) (has rebase master)
Triggered step:
1. fio write*4
2. fio randwrite*4
3. fio read *4
4. fio randread *4

all fio read report Bad file Description

but when remount some read wouldn't report  Bad file Description, but when run command again, still report.

cw123:
mount，fio write ok， fio read fail， remount， fio read ok， fio read fail

PR #1216, opened 2022-03-28
fix curvefs release read return Bad File Description

cw123:

Issue Number: #1210 

Problem Summary: fix curvefs release read return Bad File Description

What's Changed:

init fuse_bufvec struct before it use

### DIFF
--- a/curvefs/src/client/curve_fuse_op.cpp
+++ b/curvefs/src/client/curve_fuse_op.cpp
@@ -312,11 +312,7 @@ void FuseOpRead(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
         return;
     }
 
-    struct fuse_bufvec bufvec;
-    bufvec.count = 1;
-    bufvec.off = 0;
-    bufvec.idx = 0;
-    bufvec.buf[0].size = rSize;
+    struct fuse_bufvec bufvec = FUSE_BUFVEC_INIT(rSize);
     bufvec.buf[0].mem = buffer.get();
 
     fuse_reply_data(req, &bufvec, FUSE_BUF_SPLICE_MOVE);


cw123:
recheck

xu-chaojie (approved):

curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -312,7 +312,7 @@ void FuseOpRead(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
         return;
     }
 
-    struct fuse_bufvec bufvec;
+    struct fuse_bufvec bufvec = FUSE_BUFVEC_INIT(size);
```
h0hmj, line 315:
```
struct fuse_bufvec bufvec = FUSE_BUFVEC_INIT(rSize);
bufvec.buf[0].mem = buffer.get();
```
may be better

h0hmj:
```
    bufvec.count = 1;
    bufvec.off = 0;
    bufvec.idx = 0;
    bufvec.buf[0].size = rSize;
```
is unnecessary

h0hmj:
```
/* Initialize bufvec with a single buffer of given size */
#define FUSE_BUFVEC_INIT(size__)				\
	((struct fuse_bufvec) {					\
		/* .count= */ 1,				\
		/* .idx =  */ 0,				\
		/* .off =  */ 0,				\
		/* .buf =  */ { /* [0] = */ {			\
			/* .size =  */ (size__),		\
			/* .flags = */ (enum fuse_buf_flags) 0,	\
			/* .mem =   */ NULL,			\
			/* .fd =    */ -1,			\
			/* .pos =   */ 0,			\
		} }						\
	} )
```

https://github.com/libfuse/libfuse/blob/a6a01c4a76c4c8265f7d1f0a0a9b5340b15384ab/include/fuse_common.h#L770

h0hmj (approved):

PR #1217, opened 2022-03-29
fix curvefs release read return Bad File Description

cw123:

Issue Number: #1210

Problem Summary:

What's Changed:

How it Works:  init fuse_bufvec struct before it use

### Check List


cw123:
recheck

PR #1231, opened 2022-03-31
curvebs: support 512 aligned IO

### LINKED ISSUE #985: potential inconsistent data when map a clone volume by curve-nbd with 512 block-size
**Describe the bug (描述bug)**

for a clone volume, each chunk in ChunkServer has a bitmap, and each bit represents whether a 4k range has been written after clone, which means clone chunk can only receive 4k aligned write requests.

in this case, when map a clone volume by curve-nbd with 512 block-size, and user issues a 512 bytes write request, curve-client firstly reads the full 4k range data, and then paste 512 bytes into full 4k data, and send a 4k write request to chunkserver.

so, when user issues two non-overlapping 512 bytes within the same 4k range concurrently, one of the two requests' data may lost.

**To Reproduce (复现方法)**

I haven't reproduce this problem, but in theory it's reproducible, later I will construct scenarios to verify this problem.

wu-hanqing:

Issue Number: close #985 

Problem Summary:

What's Changed:

How it Works:

### Check List


wu-hanqing:
recheck

wu-hanqing:
recheck

wu-hanqing:
recheck

wu-hanqing:
recheck

wu-hanqing:
recheck

src/chunkserver/register.cpp
```diff
@@ -63,12 +63,14 @@ int Register::RegisterToMDS(ChunkServerMetadata *metadata) {
         req.set_externalip(ops_.chunkserverExternalIp);
     }
     req.set_port(ops_.chunkserverPort);
+    req.set_blocksize(ops_.blockSize);
```
ilixiaocui, line 66: Is this for verification on the mds side?

src/common/fast_align.h
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright 2015 Glen Joseph Fernandes
+(glenjofe@gmail.com)
```
ilixiaocui, line 3: Same as curvefs/src/common/fast_align.h?

src/common/namespace_define.h
```diff
@@ -62,6 +62,8 @@ const char CLONEINFOKEYEND[] = "13";
 const char DISCARDSEGMENTKEYPREFIX[] = "13";
 const char DISCARDSEGMENTKEYEND[] = "14";
 
+const char BLOCKSIZEKEY[] = "14blocksize";
```
ilixiaocui, line 65: Why not use "15blocksize" ?

src/tools/curve_format_main.cpp
```diff
@@ -175,6 +179,25 @@ int AllocateFiles(AllocateStruct* allocatestruct) {
...
+// so, check whether meta page size is enough for store bitmap
+// and other metadata
+static bool CheckChunkAndMetaPageSize() {
+    if (!curve::common::is_aligned(FLAGS_fileSize, FLAGS_blockSize)) {
+        LOG(ERROR) << "io alignment or file size is wrong";
+        return false;
+    }
+
+    auto bitmapBytes = FLAGS_fileSize / FLAGS_blockSize / 8;
+    if (bitmapBytes > FLAGS_metaPageSize * 2) {
```
ilixiaocui, line 193: If bitmapBytes > (FLAGS_metaPageSize-specified field), it should not work?
Here means if bitmapBytes = 4k, FLAGS_metaPageSize = 1k，that's ok？

src/common/namespace_define.h
wu-hanqing: fix

src/common/fast_align.h
wu-hanqing: fix, curvefs/src/common/fast_align.h is removed

conf/chunkserver.conf
```diff
@@ -12,10 +12,13 @@ global.external_subnet=127.0.0.0/24
 global.chunk_size=16777216
 # chunk 元数据页大小，一般4KB
 global.meta_page_size=4096
+# chunk's block size, IO requests must align with it, supported value is |512| and |4096|
+# it should consist with `block_size` in chunkfilepool.meta_path and `mds.volume.blockSize` in MDS's configurations
+# for clone chunk and snapshot chunk, it's also the minimum granularity that each bit represents
+# if set to |512|, we need 4096 bytes bitmap for each chunk, so meta_page_size should be 8192 or larger.
+global.block_size=4096
```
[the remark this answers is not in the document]
wu-hanqing: We have three verifications:
1. when a chunkserver starts, we will verify `global.block_size` in chunkserver's conf and chunkfilepool's metafile, if they aren't equal, chunkserver will quit.
2. then, if the chunkserver is newly deployed, and in the registration process, we will verify chunkserver's block size and block size recorded in MDS, if they aren't equal, chunkserver registration will fail.
3. MDS's conf also has a `block_size`, and when it starts, we will verify this item with block size recorded in etcd, if they aren't equal, mds will quit.

and for compatibility,
- if chunkfilepool's metafile doesn't have block size, we treat it as 4096. 
- if etcd doesn't block size, we will insert it according to the value in conf.

curve-ansible/roles/generate_config/templates/s3.conf.j2
```diff
@@ -25,3 +25,4 @@ s3.throttle.iopsWriteLimit={{ s3_throttle_iopsWriteLimit }}
 s3.throttle.bpsTotalMB= {{ s3_throttle_bpsTotalLimit }}
 s3.throttle.bpsReadMB= {{ s3_throttle_bpsReadLimit }}
 s3.throttle.bpsWriteMB= {{ s3_throttle_bpsWriteLimit }}
+s3.useVirtualAddressing=False
```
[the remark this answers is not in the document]
wu-hanqing: it was introduced by https://github.com/opencurve/curve/pull/1253

src/chunkserver/datastore/chunkserver_chunkfile.cpp
```diff
@@ -433,9 +438,9 @@ CSErrorCode CSChunkFile::Paste(const char * buf, off_t offset, size_t length) {
 
     // The request above must be pagesize aligned
     // the starting page index number of the paste area
-    uint32_t beginIndex = offset / pageSize_;
+    uint32_t beginIndex = offset / blockSize_;
```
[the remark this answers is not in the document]
wu-hanqing: fix

src/chunkserver/datastore/chunkserver_chunkfile.cpp
```diff
@@ -488,9 +494,9 @@ CSErrorCode CSChunkFile::Read(char * buf, off_t offset, size_t length) {
     if (isCloneChunk_) {
         // The request above must be pagesize aligned
         // the starting page index number of the paste area
-        uint32_t beginIndex = offset / pageSize_;
+        uint32_t beginIndex = offset / blockSize_;
```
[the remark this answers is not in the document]
wu-hanqing: fix

src/chunkserver/register.cpp
wu-hanqing: yep

src/chunkserver/config_info.h
```diff
@@ -140,6 +142,27 @@ struct ChunkServiceOptions {
     std::shared_ptr<InflightThrottle> inflightThrottle;
 };
 
+inline CopysetNodeOptions::CopysetNodeOptions()
```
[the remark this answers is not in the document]
wu-hanqing: I didn't modify it, just moved it from source file to header, because it's trivial.

nbd/src/util.cpp
```diff
@@ -343,22 +341,27 @@ int check_size_from_file(const std::string &path, uint64_t expected_size,
...
 
+static std::string sys_block_path(int nbd_index) {
+    return "/sys/block/nbd" + std::to_string(nbd_index);
+}
+
 int check_block_size(int nbd_index, uint64_t expected_size) {
-    std::string path = "/sys/block/nbd" + std::to_string(nbd_index)
-                     + "/queue/hw_sector_size";
+    std::string path =
+        sys_block_path(nbd_index) + "/queue/logical_block_size";
```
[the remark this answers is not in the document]
wu-hanqing: I'm not sure why we check `hw_sector_size` before, but what we set is block size when mapping the device,

https://github.com/opencurve/curve-nbd/blob/6231d3bc4954a5f2427148475cf39a2fdc4e3bad/src/NBDController.cpp#L50

src/tools/curve_format_main.cpp
wu-hanqing: fixed, I added a hard code value `4096 - 512` to indicate minimum size usage for encoding and storing source location for clone chunk, so that, the meta page size should be great than or equal to `3584 + bitmap bytes`

src/mds/server/mds.cpp
```diff
@@ -118,6 +122,9 @@ void MDS::StartCompaginLeader() {
 }
 
 void MDS::Init() {
+    LOG_IF(FATAL, !CheckOrInsertBlockSize(etcdClient_.get()))
```
xu-chaojie, line 125: Here Insert blocksize before  g_block_size is load from conf，Will cause a problem？

src/mds/server/mds.cpp
```diff
@@ -500,6 +507,12 @@ void MDS::InitCurveFSOptions(CurveFSOption *curveFSOptions) {
         "mds.curvefs.minFileLength", &curveFSOptions->minFileLength);
     conf_->GetValueFatalIfFail(
         "mds.curvefs.maxFileLength", &curveFSOptions->maxFileLength);
+    conf_->GetValueFatalIfFail("mds.curvefs.blockSize", &g_block_size);
```
xu-chaojie, line 510: why not define g_block_size in curveFSOptions instead of global variable

src/mds/topology/topology_service_manager.cpp
```diff
@@ -55,9 +55,21 @@ using ::curve::mds::copyset::Copyset;
 using ::curve::mds::copyset::CopysetConstrait;
 using ::curve::common::CopysetInfo;
 
+
 void TopologyServiceManager::RegistChunkServer(
     const ChunkServerRegistRequest *request,
     ChunkServerRegistResponse *response) {
+    if (request->has_blocksize()) {
```
xu-chaojie, line 62: Consider that chunkserver gets blocksize from mds instead of configuring blocksize individually and then verifying

src/tools/curve_format_main.cpp
```diff
@@ -53,10 +57,18 @@ DEFINE_uint32(fileSize,
...
               4 * 1024,
               "metapage size for every chunk");
 
+static bool ValidateMetaPageSize(const char* /*name*/, uint32_t metaPageSize) {
+    return is_aligned(metaPageSize, 4096);
+}
+
+DEFINE_validator(metaPageSize, &ValidateMetaPageSize);
+
+DEFINE_uint32(blockSize, 4096, "minimum io alignment supported");
```
xu-chaojie, line 70: Do we check blockSize and metaPageSize at the start of chunkserver ?

src/client/file_instance.cpp
```diff
@@ -106,6 +111,13 @@ void FileInstance::UnInitialize() {
 }
 
 int FileInstance::Read(char* buf, off_t offset, size_t length) {
+    if (!CheckIoAligned(offset, length, finfo_.blocksize)) {
```
xu-chaojie, line 114: should use finfo_ in metacache， the fileinfo_ in FileInstance  may be incomplete.

src/client/file_instance.cpp
```diff
@@ -117,13 +129,29 @@ int FileInstance::Write(const char* buf, off_t offset, size_t len) {
         DVLOG(9) << "open with read only, do not support write!";
         return -1;
     }
+
+    if (!CheckIoAligned(offset, len, finfo_.blocksize)) {
```
xu-chaojie, line 133: Ditto

src/client/file_instance.cpp
```diff
@@ -117,13 +129,29 @@ int FileInstance::Write(const char* buf, off_t offset, size_t len) {
...
+    }
+
     DLOG_EVERY_SECOND(INFO) << "begin write " << finfo_.fullPathName
                             << ", offset = " << offset
                             << ", len = " << len;
     return iomanager4file_.Write(buf, offset, len, mdsclient_.get());
 }
 
 int FileInstance::AioRead(CurveAioContext* aioctx, UserDataType dataType) {
+    if (!CheckIoAligned(aioctx->offset, aioctx->length, finfo_.blocksize)) {
```
xu-chaojie, line 146: Ditto

src/client/file_instance.cpp
```diff
@@ -135,6 +163,16 @@ int FileInstance::AioWrite(CurveAioContext* aioctx, UserDataType dataType) {
         DVLOG(9) << "open with read only, do not support write!";
         return -1;
     }
+
+    if (!CheckIoAligned(aioctx->offset, aioctx->length, finfo_.blocksize)) {
```
xu-chaojie, line 167: Ditto

Issue #1205, opened 2022-03-24
mdtest\ mkdir\ touch Create files not synchronized on other multi-mounted FUSE


YunhuiChen:
**Describe the bug (描述bug)**

![image](https://user-images.githubusercontent.com/67582932/159834982-aebcd1cb-8454-45a4-ae29-969d9ee906eb.png)
![image](https://user-images.githubusercontent.com/67582932/159835005-691e17ed-1e83-4882-a7bd-09f55244338c.png)

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
curve-mds:
curve-chunkserver:
curve-snapshotcloneserver:
curve-sdk:
nebd:
curve-nbd:

**Additional context/screenshots (更多上下文/截图)**

ilixiaocui:
Mount the same filesystem on another machine, the number of files in ls is correct
![image](https://user-images.githubusercontent.com/15689619/160320952-630eaa35-15f1-4899-b9c1-8d19773136f7.png)
![image](https://user-images.githubusercontent.com/15689619/160320965-64a37bab-a005-443b-9143-a5b055216d5d.png)

ls corresponds to the underlying request is ListDentry. The request log of the problem and the correct ListDentry are as follows：
`I 2022-03-24T15:38:27.438580+0800 2140650 metaserver_client.cpp:211] ListDentry done, request: poolId: 1
I 2022-03-24T15:38:27.438609+0800 2140650 dentry_cache_manager.cpp:120] ListDentry fsId = 1, parent = 1, last = , count = 65536, onlyDir = 0, ret = 0, part.size() = 12

I 2022-03-24T15:38:27.438580+0800 2140650 metaserver_client.cpp:211] ListDentry done, request: poolId: 1
copysetId: 1
partitionId: 1
fsId: 1
dirInodeId: 1
**txId: 0**
last: ""
count: 65536
onlyDir: false
response: statusCode: OK
dentrys {
fsId: 1
inodeId: 18260888
parentInodeId: 1
name: "000.log"
txId: 0
type: TYPE_S3
}`

ilixiaocui:
Solution: For operations related to dentry, we need to get the latest txid from mds.

PR #1250, opened 2022-04-06
curvefs/client: implement lease with mds

ilixiaocui:
Currently the lease only updates the txid of the partition which solves
the problem of invalidation of the txid cache in the case of multiple mounts.

Related bug issue: https://github.com/opencurve/curve/issues/1205

Issue Number:  #1205 

Problem Summary:

What's Changed:

How it Works:

### Check List


curvefs/proto/mds.proto
```diff
@@ -167,4 +177,7 @@ service MdsService {
     rpc DeleteFs(DeleteFsRequest) returns (DeleteFsResponse);
     rpc AllocateS3Chunk(AllocateS3ChunkRequest) returns (AllocateS3ChunkResponse);
     rpc ListClusterFsInfo (ListClusterFsInfoRequest) returns (ListClusterFsInfoResponse);
+
+    // client lease
+    rpc RefreshSession(RefreshSessionRequest) returns (RefreshSessionResponse);
```
SeanHai, line 182: Add this rpc interface in TopologyService maybe better, can deal request by topology_manager directly instead of pass by fs_manager.

curvefs/src/client/lease/lease_excutor.cpp
```diff
@@ -0,0 +1,97 @@
...
+    // refresh from mds
+    std::vector<PartitionTxId> latestTxIdList;
+    FSStatusCode ret = mdsCli_->RefreshSession(txIds, &latestTxIdList);
+    if (ret != FSStatusCode::OK) {
+        LOG(ERROR) << "LeaseExecutor refresh session fail, ret = " << ret;
+        return true;
+    }
+
+    // update to metacache
+    std::for_each(latestTxIdList.begin(), latestTxIdList.end(),
```
SeanHai, line 90: The issues/1205 is caused by the inconsistent of txid on multi mountpoints. Are there still an inconsistent perid in this way?

curvefs/proto/mds.proto
Wine93: > Add this rpc interface in TopologyService maybe better, can deal request by topology_manager directly instead of pass by fs_manager.

agree!

ilixiaocui: I've considered between mds and topology before.The information updated by Refreshsession currently is the txid of the partition, but it has at least one automatic umount responsibility, and there may be others later, so it is added to the unified entry.

curvefs/src/client/lease/lease_excutor.cpp
```diff
@@ -0,0 +1,97 @@
...
+    if (txIds.empty()) {
+        return true;
+    }
+
+    // refresh from mds
+    std::vector<PartitionTxId> latestTxIdList;
+    FSStatusCode ret = mdsCli_->RefreshSession(txIds, &latestTxIdList);
+    if (ret != FSStatusCode::OK) {
+        LOG(ERROR) << "LeaseExecutor refresh session fail, ret = " << ret;
+        return true;
```
Wine93, line 86: should we return `true` even if invoke `RefreshSession` failed?

curvefs/src/client/lease/lease_excutor.cpp
ilixiaocui: Yeap. The txid of the current fix partition is updated with a delay, and the old data will be obtained before the update. but this is acceptable，currently provides eventually consistent semantics.
The following options have been considered：
1. Go to mds every time to get the latest partition txid. Each dentry-related request will have one more rpc, 
This is not necessary under the semantics of eventual consistency.
2. The client registers with etcd, monitors the change of partition txid, and will notify in reverse if there is a change.This approach is also delayed update, and the client has another component that needs to interact.
3. So I chose the session method.

Wine93 (approved):
LGTM!

Wine93:
BTW: why the failover test cost about 2 hours?

ilixiaocui:
> BTW: why the failover test cost about 2 hours?
This problem is being dealt with. The current master branch also has this problem. Yesterday, ci was triggered and it ran for more than 3 hours.

SeanHai (approved):

PR #1254, opened 2022-04-08
backport-2.1: add s3.useVirtualAddressing config

h0hmj:

Issue Number: close #xxx 

Problem Summary:

We do not support virtual host.

What's Changed:

Support both virtual host and path style

How it Works:

Add a config, init client with config

### Check List

### DIFF
--- a/src/common/s3_adapter.cpp
+++ b/src/common/s3_adapter.cpp
@@ -69,41 +69,11 @@ Aws::String GetObjectRequestRange(uint64_t offset, uint64_t len) {
 }  // namespace
 
 void InitS3AdaptorOption(Configuration* conf, S3AdapterOption* s3Opt) {
-    LOG_IF(FATAL, !conf->GetIntValue("s3.loglevel", &s3Opt->loglevel));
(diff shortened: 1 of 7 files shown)


src/common/s3_adapter.cpp
```diff
@@ -132,6 +132,8 @@ void InitS3AdaptorOptionExceptS3InfoOption(Configuration* conf,
         &s3Opt->bpsReadMB));
     LOG_IF(FATAL, !conf->GetUInt64Value("s3.throttle.bpsWriteMB",
         &s3Opt->bpsWriteMB));
+    LOG_IF(FATAL, !conf->GetBoolValue("s3.useVirtualAddressing",
```
aspirer, line 135: how about func InitS3AdaptorOption? need add this new config?

h0hmj: InitS3AdaptorOption will call InitS3AdaptorOptionExceptS3InfoOption

aspirer (approved):

Issue #1230, opened 2022-03-31
When testing vdbench, the kernel prints an error, and io gets stuck for a few minutes


YunhuiChen:
**Describe the bug (描述bug)**
![image](https://user-images.githubusercontent.com/67582932/160959605-5efb9e9f-2a9c-4f45-a8d4-339e58bc6674.png)

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
curve-mds:
curve-chunkserver:
curve-snapshotcloneserver:
curve-sdk:
nebd:
curve-nbd:

**Additional context/screenshots (更多上下文/截图)**

SeanHai:
It may be caused by the existence of unreturned requests (it did not recur later), and now related metrics are added.

PR #1268, opened 2022-04-12
curvefs: add client op metric

SeanHai:

Issue Number: close #xxx 
#1230 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/curve_fuse_op.cpp
+++ b/curvefs/src/client/curve_fuse_op.cpp
@@ -36,6 +36,8 @@
 #include "curvefs/src/client/fuse_s3_client.h"
 #include "curvefs/src/client/rpcclient/mds_client.h"
 #include "curvefs/src/client/rpcclient/base_client.h"
+#include "curvefs/src/client/metric/client_metric.h"
+#include "curvefs/src/common/metric_utils.h"
 
 using ::curve::common::Configuration;
 using ::curvefs::client::CURVEFS_ERROR;
@@ -46,9 +48,13 @@ using ::curvefs::client::common::FuseClientOption;
 using ::curvefs::client::common::MAXXATTRLENGTH;
 using ::curvefs::client::rpcclient::MdsClientImpl;
 using ::curvefs::client::rpcclient::MDSBaseClient;
+using ::curvefs::client::metric::ClientOpMetric;
+using ::curvefs::common::LatencyUpdater;
+using ::curvefs::client::metric::InflightGuard;
 
 static FuseClient *g_ClientInstance = nullptr;
 static FuseClientOption *g_fuseClientOption = nullptr;
+static ClientOpMetric* g_clientOpMetric = nullptr;
 
 DECLARE_int32(v);
 
@@ -119,6 +125,8 @@ int InitGlog(const char *confPath, const char *argv0) {
 
 int InitFuseClient(const char *confPath, const char* fsName,
     const char *fsType) {
+    g_clientOpMetric = new ClientOpMetric();
+
     Configuration conf;
     conf.SetConfigPath(confPath);
     if (!conf.LoadConfig()) {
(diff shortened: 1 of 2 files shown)


curvefs/src/client/metric/client_metric.h
```diff
@@ -103,6 +103,84 @@ struct MetaServerClientMetric {
...
+    ~LatencyUpdater() {
+        timer.stop();
+        (*recorder) << timer.u_elapsed();
+    }
+
+    bvar::LatencyRecorder* recorder;
+    butil::Timer timer;
+};
+
+struct OpMetric : public InterfaceMetric {
```
wu-hanqing, line 121: I think there're only four types of metric that need to record

1. qps
2. latency
3. eps or error count
4. inflight count

so `OpMetric` is not suitable for this work, because `InterfaceMetric` has more metrics.

and, another tip is `LatencyRecorder` is an aggregate metric, it already supports qps by default.  https://github.com/apache/incubator-brpc/blob/0148f46e3b8b080808d031ba425555392dfac894/src/bvar/latency_recorder.h#L49

so, please redesign this class.

curvefs/src/client/metric/client_metric.h
```diff
@@ -103,6 +103,84 @@ struct MetaServerClientMetric {
           createPartition(prefix, "createPartition") {}
 };
 
+struct LatencyUpdater {
```
wu-hanqing, line 106: curvefs/src/common/metric_utils.h already has a LatencyUpdater, you can use it directly

curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -225,19 +232,29 @@ void FuseReplyErrByErrCode(fuse_req_t req, CURVEFS_ERROR errcode) {
 }
 
 void FuseOpLookup(fuse_req_t req, fuse_ino_t parent, const char *name) {
+    clientOpMetric->opLookup.qps.count << 1;
+    clientOpMetric->opLookup.inflightOpNum << 1;
```
wu-hanqing, line 236: maybe, we need an `InflightGuard` and implementation like this


```cpp
struct InflightGuard {
   InflightGuard(bvar::Adder<int64_t>* inflight) : inflight_(inflight) {
        (*inflight_) << 1;
   }

   ~InflightGuard() {
        (*inflight_) << -1;
   }

   bvar::Adder<int64_t>* inflight_;
};

```

curvefs/src/client/metric/client_metric.h
SeanHai: fixed

curvefs/src/client/metric/client_metric.h
SeanHai: fixed

curvefs/src/client/curve_fuse_op.cpp
SeanHai: fixed

wu-hanqing (approved):

curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -46,9 +48,13 @@ using ::curvefs::client::common::FuseClientOption;
 using ::curvefs::client::common::MAXXATTRLENGTH;
 using ::curvefs::client::rpcclient::MdsClientImpl;
 using ::curvefs::client::rpcclient::MDSBaseClient;
+using ::curvefs::client::metric::ClientOpMetric;
+using ::curvefs::common::LatencyUpdater;
+using ::curvefs::client::metric::InflightGuard;
 
 static FuseClient *g_ClientInstance = nullptr;
 static FuseClientOption *g_fuseClientOption = nullptr;
+static ClientOpMetric* clientOpMetric = nullptr;
```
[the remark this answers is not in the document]
SeanHai: fixed

Cyber-SiKu (approved):

PR #1269, opened 2022-04-12
curvefs/metaserver:fix s3compaction stackoverflow

### LINKED ISSUE #1263: curvefs/metaserver: stack overflow when doing s3compaction
**Describe the bug (描述bug)**

https://github.com/opencurve/curve/blob/79e09d35769a44586d0fe19f2d79814708a4f62e/curvefs/src/metaserver/s3compact_wq_impl.cpp#L147

is recursive, may cause stack overflow, and metaserver crash

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

make it iterative

**Versions (各种版本)**
OS:
Compiler:
curvefs-metaserver: v2.1

h0hmj:

Issue Number: close #1263 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/metaserver/s3compact_wq_impl.cpp
+++ b/curvefs/src/metaserver/s3compact_wq_impl.cpp
@@ -25,6 +25,7 @@
 #include <algorithm>
 #include <deque>
 #include <list>
+#include <map>
 #include <memory>
 #include <string>
 #include <unordered_map>
@@ -33,8 +34,8 @@
 
 #include "absl/cleanup/cleanup.h"
 #include "curvefs/src/common/s3util.h"
-#include "curvefs/src/metaserver/copyset/meta_operator.h"
 #include "curvefs/src/metaserver/copyset/copyset_node_manager.h"
+#include "curvefs/src/metaserver/copyset/meta_operator.h"
 
 using curve::common::Configuration;
 using curve::common::InitS3AdaptorOptionExceptS3InfoOption;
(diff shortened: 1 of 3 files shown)


curvefs/src/metaserver/s3compact_wq_impl.cpp
```diff
@@ -139,112 +140,77 @@ void S3CompactWorkQueueImpl::DeleteObjs(const std::vector<std::string>& objs,
...
+        const uint64_t begin = info.offset();
+        const uint64_t end = info.offset() + info.len() - 1;
+        for (auto it = freeList.begin(); it != freeList.end();) {
+            auto n = std::next(it);
+            // overlap means we can take this free
+            auto b = it->first;
+            auto e = it->second;
+            if (begin <= b) {
+                if (end < b) {
+                    return;
```
wu-hanqing, line 166: should you check overlap with the next one in freeList instead of return directly?

curvefs/src/metaserver/s3compact_wq_impl.cpp
```diff
@@ -139,112 +140,77 @@ void S3CompactWorkQueueImpl::DeleteObjs(const std::vector<std::string>& objs,
...
-        }
-        curr = next;
+    for (auto i = s3chunkinfolist.s3chunks_size() - 1; i >= 0; i--) {
+        if (freeList.empty()) break;
+        fill(i);
+    }
+
+    std::list<struct S3CompactWorkQueueImpl::Node> validList;
+    for (const auto& v : used) {
+        auto info = s3chunkinfolist.s3chunks(v.second.second);
```
wu-hanqing, line 208: const auto& info

curvefs/src/metaserver/s3compact_wq_impl.cpp
```diff
@@ -139,112 +140,77 @@ void S3CompactWorkQueueImpl::DeleteObjs(const std::vector<std::string>& objs,
...
-                    addToValidList(std::move(n));
+                                       uint64_t inodeLen, uint64_t index,
+                                       uint64_t chunkSize) {
+    std::list<std::pair<uint64_t, uint64_t>> freeList;  // [begin, end]
+    freeList.emplace_back(chunkSize * index,
+                          std::min(chunkSize * (index + 1) - 1,
+                                   inodeLen - 1));  // start with full chunk
+    std::map<uint64_t, std::pair<uint64_t, uint64_t>>
+        used;  // begin -> pair(end, i)
+    auto fill = [&](uint64_t i) {
```
wu-hanqing, line 151: it's better to make this a util function and do more testing

curvefs/src/metaserver/s3compact_wq_impl.cpp
h0hmj: item in freeList is ordered
which means next item's begin always bigger than this one
if end < b, end will < all b next, so just return

curvefs/src/metaserver/s3compact_wq_impl.cpp
h0hmj: current test cases have covered all branchs in this lambda function

curvefs/src/metaserver/s3compact_wq_impl.cpp
h0hmj: fix

SeanHai:
recheck

SeanHai (approved):

wu-hanqing (approved):

PR #1288, opened 2022-04-18
curvefs: adapt to curveadm to support curvebs as data backend

wu-hanqing:

Issue Number: close #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/conf/client.conf
+++ b/conf/client.conf
@@ -129,6 +129,9 @@ global.fileIOSplitMaxSizeKB=64
 #
 ################# log相关配置 ###############
 #
+# enable logging or not
+global.logging.enable=True
+#
 # log等级 INFO=0/WARNING=1/ERROR=2/FATAL=3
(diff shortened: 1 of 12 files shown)


wu-hanqing:
recheck

conf/client.conf
```diff
@@ -129,6 +129,9 @@ global.fileIOSplitMaxSizeKB=64
 #
 ################# log相关配置 ###############
 #
+# enable logging or not
+global.logging.enable=True
```
SeanHai, line 133: Why need this configuration item？

wu-hanqing: because in previous version, curve-client will always init glog in which it will set a different logging dir, then curve-fuse's logging destination will be affected.

SeanHai (approved):

Wine93 (approved):
LGTM!

PR #1298, opened 2022-04-19
curvefs: add applied index in client read rpc requests

SeanHai:

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/curve_fuse_op.cpp
+++ b/curvefs/src/client/curve_fuse_op.cpp
@@ -227,6 +227,9 @@ void FuseReplyErrByErrCode(fuse_req_t req, CURVEFS_ERROR errcode) {
     case CURVEFS_ERROR::OUT_OF_RANGE:
         fuse_reply_err(req, ERANGE);
         break;
+    case CURVEFS_ERROR::NODATA:
+        fuse_reply_err(req, ENODATA);
+        break;
     default:
         fuse_reply_err(req, EIO);
         break;
@@ -263,20 +266,17 @@ void FuseOpGetXattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     size_t size) {
     InflightGuard guard(&g_clientOpMetric->opGetXattr.inflightOpNum);
     LatencyUpdater updater(&g_clientOpMetric->opGetXattr.latency);
-    std::unique_ptr<char[]> buf(new char[MAXXATTRLENGTH]);
-    std::memset(buf.get(), 0, MAXXATTRLENGTH);
-    CURVEFS_ERROR ret = g_ClientInstance->FuseOpGetXattr(req, ino, name,
-                                                         buf.get(), size);
-    if (ret != CURVEFS_ERROR::OK && ret != CURVEFS_ERROR::NODATA) {
-        g_clientOpMetric->opGetXattr.ecount << 1;
-        FuseReplyErrByErrCode(req, ret);
-        return;
-    }
-
+    char buf[MAXXATTRLENGTH] = {0};
     if (size == 0) {
-        fuse_reply_xattr(req, strlen(buf.get()));
+        fuse_reply_xattr(req, MAXXATTRLENGTH);
     } else {
-        fuse_reply_buf(req, buf.get(), strlen(buf.get()));
+        CURVEFS_ERROR ret = g_ClientInstance->FuseOpGetXattr(req, ino, name,
+                                                             buf, size);
+        if (ret != CURVEFS_ERROR::OK) {
+            FuseReplyErrByErrCode(req, ret);
+            return;
+        }
+        fuse_reply_buf(req, buf, strlen(buf));
     }
 }
 
--- a/curvefs/src/client/rpcclient/metaserver_client.h
+++ b/curvefs/src/client/rpcclient/metaserver_client.h
@@ -29,6 +29,7 @@
 #include <vector>
 #include <set>
 #include <unordered_map>
+#include <utility>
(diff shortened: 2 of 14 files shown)


curvefs/src/metaserver/copyset/meta_operator.cpp
```diff
@@ -224,7 +230,7 @@ void GetOrModifyS3ChunkInfoOperator::OnApply(int64_t index,
         TYPE##Response response;                                             \
         auto status = node_->GetMetaStore()->TYPE(                           \
             static_cast<const TYPE##Request*>(request_), &response);         \
-        node_->GetMetric()->OnOperatorComplete(                              \
+        node_->GetMetric()->OnOperatorCompleteFromLog(                       \
```
wu-hanqing, line 233: you forget `GetOrModifyS3ChunkInfoOperator::OnApplyFromLog`

curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -265,17 +268,15 @@ void FuseOpGetXattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
-        g_clientOpMetric->opGetXattr.ecount << 1;
-        FuseReplyErrByErrCode(req, ret);
-        return;
-    }
-
     if (size == 0) {
-        fuse_reply_xattr(req, strlen(buf.get()));
+        fuse_reply_xattr(req, MAXXATTRLENGTH);
     } else {
+        CURVEFS_ERROR ret = g_ClientInstance->FuseOpGetXattr(req, ino, name,
```
wu-hanqing, line 274: there's no need to allocated `buf` on heap, `char buf[MAXXATTRLENGTH] = {0}` is enough.

wu-hanqing (commented):
the commit message might be not very suitable, you can rename it like performance optimization and add what you did in the body.

curvefs/src/client/curve_fuse_op.cpp
SeanHai: > there's no need to allocated `buf` on heap, `char buf[MAXXATTRLENGTH] = {0}` is enough.

fixed

curvefs/src/metaserver/copyset/meta_operator.cpp
SeanHai: > you forget `GetOrModifyS3ChunkInfoOperator::OnApplyFromLog`

fixed

curvefs/src/client/rpcclient/metaserver_client.h
```diff
@@ -37,8 +38,10 @@
 #include "curvefs/src/client/rpcclient/task_excutor.h"
 #include "curvefs/src/client/metric/client_metric.h"
 #include "curvefs/src/common/rpc_stream.h"
+#include "curvefs/src/common/metric_utils.h"
 
 using ::curvefs::client::metric::MetaServerClientMetric;
+using ::curvefs::common::LatencyUpdater;
```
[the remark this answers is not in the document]
SeanHai: > you can put the `using` and `include` in the source file, because you don't directly use the header file.
> 
> the side effect is any other source files that include the header file will implicitly include the newly added header file, and this will slow down compilation time

fixed

SeanHai:
> the commit message might be not very suitable, you can rename it like performance optimization and add what you did in the body.

ok

wu-hanqing (approved):

curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -263,20 +266,17 @@ void FuseOpGetXattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
-    std::memset(buf.get(), 0, MAXXATTRLENGTH);
-    CURVEFS_ERROR ret = g_ClientInstance->FuseOpGetXattr(req, ino, name,
-                                                         buf.get(), size);
-    if (ret != CURVEFS_ERROR::OK && ret != CURVEFS_ERROR::NODATA) {
-        g_clientOpMetric->opGetXattr.ecount << 1;
-        FuseReplyErrByErrCode(req, ret);
-        return;
-    }
-
+    char buf[MAXXATTRLENGTH] = {0};
```
xu-chaojie, line 269: use std::memset to clear buf,  = {0} is not work.

curvefs/src/client/dentry_cache_manager.cpp
```diff
@@ -155,6 +155,11 @@ CURVEFS_ERROR DentryCacheManagerImpl::ListDentry(uint64_t parent,
         }
     } while (perceed);
 
+    for (const auto &it : *dentryList) {
+        std::string key = GetDentryCacheKey(it.parentinodeid(), it.name());
+        dCache_->Put(key, it);
```
xu-chaojie, line 160: I don't think it is a good idea, putting all dentry of ListDentry in dCache will result in lower cache hit rate.

curvefs/src/client/curve_fuse_op.cpp
SeanHai: = {0} works and test a demo.

curvefs/src/client/dentry_cache_manager.cpp
SeanHai: Whether it will reduce the cache hit rate is not easy to say, not modify this time and can do some test later.

xu-chaojie (approved):

SeanHai:
recheck

Issue #1304, opened 2022-04-20
curvefs/metaserver: cannot get a full copy of one inode


h0hmj:
**Describe the bug (描述bug)**

if we want to get a copy of a inode,
we need GetInode and PaddingInodeS3ChunkInfo two rpc from InodeManager.

rpc1 -> GetInode: len=1 (s3chunkinfolist: slice 1)
changes -> some write, s3chunkinfolist: silce 1, slice 2
rpc2 -> PaddingInodeS3ChunkInfo: s3chunkinfolist: slice 1, slice 2

so we get inode, len=1 and s3chunkinfolist: slice1, slice2, which is wrong

if our s3compact threshold is 1, it will be compact to len=1, s3chunkinfolist: slice1
slice2 is regarded as useless data maybe from truncate smalle, and will be deleted.

inode will be updated to len=2, s3chunkinfolist: slice 1; we lose our data.

So we need a method to get a full copy of a inode when needed.

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
commit: c783e83dde3eb6a292bc6dbc659c6461cfcea5c0

**Additional context/screenshots (更多上下文/截图)**

Issue #1049, opened 2022-01-21
s3compact may not delete useless data after truncate smaller


h0hmj:
**Describe the bug (描述bug)**

s3compact threshold = 5
chunkindex 0: [0,1,2,3]
chunkindex 1: [5,6]

truncate -s 1
chunkindex 0: [0,1,2,3]
chunkindex 1: [5,6]

only chunkid 0 is valid.
len=4 and 2 smaller than 5, will not get s3compaction to delete invalid ones.

**Expected behavior (期望行为)**

s3compact will traverse and delete them

PR #1306, opened 2022-04-20
curvefs/metaserver: s3compat support "truncate smaller" correctly

h0hmj:
s3compact now will process two kinds of chunk
1. the whole chunk is useless; chunkSize * chunkIndex > inodeLen-1
2. part of the chunk is useless; some info in list off+len-1 > inodeLen -1

i wll delete these useless data


h0hmj:
solve #1049 
affected by #1304 which will make input of s3compact is wrong
affected by #1308 which will make metadata deletion from s3compact be dropped

wu-hanqing (approved):

Wine93 (approved):
LGTM!

PR #1322, opened 2022-04-21
curvefs/client:change log level on fly

Cyber-SiKu:

Issue Number: #1321 

Problem Summary:

Use vlog_level to dynamically modify the client's vlog level.

What's Changed:

Set the validator to the flag vlog_level, and only modify the log level
by assigning it to FLAG_v when the vlog_level value is legal.

How it Works:

You can modify the vlog level to 0 using:
curl -s http://127.0.0.1:9000/flags/vlog_level?setvalue=0

### Check List

### DIFF
--- a/curvefs/src/client/curve_fuse_op.cpp
+++ b/curvefs/src/client/curve_fuse_op.cpp
@@ -58,6 +58,21 @@ static ClientOpMetric* g_clientOpMetric = nullptr;
 
 DECLARE_int32(v);
 
+/**
+ * use vlog_level to set vlog level on the fly
+ * When vlog_level is set, CheckVLogLevel is called to check the validity of the
+ * value. Dynamically modify the vlog level by setting FLAG_v in CheckVLogLevel.
+ *
+ * You can modify the vlog level to 0 using:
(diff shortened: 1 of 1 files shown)


Cyber-SiKu:
Screenshot_20220421_105659

curvefs/src/client/curve_fuse_op.cpp
```diff
@@ -58,6 +58,16 @@ static ClientOpMetric* g_clientOpMetric = nullptr;
...
 
+DEFINE_int32(vlog_level, 0, "set v log level");
+static bool CheckVLogLevel(const char*, int32_t value) {
+    if (value >= 0 && value <= 9) {
+        FLAGS_v = value;
+        return true;
+    }
+    return false;
+}
+DEFINE_validator(vlog_level, CheckVLogLevel);
```
SeanHai, line 69: Add comments to explain the logic of the modification.

Cyber-SiKu: > Add comments to explain the logic of the modification.

fix

SeanHai (approved):

wu-hanqing (approved):

PR #1328, opened 2022-04-21
curvefs: add Owner attr to FsInfo

h0hmj:
1. Add Owner to FsInfo
2. Add Owner and Capacity to CreateFsRequest
3. Add -user (deafult: anonymous) and -capacity (default: 100GB) to curvefs_tool create-fs


YunhuiChen:
recheck

curvefs/proto/mds.proto
```diff
@@ -95,6 +96,8 @@ message CreateFsRequest {
     required common.FSType fsType = 3;
     required FsDetail fsDetail = 4;
     required bool enableSumInDir = 5;
+    required string owner = 6;
+    required uint64 capacity = 7;
```
SeanHai, line 100: This capacity seems not works after read the pr. How to limit the capacity of a fs?

curvefs/proto/mds.proto
```diff
@@ -82,6 +82,7 @@ message FsInfo {
     required common.FSType fsType = 9;
     required FsDetail detail = 10;
     required bool enableSumInDir = 11;
+    required string owner = 12;
```
wu-hanqing, line 85: new fields should be `optional`, because when you have an older version client which doesn't have this fields and newer version mds, then client RPC to MDS will always fail

curvefs/proto/mds.proto
wu-hanqing, line 100: ditto

curvefs/src/mds/fs_info_wrapper.h
```diff
@@ -55,6 +55,33 @@ class FsInfoWrapper {
         Swap(other);
     }
 
+    FsInfoWrapper(const ::curvefs::mds::CreateFsRequest* request, uint64_t fsId,
+                  uint64_t rootInodeId)
+        : FsInfoWrapper() {
```
wu-hanqing, line 60: `: FsInfoWrapper` is redundant

curvefs/src/mds/fs_manager.cpp
```diff
@@ -169,9 +169,14 @@ void FsManager::BackEndFunc() {
...
-FSStatusCode FsManager::CreateFs(const std::string& fsName, FSType fsType,
-                                 uint64_t blockSize, bool enableSumInDir,
-                                 const FsDetail& detail, FsInfo* fsInfo) {
+FSStatusCode FsManager::CreateFs(const ::curvefs::mds::CreateFsRequest* request,
+                                 FsInfo* fsInfo) {
+    std::string fsName = request->fsname();
+    uint64_t blockSize = request->blocksize();
+    FSType fsType = request->fstype();
+    bool enableSumInDir = request->enablesumindir();
+    FsDetail detail = request->fsdetail();
```
wu-hanqing, line 178: mark these variables as const reference to avoid unnecessary copy

curvefs/proto/mds.proto
h0hmj: this commit just adds capacity and owner attr, not how to use them. 
i will call "create-fs --user xx --capacity xx" when develop curvefs csi.

quota limit and user mount/r/w restriction are follow-ups

curvefs/proto/mds.proto
h0hmj: if we have a stable version, we need to consider the compatibility, but now all are under development.
owner and capacity are all required for resource limitation and fs management.

curvefs/src/mds/fs_info_wrapper.h
h0hmj: fix

curvefs/src/mds/fs_manager.cpp
h0hmj: fix

curvefs/src/mds/fs_info_wrapper.h
```diff
@@ -92,6 +127,14 @@ class FsInfoWrapper {
         return fsInfo_.blocksize();
     }
 
+    uint64_t GetCapacity() const {
+        return fsInfo_.capacity();
```
SeanHai, line 131: Maybe need 'curve::common::UniqueLock lg(mtx_)' to protect the fsInfo_

curvefs/src/mds/fs_info_wrapper.h
```diff
@@ -92,6 +127,14 @@ class FsInfoWrapper {
...
+    }
+
+    std::string GetOwner() const {
+        return fsInfo_.owner();
```
SeanHai, line 135: Ditto

curvefs/src/mds/fs_info_wrapper.h
```diff
@@ -72,6 +99,14 @@ class FsInfoWrapper {
         fsInfo_.set_fsname(name);
     }
 
+    void SetCapacity(uint64_t capacity) {
+        fsInfo_.set_capacity(capacity);
```
SeanHai, line 103: Need set dirty_ to true?

curvefs/src/mds/fs_info_wrapper.h
```diff
@@ -72,6 +99,14 @@ class FsInfoWrapper {
...
+    }
+
+    void SetOwner(const std::string& owner) {
+        fsInfo_.set_owner(owner);
```
SeanHai, line 107: Ditto

SeanHai (approved):

wu-hanqing (approved):

PR #1329, opened 2022-04-22
curvefs: prefetch inodes when readdir to improve meta data performance

SeanHai:
prefetch inode when readdir through a asynchronous rpc request

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/conf/client.conf
+++ b/curvefs/conf/client.conf
@@ -52,7 +52,10 @@ excutorOpt.maxRPCTimeoutMS=8000
 excutorOpt.maxRetrySleepIntervalUS=8000000
 excutorOpt.minRetryTimesForceTimeoutBackoff=5
 excutorOpt.maxRetryTimesBeforeConsiderSuspend=20
-excutorOpt.batchLimit=10000
+# batch limit of get inode attr and xattr
+excutorOpt.batchInodeAttrLimit=10000
+# batch limit of get inode
+excutorOpt.batchInodeLimit=100
 
 #### spaceserver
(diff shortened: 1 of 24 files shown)


curvefs/conf/client.conf
```diff
@@ -52,7 +52,10 @@ excutorOpt.maxRPCTimeoutMS=8000
 excutorOpt.maxRetrySleepIntervalUS=8000000
 excutorOpt.minRetryTimesForceTimeoutBackoff=5
 excutorOpt.maxRetryTimesBeforeConsiderSuspend=20
-excutorOpt.batchLimit=10000
+# batch limit of get inode attr and xattr
+excutorOpt.batchInodeAttrLimit=10000
```
xu-chaojie, line 56: excutorOpt is common， not only for inode， this config option seems not belong to the excutor freamwork

curvefs/src/client/fuse_client.cpp
```diff
@@ -634,6 +642,13 @@ CURVEFS_ERROR FuseClient::FuseOpReadDir(fuse_req_t req, fuse_ino_t ino,
         *buffer = nullptr;
         *rSize = 0;
     }
+
+    VLOG(1) << "batch get inode size = " << inodeIds.size();
+    auto rt = inodeManager_->BatchGetInodeAsync(&inodeIds);
```
xu-chaojie, line 647: Use another attr cache to avoid the corruption of inode lru cache， and use BatchGetInodeAttr also.

PR #1330, opened 2022-04-22
curvefs/metaserver: fixed sseek so slow which caused by

Wine93:
What's Changed
---

* fixed sseek so slow which caused by unusing rocksdb's table prefix bloom filter
* speed up getting inode's s3chunkinfo by padding inode's s3chunkinfo which small enought in GetInode RPC response

### DIFF
--- a/curvefs/proto/metaserver.proto
+++ b/curvefs/proto/metaserver.proto
@@ -46,6 +46,7 @@ enum MetaStatusCode {
     PARSE_FROM_STRING_FAILED = 23;
     STORAGE_INTERNAL_ERROR = 24;
     RPC_STREAM_ERROR = 25;
+    INODE_S3_META_TOO_LARGE = 26;
 }
 
 // dentry interface
@@ -148,6 +149,7 @@ message GetInodeRequest {
     required uint32 fsId = 4;
     required uint64 inodeId = 5;
     optional uint64 appliedIndex = 6;
+    optional bool supportStreaming = 7;  // for backward compatibility
 }
 
 enum FsFileType {
@@ -212,6 +214,7 @@ message GetInodeResponse {
     required MetaStatusCode statusCode = 1;
     optional Inode inode = 2;
     optional uint64 appliedIndex = 3;
+    optional bool streaming = 4;
 }
 
 message CreateInodeRequest {
@@ -335,6 +338,7 @@ message GetOrModifyS3ChunkInfoRequest {
     required bool returnS3ChunkInfoMap = 8;
     optional bool fromS3Compaction = 9;
     // todo: we only need a bit flag to indicate a lot of bool
+    optional bool supportStreaming = 10;  // for backward compatibility
 }
 
 message GetOrModifyS3ChunkInfoResponse {
--- a/curvefs/src/metaserver/inode_storage.h
+++ b/curvefs/src/metaserver/inode_storage.h
@@ -45,12 +45,14 @@ using ::curvefs::metaserver::storage::Status;
 using ::curvefs::metaserver::storage::Iterator;
 using ::curvefs::metaserver::storage::KVStorage;
 using ::curvefs::metaserver::storage::StorageTransaction;
+using ::curvefs::metaserver::storage::Hash;
 
 namespace curvefs {
 namespace metaserver {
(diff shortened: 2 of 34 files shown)


curvefs/proto/metaserver.proto
```diff
@@ -206,6 +208,7 @@ message Inode {
     optional uint32 openmpcount = 20; // openmpcount mount points had the file open
     map<string, string> xattr = 21;
     repeated uint64 parent = 22;
+    optional bool streaming = 23;
```
SeanHai, line 211: The streaming is not a attribute of Inode and I think put it in getInodeResponse is better.

curvefs/src/metaserver/inode_storage.h
```diff
@@ -138,19 +141,45 @@ class InodeStorage {
...
+        size_t key = InodeS3MetaSizeKey(fsId, inodeId);
+        uint64_t size = inodeS3MetaSize_[key] + size4add;
+        if (size < size4del) {
+            return false;
+        }
+        inodeS3MetaSize_[key] = size - size4del;
+        return true;
+    }
+
+    uint64_t GetInodeS3MetaSize(uint32_t fsId, uint64_t inodeId) {
```
SeanHai, line 178: Why not get sChunkInfoList size directly?

curvefs/proto/metaserver.proto
Wine93: > The streaming is not a attribute of Inode and I think put it in getInodeResponse is better.

fixed.

curvefs/src/metaserver/inode_storage.h
Wine93: > Why not get sChunkInfoList size directly?

actually, we had store `s3chunkinfo` by alone, so we can't get it from inode by directly.

curvefs/src/metaserver/storage/storage.h
```diff
@@ -110,6 +110,8 @@ class KVStorage : public BaseStorage {
 
     virtual bool GetStatistics(StorageStatistics* Statistics) = 0;
 
+    virtual StorageOptions GetStorageOptions() = 0;
```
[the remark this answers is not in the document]
Wine93: > `virtual StorageOptions GetStorageOptions() const = 0;`

fixed.

curvefs/src/metaserver/inode_storage.h
```diff
@@ -138,19 +141,45 @@ class InodeStorage {
...
+                              uint64_t delFirstChunkId,
+                              S3ChunkInfoList* list2add);
 
     std::string RealTablename(TABLE_TYPE type, std::string tablename) {
         std::ostringstream oss;
         oss << type << ":" << tablename;
         return oss.str();
     }
 
+    size_t InodeS3MetaSizeKey(uint32_t fsId, uint64_t inodeId) {
```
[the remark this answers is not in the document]
Wine93: > `static size_t InodeS3MetaSizeKey`
> 
> and why hashing the combined string, and what if two inodes have the same hashed key?

fixed~ 
yep, maybe different combined string has the same hashed key.

curvefs/src/metaserver/storage/rocksdb_storage.h
```diff
@@ -335,13 +339,21 @@ class RocksDBStorageIterator : public Iterator {
           ordered_(ordered) {
         if (status_ == 0) {
             readOptions_ = storage_->ReadOptions();
-            readOptions_.snapshot = storage_->db_->GetSnapshot();
+            if (storage_->InTransaction_) {
+                readOptions_.snapshot = storage_->txn_->GetSnapshot();
+            } else {
+                readOptions_.snapshot = storage_->db_->GetSnapshot();
+            }
```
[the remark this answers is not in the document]
Wine93: > what's the difference between in transaction and not in transaction?

if we use `db_->GetSnapshot()` to get snapshot , we will can't access the keys which set by transaction, consider the following situation:
```
txn = kvStorage->BeginTransaction();
txn->Set("key1", "value1");
Status s = txn->Get("key1");  // it will return Status:NotFound()
txn->Commit();
```

PR #1332, opened 2022-04-22
curvefs/client: edit log

wuhongsong:

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/s3/client_s3_cache_manager.cpp
+++ b/curvefs/src/client/s3/client_s3_cache_manager.cpp
@@ -155,7 +155,7 @@ CURVEFS_ERROR FsCacheManager::FsSync(bool force) {
     }
 
     auto iter = tmp.begin();
-    VLOG(3) << "FsSync force: " << force;
+    VLOG(9) << "FsSync force: " << force;
     for (; iter != tmp.end(); iter++) {
         ret = iter->second->Flush(force);
         if (ret == CURVEFS_ERROR::OK) {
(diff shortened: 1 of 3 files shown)


curvefs/src/client/s3/disk_cache_manager.cpp
```diff
@@ -271,7 +271,7 @@ int64_t DiskCacheManager::SetDiskFsUsedRatio() {
     int64_t usedBytes = totalBytes - freeBytes;
     int64_t usedPercent = 100 * usedBytes / (usedBytes + availableBytes) + 1;
 
-    VLOG(3) << "cache disk usage = " << usedPercent;
+    VLOG(9) << "cache disk usage = " << usedPercent;
```
ilixiaocui, line 274: LOG_ERVERY?
TODO: get from metric instead of log.

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -155,7 +155,7 @@ CURVEFS_ERROR FsCacheManager::FsSync(bool force) {
     }
 
     auto iter = tmp.begin();
-    VLOG(3) << "FsSync force: " << force;
+    VLOG(9) << "FsSync force: " << force;
```
ilixiaocui, line 158: delete

baijiaruo (approved):

PR #1365, opened 2022-04-27
curvefs/client: list dentry concurrently when get summary xattr

SeanHai:

Issue Number: #xxx 

Problem Summary:
1. list dentry concurrently when get summary xattr to improve performance.
2. rebuild xattr code in client and use XattrManager to deal with.
example: du dir1
```
dir1
├── dir2
│   └── file2
├── dir3
├── dir4
├── dir5
└── file1
```
before: need 5 times listDentry synchronously one by one.
now: use 10 threads default to listDentry concurrently.

The test result:
截屏2022-05-12 17 07 28

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/fuse_client.cpp
+++ b/curvefs/src/client/fuse_client.cpp
@@ -28,21 +28,19 @@
 #include <cstring>
 #include <string>
 #include <vector>
-#include <stack>
 #include <set>
 #include <unordered_map>
 
 #include "curvefs/proto/mds.pb.h"
 #include "curvefs/src/client/fuse_common.h"
 #include "curvefs/src/client/client_operator.h"
+#include "curvefs/src/client/xattr_manager.h"
 #include "src/common/net_common.h"
 #include "src/common/dummyserver.h"
 #include "src/client/client_common.h"
-#include "src/common/string_util.h"
 
 #define PORT_LIMIT 65535
 
-using ::curve::common::StringToUll;
 using ::curvefs::common::S3Info;
 using ::curvefs::common::Volume;
 using ::curvefs::mds::topology::PartitionTxId;
@@ -91,6 +89,9 @@ CURVEFS_ERROR FuseClient::Init(const FuseClientOption &option) {
 
     leaseExecutor_ =
         std::make_shared<LeaseExecutor>(option.leaseOpt, metaCache, mdsClient_);
+
+    xattrManager_ = std::make_shared<XattrManager>(inodeManager_,
+        dentryManager_, option_.listDentryLimit, option_.listDentryThreads);
 
     uint32_t listenPort = 0;
     if (!curve::common::StartBrpcDummyserver(option.dummyServerStartPort,
@@ -150,6 +151,7 @@ void FuseClient::Fini() {
         LOG(INFO) << "stop fuse client flush thread ...";
         sleeper_.interrupt();
         flushThread_.join();
+        xattrManager_->Stop();
     }
     LOG(INFO) << "stop fuse client flush thread ok.";
 }
@@ -326,7 +328,8 @@ CURVEFS_ERROR FuseClient::FuseOpOpen(fuse_req_t req, fuse_ino_t ino,
                 xattr.mutable_xattrinfos()->insert({XATTRFBYTES,
                     std::to_string(length)});
                 for (const auto &it : inode->parent()) {
-                    auto tret = UpdateParentInodeXattr(it, xattr, false);
+                    auto tret = xattrManager_->UpdateParentInodeXattr(
+                        it, xattr, false);
                     if (tret != CURVEFS_ERROR::OK) {
                         LOG(ERROR) << "UpdateParentInodeXattr failed,"
                                    << " inodeId = " << it
@@ -432,7 +435,7 @@ CURVEFS_ERROR FuseClient::MakeNode(fuse_req_t req, fuse_ino_t parent,
         }
         xattr.mutable_xattrinfos()->insert({XATTRFBYTES,
             std::to_string(inodeWrapper->GetLength())});
-        auto tret = UpdateParentInodeXattr(parent, xattr, true);
+        auto tret = xattrManager_->UpdateParentInodeXattr(parent, xattr, true);
         if (tret != CURVEFS_ERROR::OK) {
             LOG(ERROR) << "UpdateParentInodeXattr failed,"
                        << " inodeId = " << parent
@@ -544,7 +547,7 @@ CURVEFS_ERROR FuseClient::RemoveNode(fuse_req_t req, fuse_ino_t parent,
         }
         xattr.mutable_xattrinfos()->insert({XATTRFBYTES,
             std::to_string(inodeWrapper->GetLength())});
-        auto tret = UpdateParentInodeXattr(parent, xattr, false);
+        auto tret = xattrManager_->UpdateParentInodeXattr(parent, xattr, false);
         if (tret != CURVEFS_ERROR::OK) {
             LOG(ERROR) << "UpdateParentInodeXattr failed,"
                        << " inodeId = " << parent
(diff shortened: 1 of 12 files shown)


SeanHai:
recheck

SeanHai:
recheck

SeanHai:
recheck

SeanHai:
recheck

curvefs/src/client/fuse_client.cpp
```diff
@@ -946,68 +948,154 @@ CURVEFS_ERROR FuseClient::CalOneLayerSumInfo(Inode *inode) {
...
-            inodeIds.emplace(it.inodeid());
+
+        auto tret = dentryManager_->ListDentry(ino, dentrys,
+                                              listDentryLimit, dirOnly);
+        if (CURVEFS_ERROR::OK != tret) {
+            LOG(ERROR) << "ListDentry failed, inodeId = " << ino
+                       << ", limit = " << listDentryLimit << ", onlyDir = "
+                       << dirOnly << ", ret = " << tret;
+            ret->exchange(false);
+            inflightNum->fetch_sub(1, std::memory_order_seq_cst);
```
wu-hanqing, line 991: ditto

curvefs/src/client/fuse_client.cpp
```diff
@@ -946,68 +948,154 @@ CURVEFS_ERROR FuseClient::CalOneLayerSumInfo(Inode *inode) {
...
+                        (*rfbytes) += it.length();
+                        // record hardlink
+                        if (it.type() != FsFileType::TYPE_DIRECTORY &&
+                            it.nlink() > 1) {
+                            std::lock_guard<std::mutex> guard(*mapMutex);
+                            auto iter = hardLinkMap->find(it.inodeid());
+                            if (iter != hardLinkMap->end()) {
+                                iter->second += it.length();
+                            } else {
+                                hardLinkMap->emplace(it.inodeid(), 0);
```
wu-hanqing, line 1050: why insert a key-value pair and the value is 0 ?

should it be `it.length()` ?

if so, you can simplify these to `hardLinkMap[it.inodeid()] += it.length()`

curvefs/src/client/fuse_client.cpp
```diff
@@ -946,68 +948,154 @@ CURVEFS_ERROR FuseClient::CalOneLayerSumInfo(Inode *inode) {
...
+                            }
                         }
                     }
+                } else {
+                    ret->exchange(false);
```
wu-hanqing, line 1055: ditto, use `store`

curvefs/src/client/fuse_client.cpp
```diff
@@ -1025,16 +1113,88 @@ CURVEFS_ERROR FuseClient::CalAllLayerSumInfo(Inode *inode) {
...
+                        if (it.xattrinfos().count(XATTRFBYTES)) {
+                            if (!AddUllStringToFirst(
+                                &(inode->mutable_xattr()->find(XATTRRFBYTES)->second),    // NOLINT
+                                it.xattrinfos().find(XATTRFBYTES)->second,
+                                true)) {
+                                ret->exchange(false);
+                                break;
+                            }
+                        }
+                    }
```
wu-hanqing, line 1182: you add all subdirs/subfiles information into the target inode, for the calculations here, you use string addition and subtraction.

and I think you can just use integrals to sum the middle values, and finally, convert the value to string and assign the result to inode.

curvefs/src/client/fuse_client.cpp
```diff
@@ -946,68 +948,154 @@ CURVEFS_ERROR FuseClient::CalOneLayerSumInfo(Inode *inode) {
...
+    std::atomic<uint32_t> inflightNum(0);
+    std::atomic<bool> ret(true);
+
+    auto threadNum = option_.listDentryThreads;
+    for (auto i = threadNum; i > 0; i--) {
+        threadpool.emplace_back(Thread(
+                        &FuseClient::ConcurrentGetInodeAttr,
+                        this, &iStack, &stackMutex, &hardLinkMap, &mapMutex,
+                        false, &rfiles, &rsubdirs, &rentries, &rfbytes,
+                        &valueMutex, &inflightNum, &ret));
```
wu-hanqing, line 1089: it's better to pack all the parameters into a struct.

and another suggestion is, the logic here is relatively independent, you can extract the logic into a small class, it will be more readable and maintainable

curvefs/src/client/fuse_client.cpp
```diff
@@ -946,68 +948,154 @@ CURVEFS_ERROR FuseClient::CalOneLayerSumInfo(Inode *inode) {
...
+
+    uint64_t rfiles = 0;
+    uint64_t rsubdirs = 0;
+    uint64_t rentries = 0;
+    uint64_t rfbytes = 0;
+    std::mutex valueMutex;
+
+    auto ino = inode->inodeid();
+    iStack.emplace(ino);
+    std::vector<Thread> threadpool;
```
wu-hanqing, line 1079: you should take care of the dynamic threadpool, if there're many requests, it will start many many threads, but the total number of threads that a process can start is limited, and if the limit is reached, creating a new thread will throw an exception which will lead to abort the process.

curvefs/src/client/fuse_client.cpp
```diff
@@ -946,68 +948,154 @@ CURVEFS_ERROR FuseClient::CalOneLayerSumInfo(Inode *inode) {
...
+    std::stack<uint64_t> *iStack,
+    std::mutex *stackMutex,
+    bool dirOnly,
+    std::atomic<uint32_t> *inflightNum,
+    std::atomic<bool> *ret) {
+    auto listDentryLimit = option_.listDentryLimit;
+    InterruptibleSleeper sleeper;
+    uint64_t sleepInterval = 100;
+    while (1) {
+        auto ino = 0;
```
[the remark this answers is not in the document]
SeanHai: fixed

curvefs/src/client/fuse_client.h
```diff
@@ -241,6 +243,10 @@ class FuseClient {
         enableSumInDir_ = enable;
     }
 
+    void UpdateStop(bool isStop) {
+        isStop_.exchange(isStop);
```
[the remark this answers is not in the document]
SeanHai: fixed

curvefs/src/client/fuse_client.cpp
```diff
@@ -946,68 +948,154 @@ CURVEFS_ERROR FuseClient::CalOneLayerSumInfo(Inode *inode) {
...
-                    rfbytes += it.length();
-                    // record hardlink
-                    if (it.type() != FsFileType::TYPE_DIRECTORY &&
-                        it.nlink() > 1) {
-                        if (hardLinkMap.count(it.inodeid())) {
-                            hardLinkMap[it.inodeid()] += it.length();
+                    inodeIds.emplace(it.inodeid());
+                }
+            }
+            inflightNum->fetch_sub(1, std::memory_order_seq_cst);
```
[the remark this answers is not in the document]
SeanHai: fixed all

curvefs/src/client/fuse_client.cpp
```diff
@@ -946,68 +948,154 @@ CURVEFS_ERROR FuseClient::CalOneLayerSumInfo(Inode *inode) {
...
-        auto ret = dentryManager_->ListDentry(ino, &dentryList, limit, false);
-        if (CURVEFS_ERROR::OK != ret) {
-            LOG(ERROR) << "ListDentry failed, inodeId = " << ino
-                       << ", limit = " << limit << ", onlyDir = false";
-            return ret;
+        if (!iStack->empty()) {
+            ino = iStack->top();
+            iStack->pop();
+            stackMutex->unlock();
+            inflightNum->fetch_add(1, std::memory_order_seq_cst);
```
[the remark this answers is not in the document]
SeanHai: fixed

curvefs/src/client/fuse_client.cpp
SeanHai: fixed

curvefs/src/client/fuse_client.cpp
SeanHai: The logic is: if (it.type() != FsFileType::TYPE_DIRECTORY && it.nlink() > 1) means this file is a hardlink, but if others hardlinks of the file not exists in the dir which summary, the size is correct.
 The hardLinkMap <inodeId, the size need to minus>, so the first time is 0.

curvefs/src/client/fuse_client.cpp
SeanHai: fixed

curvefs/src/client/fuse_client.cpp
SeanHai: fixed

curvefs/src/client/fuse_client.cpp
SeanHai: fixed, use the class 'XattrManager' to deal with all the request about xattr.

curvefs/src/client/fuse_client.cpp
```diff
@@ -946,68 +948,154 @@ CURVEFS_ERROR FuseClient::CalOneLayerSumInfo(Inode *inode) {
...
+        } else {
+            stackMutex->unlock();
+            sleeper.wait_for(std::chrono::milliseconds(sleepInterval));
```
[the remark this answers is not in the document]
SeanHai: The conditions are dispersed and It is tricky to notify  when use conditional_variable.
This function is used seems tool, the sleep here is acceptable.

curvefs/src/client/fuse_client.cpp
SeanHai: yeap, using try and catch block to deal the failure when creat thread. It is doesn't matter if the thread num created less the wanted.

curvefs/src/client/xattr_manager.cpp
```diff
@@ -0,0 +1,543 @@
...
+    Atomic<bool> *ret) {
+    while (1) {
+        std::list<Dentry> dentryList;
+        std::set<uint64_t> inodeIds;
+        std::list<XAttr> xattrs;
+        auto tret = ConcurrentListDentry(&dentryList, iStack, stackMutex,
+                                         true, inflightNum, ret);
+        if (!tret) {
+            break;
+        } else {
```
Wine93, line 301: You can remove the `else` block, it can make logic more clearly:
```
if (!tret) {
    break;
}

// do something for else
```

SeanHai: fixed all

Wine93 (approved):
LGTM!

SeanHai:
recheck

curvefs/src/client/xattr_manager.cpp
```diff
@@ -0,0 +1,539 @@
...
+                this, &iStack, &stackMutex, &hardLinkMap, &mapMutex,
+                &summaryInfo, &valueMutex, &inflightNum, &ret));
+        } catch (const std::exception& e) {
+            LOG(WARNING) << "CalAllLayerSumInfo create thread failed,"
+                         << " err: " << e.what();
+        }
+    }
+
+    if (threadpool.empty()) {
+        return CURVEFS_ERROR::INTERNAL;
```
wu-hanqing, line 264: what happens when we return an error, does it will affect all io requests?

curvefs/src/client/xattr_manager.cpp
```diff
@@ -0,0 +1,539 @@
...
+
+    auto ino = inode->inodeid();
+    iStack.emplace(ino);
+    std::vector<Thread> threadpool;
+    Atomic<uint32_t> inflightNum(0);
+    Atomic<bool> ret(true);
+
+    for (auto i = listDentryThreads_; i > 0; i--) {
+        try {
+            threadpool.emplace_back(Thread(
```
wu-hanqing, line 253: can it works correctly when you only have one thread?

curvefs/src/client/xattr_manager.cpp
SeanHai: The error returned here is same as getXattrInfo failed, the error will return as EIO, and not affect other io requests.

curvefs/src/client/xattr_manager.cpp
SeanHai: Yes, I have test and the reselt is the picture top.

wu-hanqing (approved):

Issue #1278, opened 2022-04-14
Go segmentfault when compiling etcd in curvefs, the machine is stuck


YunhuiChen:
**Describe the bug (描述bug)**

1、mount curvefs
2、clone curve 
3、bash replace-curve-repo.sh

cd thirdparties/etcdclient && make clean && make all
4、
![image](https://user-images.githubusercontent.com/67582932/163372056-788797bd-cf25-41ef-ba52-098c98ec46d6.png)

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
curve-mds:
curve-chunkserver:
curve-snapshotcloneserver:
curve-sdk:
nebd:
curve-nbd:

**Additional context/screenshots (更多上下文/截图)**

ilixiaocui:
**go cmd md5sum check**

`after make install go
`curvefs：
/mnt/lxc/pubt1/curve/thirdparties/etcdclient$ md5sum tmp/go/bin/go
e54b6abda144554d352e6dce5f3deb15  tmp/go/bin/go

local：
curve/thirdparties/etcdclient/tmp$ md5sum go/bin/go
e54b6abda144554d352e6dce5f3deb15  go/bin/go

ilixiaocui:
https://elixir.bootlin.com/linux/v4.9.65/source/fs/fuse/file.c#L179：
<img width="572" alt="image" src="https://user-images.githubusercontent.com/15689619/163804132-2c305912-9c77-4454-bc6d-7d6ab4649f60.png">
<img width="1272" alt="image" src="https://user-images.githubusercontent.com/15689619/163804185-f4de9078-39d9-4feb-8155-e7fd8a1801cf.png">

ilixiaocui:
```
write(1, "/mnt/lxc/pubt1/curve/thirdpartie"..., 254/mnt/lxc/pubt1/curve/thirdparties/etcdclient/tmp/go/bin/go build -buildmode=c-shared -o libetcdclient.so /mnt/lxc/pubt1/curve/thirdparties/etcdclient/tmp/gosrc/src/etcdclient.go /mnt/lxc/pubt1/curve/thirdparties/etcdclient/tmp/gosrc/src/objectManager.go
) = 254
rt_sigprocmask(SIG_BLOCK, [HUP INT QUIT TERM XCPU XFSZ], NULL, 8) = 0
[...]

PR #1366, opened 2022-04-27
curvefs: add switch for fuse splice

wu-hanqing:

Issue Number: #1278

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/conf/client.conf
+++ b/curvefs/conf/client.conf
@@ -88,6 +88,11 @@ fuseClient.dCacheLruSize=65536
 fuseClient.enableICacheMetrics=true
 fuseClient.enableDCacheMetrics=true
 fuseClient.cto=true
+# splice will bring higher performance in some cases
+# but there're might be a kernel issue that will cause kernel panic when enabling it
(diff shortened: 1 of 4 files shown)


curvefs/conf/client.conf
```diff
@@ -88,6 +88,11 @@ fuseClient.dCacheLruSize=65536
 fuseClient.enableICacheMetrics=true
 fuseClient.enableDCacheMetrics=true
 fuseClient.cto=true
+# splice will bring higher performance in some cases
+# but there're might be a kernel issue that will cause kernel panic when enabling it
+# see https://lore.kernel.org/all/CAAmZXrsGg2xsP1CK+cbuEMumtrqdvD-NKnWzhNcvn71RV3c1yw@mail.gmail.com/
+# until this issue has been fixed, splice should be disabled
+fuseClient.enableSplice=false
```
ilixiaocui, line 95: Note that there is no problem starting from 5.10, there is a problem with 5.4 and before, and the version between 5.4 and 5.10 is uncertain

wu-hanqing: yep, but hard code a kernel version is not flexible, because someone may patch fixes and then splice can be enabled.

ilixiaocui: OK

ilixiaocui (approved):

Issue #1370, opened 2022-04-28
brpc can not be compiled on high version glibc


skypexu:
**Describe the bug (描述bug)**
brpc redefined gcc atttribute `__const__ `to empty string, this does not work on high version glibc, for example on version 2.35-0ubuntu3. the problem is glibc defined a macro glibc_has_attribute which is a redefinition of `__has_attribute`, `__has_attribute` needs a non-empty argument, for example `__has_attribute(__const__)`, if `__const__` is empty string, compiler will fail.

**To Reproduce (复现方法)**
compile curve on ubuntu jammy 

**Expected behavior (期望行为)**
can be compiled

**Versions (各种版本)**
OS: ubuntu jammy 
Compiler: gcc (Ubuntu 11.2.0-19ubuntu1) 11.2.0
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

fansehep:
The problem has been solved. https://github.com/apache/incubator-brpc/pull/1783

PR #1371, opened 2022-04-28
brpc: Fix brpc __const__ attribute problem.

skypexu:
replace __const__ with __used__, so brpc can be compiled on high version glibc.

Issue Number: #1370

Problem Summary:
The brpc can not be compiled on ubuntu jammy

What's Changed:
add a patch brpc_2.patch to replace __const__ with __used__
How it Works:

no side effects

### Check List


thirdparties/brpc/brpc_2.patch
```diff
@@ -0,0 +1,325 @@
...
+index 0de00b2..159a273 100644
+--- a/src/bthread/mutex.cpp
++++ b/src/bthread/mutex.cpp
+@@ -405,8 +405,8 @@ static void init_sys_mutex_lock() {
+ #if defined(OS_LINUX)
+     // TODO: may need dlvsym when GLIBC has multiple versions of a same symbol.
+     // http://blog.fesnel.com/blog/2009/08/25/preloading-with-multiple-symbol-versions
+-    sys_pthread_mutex_lock = (MutexOp)_dl_sym(RTLD_NEXT, "pthread_mutex_lock", (void*)init_sys_mutex_lock);
+-    sys_pthread_mutex_unlock = (MutexOp)_dl_sym(RTLD_NEXT, "pthread_mutex_unlock", (void*)init_sys_mutex_lock);
++    sys_pthread_mutex_lock = (MutexOp)dlvsym(RTLD_NEXT, "pthread_mutex_lock", "GLIBC_2.2.5");
```
xu-chaojie, line 284: why change here?

thirdparties/brpc/brpc_2.patch
```diff
@@ -0,0 +1,325 @@
...
++    sys_pthread_mutex_unlock = (MutexOp)dlvsym(RTLD_NEXT, "pthread_mutex_unlock", "GLIBC_2.2.5");
```
cw123, line 285: why specify "GLIBC_2.2.5"

skypexu: I have removed patch for mutex.
bthread mutex can not be compiled on high version glibc too, it should be fixed in another patch.

cw123 (approved):

PR #1372, opened 2022-04-28
curvefs/client: add -o mdsaddr=xxxx support

h0hmj:
### What's Changed

1. remove option: user, volume
2. add option: mdsaddr
3. format help message
4. fix coredump issue

``` bash 
$ ../bazel-bin/curvefs/src/client/curve-fuse --help
usage: ../bazel-bin/curvefs/src/client/curve-fuse -o conf=/etc/curvefs/client.conf -o fsname=testfs \
       -o fstype=s3 [-o mdsAddr=1.1.1.1] [OPTIONS] <mountpoint>
Fuse Options:
    -h   --help            print help
    -V   --version         print version
    -d   -o debug          enable debug output (implies -f)
    -f                     foreground operation
    -s                     disable multi-threaded operation
    -o clone_fd            use separate fuse device fd for each thread
                           (may improve performance)
    -o max_idle_threads    the maximum number of idle worker threads
                           allowed (default: 10)
    -o allow_other         allow access by all users
    -o allow_root          allow access by root
    -o auto_unmount        auto unmount on process termination

Extra options:
    -o fsname              [required] name of filesystem to be mounted
    -o fstype              [required] type of filesystem to be mounted (s3/volume)
    -o conf                [required] path of config file
    -o mdsaddr             mdsAddr of curvefs cluster
```

### DIFF
--- a/curvefs/src/client/main.c
+++ b/curvefs/src/client/main.c
@@ -54,21 +54,38 @@ static const struct fuse_lowlevel_ops curve_ll_oper = {
     .statfs     = FuseOpStatFs,
 };
 
+void print_option_help(const char* o, const char* msg) {
+    printf("    -o %-20s%s\n", o, msg);
+}
+
+void extra_options_help() {
+    printf("\nExtra options:\n");
+    print_option_help("fsname", "[required] name of filesystem to be mounted");
+    print_option_help(
+        "fstype", "[required] type of filesystem to be mounted (s3/volume)");
+    print_option_help("conf", "[required] path of config file");
+    print_option_help("mdsAddr", "mdsAddr of curvefs cluster");
+}
+
 int main(int argc, char *argv[]) {
     struct fuse_args args = FUSE_ARGS_INIT(argc, argv);
     struct fuse_session *se;
     struct fuse_cmdline_opts opts;
     struct fuse_loop_config config;
-    struct MountOption mOpts = { .mountPoint = 0,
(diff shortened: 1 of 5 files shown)


curvefs/src/client/main.c
```diff
@@ -54,21 +54,38 @@ static const struct fuse_lowlevel_ops curve_ll_oper = {
...
+    printf("    -o %-20s%s\n", o, msg);
+}
+
+void extra_options_help() {
+    printf("\nExtra options:\n");
+    print_option_help("fsname", "[required] name of filesystem to be mounted");
+    print_option_help(
+        "fstype", "[required] type of filesystem to be mounted (s3/volume)");
+    print_option_help("conf", "[required] path of config file");
+    print_option_help("mdsAddr", "mdsAddr of curvefs cluster");
```
SeanHai, line 67: Why pick mdsAddr out? If so, the three mds address are all needed.

h0hmj: this is for multi clients situation multi backend clusters
client A -> backend cluster A
client B -> backend cluster B

the only difference between multi clients' conf is the mdsAddr parameter

so we need add the ability to set mdsAddr using cmd to avoid too many conf files, it's messy

old: client number * 1 conf file per client
new: 1 conf file shared and different mdsAddr set by cmd

wu-hanqing (approved):

SeanHai (approved):

PR #1384, opened 2022-04-29
curvefs/client: fix s3 takes up more space than  actually uses

baijiaruo:
Although there is no error in repeatedly flushing datacache, the total amount of data in s3 is much larger than the amount of used files in the file system.
Issue Number: #1348  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/s3/client_s3_cache_manager.cpp
+++ b/curvefs/src/client/s3/client_s3_cache_manager.cpp
@@ -281,7 +281,10 @@ int FileCacheManager::Read(uint64_t inodeId, uint64_t offset, uint64_t length,
         } else {
             readLen = length;
         }
-        ReadChunk(index, chunkPos, readLen, dataBuf, readOffset, &requests);
+        ChunkCacheManagerPtr chunkCacheManager =
+        FindOrCreateChunkCacheManager(index);
+        chunkCacheManager->ReadChunk(index, chunkPos, readLen, dataBuf,
+                                     readOffset, &requests);
         totalRequests.insert(totalRequests.end(), requests.begin(),
                              requests.end());
         length -= readLen;
@@ -874,26 +877,6 @@ void FileCacheManager::GenerateS3Request(ReadRequest request,
     return;
 }
 
-void FileCacheManager::ReadChunk(uint64_t index, uint64_t chunkPos,
-                                 uint64_t readLen, char *dataBuf,
-                                 uint64_t dataBufOffset,
-                                 std::vector<ReadRequest> *requests) {
-    ChunkCacheManagerPtr chunkCacheManager =
-        FindOrCreateChunkCacheManager(index);
-    ReadLockGuard readLockGuard(chunkCacheManager->rwLockChunk_);
-    std::vector<ReadRequest> cacheMissRequests;
-    chunkCacheManager->ReadByWriteCache(chunkPos, readLen, dataBuf,
-                                        dataBufOffset, &cacheMissRequests);
-    for (auto request : cacheMissRequests) {
-        std::vector<ReadRequest> tmpRequests;
-        chunkCacheManager->ReadByReadCache(request.chunkPos, request.len,
-                                           dataBuf, request.bufOffset,
-                                           &tmpRequests);
-        requests->insert(requests->end(), tmpRequests.begin(),
-                         tmpRequests.end());
-    }
-    return;
-}
 
 void FileCacheManager::ReleaseCache() {
     WriteLockGuard writeLockGuard(rwLock_);
@@ -1009,6 +992,48 @@ CURVEFS_ERROR FileCacheManager::Flush(bool force, bool toS3) {
         }
         VLOG(6) << "file cache flush over";
         return ret;
+}
+
+void ChunkCacheManager::ReadChunk(uint64_t index, uint64_t chunkPos,
+                                 uint64_t readLen, char *dataBuf,
+                                 uint64_t dataBufOffset,
+                                 std::vector<ReadRequest> *requests) {
+    ReadLockGuard readLockGuard(rwLockChunk_);
+    std::vector<ReadRequest> cacheMissWriteRequests, cacheMissFlushDataRequest;
+    ReadByWriteCache(chunkPos, readLen, dataBuf,
+                                        dataBufOffset, &cacheMissWriteRequests);
+    flushingDataCacheMtx_.lock();
+    if (!IsFlushDataEmpty()) {
+        for (auto request : cacheMissWriteRequests) {
+            std::vector<ReadRequest> tmpRequests;
+            ReadByFlushData(request.chunkPos, request.len,
+                                               dataBuf, request.bufOffset,
+                                               &tmpRequests);
+            cacheMissFlushDataRequest.insert(cacheMissFlushDataRequest.end(),
+                                             tmpRequests.begin(),
+                                             tmpRequests.end());
+        }
+        flushingDataCacheMtx_.unlock();
+        for (auto request : cacheMissFlushDataRequest) {
+            std::vector<ReadRequest> tmpRequests;
+            ReadByReadCache(request.chunkPos, request.len,
+                                               dataBuf, request.bufOffset,
+                                               &tmpRequests);
+            requests->insert(requests->end(), tmpRequests.begin(),
+                             tmpRequests.end());
+        }
+        return;
+    }
+    flushingDataCacheMtx_.unlock();
+    for (auto request : cacheMissWriteRequests) {
+        std::vector<ReadRequest> tmpRequests;
+        ReadByReadCache(request.chunkPos, request.len,
+                                               dataBuf, request.bufOffset,
+                                               &tmpRequests);
+        requests->insert(requests->end(), tmpRequests.begin(),
+                             tmpRequests.end());
+    }
+    return;
 }
 
 void ChunkCacheManager::ReadByWriteCache(uint64_t chunkPos, uint64_t readLen,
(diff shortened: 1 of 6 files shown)


aspirer (changes requested):
need more comments.

baijiaruo:
recheck

baijiaruo:
recheck

baijiaruo:
recheck

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -2017,115 +2023,98 @@ CURVEFS_ERROR DataCache::Flush(uint64_t inodeId, bool force, bool toS3) {
...
+    }
+    data = new (std::nothrow) char[len_];
+    if (!data) {
+        LOG(ERROR) << "new data failed.";
         mtx_.unlock();
+        return CURVEFS_ERROR::INTERNAL;
+    }
+    CopyDataCacheToBuf(0, len_, data);
+    status_.store(DataCacheStatus::Flush, std::memory_order_release);
+    mtx_.unlock();
```
wu-hanqing, line 2051: The critical section here is unreasonable.

I think only a few things should be protected by the lock, like line 2032-2035 and line 2049, and others should not.

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -2180,6 +2162,24 @@ void DataCache::PrepareS3ChunkInfo(uint64_t chunkId, uint64_t offset,
...
 }
 
+bool DataCache::CanFlush(bool force) {
+    uint64_t chunkSize = s3ClientAdaptor_->GetChunkSize();
+    uint64_t now = ::curve::common::TimeUtility::GetTimeofDaySec();
+    uint32_t flushIntervalSec = s3ClientAdaptor_->GetFlushInterval();
+
+    if (force) {
+        return true;
+    }
```
wu-hanqing, line 2172: it's better to put this condition to the begining.

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -2180,6 +2162,24 @@ void DataCache::PrepareS3ChunkInfo(uint64_t chunkId, uint64_t offset,
...
+
+    if (len_ == chunkSize) {
```
wu-hanqing, line 2174: is there any chance that will lead to len_ > chunkSize ?

curvefs/src/client/s3/client_s3_cache_manager.h
```diff
@@ -95,6 +95,11 @@ struct PageData {
 };
 using PageDataMap = std::map<uint64_t, PageData *>;
 
+enum DataCacheStatus {
```
wu-hanqing, line 98: use enum class is better

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1436,38 +1426,55 @@ CURVEFS_ERROR ChunkCacheManager::Flush(uint64_t inodeId, bool force,
...
+                ret = dataCache->Flush(inodeId, toS3);
+                if (ret == CURVEFS_ERROR::NOTEXIST) {
+                    LOG(WARNING)
+                        << "dataCache flush failed. ret:" << ret
+                        << ",index:" << index_
+                        << ",data chunkpos:" << dataCache->GetChunkPos();
+                    ReleaseWriteDataCache(dataCache);
+                    break;
+                } else if (ret == CURVEFS_ERROR::INTERNAL) {
+                    ::sleep(3);
```
wu-hanqing, line 1462: why hard code 3 here? should it be configurable?

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1436,38 +1426,55 @@ CURVEFS_ERROR ChunkCacheManager::Flush(uint64_t inodeId, bool force,
...
-                    AddReadDataCache(iter->second);
+                    WriteLockGuard lockGuard(rwLockChunk_);
+                    AddReadDataCache(dataCache);
                 }
-                ReleaseWriteDataCache(iter->second);
-            } else {
-                VLOG(6) << "data cache is dirty.";
-            }
+                ReleaseWriteDataCache(dataCache);
+            } while (ret != CURVEFS_ERROR::OK);
```
wu-hanqing, line 1474: you should reset `dataCache` to `nullptr` after `while`, because if dataCache is not null, and if you find another datacache that can be flushed in line 1438, the assignment operator will call previous datacache's destructor, and this happens when holding `rwLockChunk_`

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1436,38 +1426,55 @@ CURVEFS_ERROR ChunkCacheManager::Flush(uint64_t inodeId, bool force,
...
+    DataCachePtr dataCache;
+    while (1) {
+        bool isFlush = false;
+        {
+            WriteLockGuard writeLockGuard(rwLockChunk_);
+
+            auto iter = dataWCacheMap_.begin();
+            while (iter != dataWCacheMap_.end()) {
+                if (iter->second->CanFlush(force)) {
+                    dataCache = iter->second;
```
wu-hanqing, line 1438: `dataCache = std::move(iter->second)`

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1501,7 +1508,7 @@ DataCache::DataCache(S3ClientAdaptorImpl *s3ClientAdaptor,
                      ChunkCacheManagerPtr chunkCacheManager, uint64_t chunkPos,
                      uint64_t len, const char *data)
     : s3ClientAdaptor_(s3ClientAdaptor), chunkCacheManager_(chunkCacheManager),
-      dirty_(true), delete_(false), inReadCache_(false) {
+      status_(DataCacheStatus::Dirty), inReadCache_(false) {
```
wu-hanqing, line 1511: why the initial status is Dirty?

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1767,7 +1774,7 @@ void DataCache::Write(uint64_t chunkPos, uint64_t len, const char *data,
                 << ", len:" << (*iter)->GetLen();
     }
     curve::common::LockGuard lg(mtx_);
-    dirty_.exchange(true, std::memory_order_acq_rel);
+    status_.exchange(DataCacheStatus::Dirty, std::memory_order_acq_rel);
```
wu-hanqing, line 1777: `status_.store(` is enough

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1411,23 +1410,14 @@ void ChunkCacheManager::TruncateReadCache(uint64_t chunkPos) {
...
-        }
-    } else {
-        VLOG(9) << "ReleaseWriteDataCache is deleted key:" << key;
-        rwLockWrite_.Unlock();
+    s3ClientAdaptor_->GetFsCacheManager()->DataCacheNumFetchSub(1);
+    VLOG(9) << "chunk flush DataCacheByteDec len:"
+            << dataCache->GetActualLen();
+    s3ClientAdaptor_->GetFsCacheManager()->DataCacheByteDec(
+        dataCache->GetActualLen());
+    if (!s3ClientAdaptor_->GetFsCacheManager()->WriteCacheIsFull()) {
```
xu-chaojie, line 1418: why WriteCacheIsFull no need  FlushSignal?

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1411,23 +1410,14 @@ void ChunkCacheManager::TruncateReadCache(uint64_t chunkPos) {
 }
 
 void ChunkCacheManager::ReleaseWriteDataCache(const DataCachePtr &dataCache) {
-    rwLockWrite_.WRLock();
-    uint64_t key = dataCache->GetChunkPos();
-    if (dataWCacheMap_.count(key)) {
```
xu-chaojie, line 1416: why not need judge key here?

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1436,38 +1426,55 @@ CURVEFS_ERROR ChunkCacheManager::Flush(uint64_t inodeId, bool force,
...
-                         << ",data chunkpos:" << iter->second->GetChunkPos();
-            return ret;
+    DataCachePtr dataCache;
+    while (1) {
+        bool isFlush = false;
+        {
+            WriteLockGuard writeLockGuard(rwLockChunk_);
+
+            auto iter = dataWCacheMap_.begin();
+            while (iter != dataWCacheMap_.end()) {
```
xu-chaojie, line 1436: Possibly dataWCacheMap_ will be empty? The while (1) loop never stops.

curvefs/src/client/s3/client_s3_cache_manager.cpp
xu-chaojie: add some log here ?

curvefs/src/client/s3/client_s3_cache_manager.cpp
xu-chaojie: why len_ == chunkSize  must to flush? maybe we can wait more time to make more io merge?

curvefs/src/client/s3/client_s3_cache_manager.cpp
baijiaruo: > why hard code 3 here? should it be configurable?
CURVEFS_ERROR::INTERNAL  this scenario will only occur in a few cases, there is no need to retry a configuration separately

baijiaruo: > add some log here ?

fixed

curvefs/src/client/s3/client_s3_cache_manager.cpp
baijiaruo: > you should reset `dataCache` to `nullptr` after `while`, because if dataCache is not null, and if you find another datacache that can be flushed in line 1438, the assignment operator will call previous datacache's destructor, and this happens when holding `rwLockChunk_`

fixed

curvefs/src/client/s3/client_s3_cache_manager.cpp
baijiaruo: Because the data has been copied to the datacache during initialization

curvefs/src/client/s3/client_s3_cache_manager.cpp
baijiaruo: fixed

curvefs/src/client/s3/client_s3_cache_manager.cpp
baijiaruo: > The critical section here is unreasonable.
> 
> I think only a few things should be protected by the lock, like line 2032-2035 and line 2049, and others should not.
I plan to remove this lock because datacache has been removed from dataWCacheMap_, there is no concurrency

curvefs/src/client/s3/client_s3_cache_manager.cpp
baijiaruo: fixed

curvefs/src/client/s3/client_s3_cache_manager.h
baijiaruo: fixed

curvefs/src/client/s3/client_s3_cache_manager.cpp
baijiaruo: > why len_ == chunkSize must to flush? maybe we can wait more time to make more io merge?

chunkSize is already the largest and will not be merged unless it is overwrite

curvefs/src/client/s3/client_s3_cache_manager.cpp
baijiaruo: fixed

xu-chaojie (approved):

wu-hanqing (approved):

baijiaruo:
recheck

baijiaruo:
recheck

baijiaruo:
> need more comments.

fixed

aspirer (approved):

xu-chaojie (approved):

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -1218,6 +1243,85 @@ void ChunkCacheManager::ReadByReadCache(uint64_t chunkPos, uint64_t readLen,
     return;
 }
 
+void ChunkCacheManager::ReadByFlushData(uint64_t chunkPos, uint64_t readLen,
```
wu-hanqing, line 1246: the main logic of this function is very similar to the loop in `ReadByReadCache`, so can you union the logic of these two functions?

baijiaruo: yes，todo later

Issue #1398, opened 2022-05-06
different filesystems generate identical inodeid sequences


wu-hanqing:
**Describe the bug (描述bug)**

According to our design, create an inode will randomly choices one partition of current filesystem.

But I saw determined results of create inodes.
 
**To Reproduce (复现方法)**

I create and mount 10 different filesystems and within each filesystem, create 10 files using the command 

`for i in {1..10}; do touch hello_${i}; done`

and show the inode id of created files of different filesystems, and the result is interesting

<img width="1445" alt="image" src="https://user-images.githubusercontent.com/10327590/167079814-c75cba2e-9fd9-446f-87b2-20000d0a1bc8.png">

**Versions (各种版本)**
commit id: f478e2ed7a162c85f5cc82232363e652a13ea48d

**Additional context/screenshots (更多上下文/截图)**

PR #1400, opened 2022-05-06
curvefs/fuse: fix generate identical inodeid sequences

wu-hanqing:

Issue Number: #1398 

Problem Summary: previous policy use `rand` without `srand()`, so each filesystem will generate same random number sequences

What's Changed: replace `rand` with `butil::fast_rand()`

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/rpcclient/metacache.cpp
+++ b/curvefs/src/client/rpcclient/metacache.cpp
@@ -20,6 +20,7 @@
  * Author: lixiaocui
  */
 
+#include <butil/fast_rand.h>
 #include <algorithm>
 #include <iterator>
 #include <vector>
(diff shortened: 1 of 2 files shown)


Cyber-SiKu (approved):

curvefs/src/client/rpcclient/metacache.cpp
```diff
@@ -492,7 +493,7 @@ bool MetaCache::SelectPartition(CopysetTarget *target) {
         target->txId = newPartitions[0].txid();
     } else {
         // random select a partition
-        int index = rand() % candidate.size();  // NOLINT
+        const auto index = butil::fast_rand() % candidate.size();
```
SeanHai, line 496: butil::fast_rand() doesn't init with different seed, will generate same sequence？

wu-hanqing: `butil::fast_rand` keeps a hidden state for each thread and init state the first time when calling the function.

you can see this from the source code

https://github.com/apache/incubator-brpc/blob/c75d53a62c181db0aec300f41787c0c0786c69d0/src/butil/fast_rand.cpp#L88

SeanHai (approved):

Issue #1382, opened 2022-04-29
Copyset is not automatically migrated after stopping metaserver


YunhuiChen:
**Describe the bug (描述bug)**

Copyset is not automatically migrated after stopping metaserver
**To Reproduce (复现方法)**
1、stop metaserver
2、wait 4 hours
3、
![image](https://user-images.githubusercontent.com/67582932/165904329-707acd96-3d97-42c5-81c0-894178ab8ef5.png)

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

cw123:
I 2022-04-29T12:02:54.543547+0800    80 recoverScheduler.cpp:34] recoverScheduler begin.
I 2022-04-29T12:02:54.543730+0800    80 operatorControllerTemplate.h:141] add operator [startEpoch: 3, copysetID: (1,5), priority: 2, step: change peer from 5 to 6] fail because of oncurrency exceed
W 2022-04-29T12:02:54.543761+0800    80 recoverScheduler.cpp:95] recover scheduler add operator [startEpoch: 3, copysetID: (1,5), priority: 2, step: change peer from 5 to 6] on [copysetId:(1,5), epoch:3, leader:2, peers:(1,2,5,), canidate:0, has configChangeInfo:0] fail
I 2022-04-29T12:02:54.543777+0800    80 operatorControllerTemplate.h:141] add operator [startEpoch: 3, copysetID: (1,6), priority: 2, step: change peer from 5 to 6] fail because of oncurrency exceed
W 2022-04-29T12:02:54.543783+0800    80 recoverScheduler.cpp:95] recover scheduler add operator [startEpoch: 3, copysetID: (1,6), priority: 2, step: change peer from 5 to 6] on [copysetId:(1,6), epoch:3, leader:4, peers:(2,4,5,), canidate:0, has configChangeInfo:0] fail

cw123:
the recover generate operator, but the operator exceed the  concurrency
the concurrent is set 1 at mds.conf.  
mds.schduler.operator.concurrent=1

cw123:
过了6天，copyset配置变更还是卡在这里
![image](https://user-images.githubusercontent.com/4277834/166895753-35cf3739-0486-4676-901d-891117c9c02e.png)

cw123:
mds日志里面，心跳有报这个错误
[...]

PR #1417, opened 2022-05-10
fix clear copyset creating flag when create copyset success

cw123:

Issue Number: #1382

Problem Summary: Copyset is not automatically migrated after stopping metaserver

What's Changed: fix clear copyset creating flag when create copyset success

When copyset created success, it should clear the copyset creating flag. But the code miss this in code.

### DIFF
--- a/curvefs/src/mds/topology/topology_manager.cpp
+++ b/curvefs/src/mds/topology/topology_manager.cpp
@@ -853,7 +853,7 @@ TopoStatusCode TopologyManager::CreateCopyset(
         return TopoStatusCode::TOPO_CREATE_COPYSET_ON_METASERVER_FAIL;
     }
 
-    // add copyset record to topogy
+    // add copyset record to topology
     CopySetInfo copysetInfo(copyset.poolId, copyset.copysetId);
     copysetInfo.SetCopySetMembers(copyset.metaServerIds);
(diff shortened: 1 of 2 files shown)


cw123:
recheck

cw123:
recheck

YunhuiChen:
recheck

Wine93 (changes requested):
LGTM!

Wine93 (approved):
LGTM!

curvefs/src/mds/topology/topology_manager.cpp
```diff
@@ -866,6 +866,7 @@ TopoStatusCode TopologyManager::CreateCopyset(
         return ret;
     }
 
+    ClearCopysetCreating(copyset.poolId, copyset.copysetId);
```
wu-hanqing, line 869: it's better to move this line of code before 856 and remove line 865, so you don't have to repeat it.

and should you add some unit test?

cw123: 虽然都是去掉creating的标记，但是清理的时机是不一样的。如果都挪到856行，和原意还是有区别的。

Although the marking of creating is removed, the timing of cleaning is different. If it is moved to line 856, it is still different from the original intention.

wu-hanqing (approved):

PR #1432, opened 2022-05-12
Fix clear copyset creating flag to v2.2

cw123:

Issue Number: #1382

Problem Summary: Copyset is not automatically migrated after stopping metaserver

What's Changed: fix clear copyset creating flag when create copyset success

When copyset created success, it should clear the copyset creating flag. But the code miss this in code.


cw123:
It should merge to branch v2.2.

Issue #1369, opened 2022-04-28
report bug: dentry not exist


Wine93:
Describe
---

We found that some files were lost which means we can't get its dentry. In the next, I will show 2 test cases to explain in which situation using `mv` operation will cause file exception.

* CurveFS version: `v2.1.0`

Test Cases:
---

### CASE 1: three different directories under there different mountpoints of one filesystem

* step1: mount filesystem
```
$ sudo curveadm mount /test01 /path/to/mountpoint1
$ sudo curveadm mount /test01 /path/to/mountpoint2
$ sudo curveadm mount /test01 /path/to/mountpoint3
```

* step2: generate 1000 files in three directories

```
generate_file.sh
for ((i=1;i<=3;i++)); do
    cd /path/to/mountpoint$i
    mkdir dir$i
    cd dir$i
    for ((j=1;j<=1000;j++)); do
       echo $i > $j
    done    
done
```

* step3: run `mv` operator in three directories with parallel

```
# run_single.sh
cd /path/to/mountpoint$1/dir$1
for ((i=1;i<=1000;i++)); do
   mv $i $i.back
done
```

```
# run_multi.sh
./run_signle.sh 1 &
./run_signle.sh 2 &
./run_signle.sh 3 &
```

#### output summary: 
* except: all files in three different directories will be renamed to `xxx.back`
* actuall: many files are not renamed

### CASE 2: three different directories under one mountpoint of one filesyetem

* step1: mount filesystem
```
$ sudo curveadm mount /test01 /path/to/mountpoint1
```

* step2: generate 1000 files in three directories

```
generate_file.sh
for ((i=1;i<=3;i++)); do
    cd /path/to/mountpoint1
    mkdir dir$i
    cd dir$i
    for ((j=1;j<=1000;j++)); do
       echo $i > $j
    done    
done
```

* step3: run `mv` operator in three directories with parallel

```
# run_single.sh
cd /path/to/mountpoint1/dir$1
for ((i=1;i<=1000;i++)); do
   mv $i $i.back
done
```

```
# run_multi.sh
./run_single.sh 1 &
./run_single.sh 2 &
./run_single.sh 3 &
```

#### output summary: 
* except: all files in three different directories will be renamed to `xxx.back`
* actuall: all files are renamed to `xxx.back`

Hint
---

for case 1, we found that the `txid`in client's RPC request is not increasing strictly, it maybe lead to the `rename` operation not as expected.

PR #1437, opened 2022-05-12
curvefs: support rename by parallel for multiple mountpoints, you MUST

Wine93:
set client configure item fuseClient.enableMultiMountPointRename to true
for gurantee the consistent of file after rename in multiple mountpoints. (#1369)

Issue Number: #1369

Problem Summary: dentry losed after rename in multiple mountpoints.

### How it Works:
* for multiple mointpoints, each rename operation needs to acquire a distributed mutex lock from MDS for gurantee the `txId` incremented.
* each lock bind a `uuid` for gurantee the lock can reentrant.
* each rename operation will bind a transaction sequence for discard the stale transaction, which means if the `sequence` of current transaction not greater than the `sequence` of pending transaction, the current transaction should be discarded.

### DIFF
--- a/curvefs/conf/client.conf
+++ b/curvefs/conf/client.conf
@@ -88,6 +88,10 @@ fuseClient.dCacheLruSize=65536
 fuseClient.enableICacheMetrics=true
 fuseClient.enableDCacheMetrics=true
 fuseClient.cto=true
+# you shoudle enable it when mount one filesystem to multi mountpoints,
+# it gurantee the consistent of file after rename, otherwise you should
+# disable it for performance.
+fuseClient.enableMultiMountPointRename=true
 # splice will bring higher performance in some cases
 # but there're might be a kernel issue that will cause kernel panic when enabling it
 # see https://lore.kernel.org/all/CAAmZXrsGg2xsP1CK+cbuEMumtrqdvD-NKnWzhNcvn71RV3c1yw@mail.gmail.com/
--- a/curvefs/proto/mds.proto
+++ b/curvefs/proto/mds.proto
@@ -49,6 +49,9 @@ enum FSStatusCode {
     PARTITION_EXIST = 25;
     DELETE_PARTITION_ERROR = 26;
     S3_INFO_ERROR = 27;
+    LOCK_FAILED = 28;
+    LOCK_TIMEOUT = 29;
+    COMMIT_TX_SEQUENCE_MISMATCH = 30;
 }
 
 // fs interface
@@ -83,6 +86,8 @@ message FsInfo {
     required FsDetail detail = 10;
     required bool enableSumInDir = 11;
     required string owner = 12;
+    optional uint64 txSequence = 13;
+    optional string txOwner = 14;
 }
 
 message GetFsInfoResponse {
@@ -169,6 +174,36 @@ message RefreshSessionResponse {
     repeated topology.PartitionTxId latestTxIdList = 2;
 }
 
+message DLockValue {
+    required string owner = 1;
+    required uint64 expireTime = 2;
+}
+
+message GetLatestTxIdRequest {
+    optional bool lock = 1;
+    optional uint32 fsId = 2;
+    optional string fsName = 3;
+    optional string uuid = 4;
+}
+
+message GetLatestTxIdResponse {
+    required FSStatusCode statusCode = 1;
+    repeated topology.PartitionTxId txIds = 2;
+    optional uint64 txSequence = 3;
+}
+
+message CommitTxRequest {
+    repeated topology.PartitionTxId partitionTxIds = 1;
+    optional bool lock = 2;
+    optional string fsName = 3;
+    optional string uuid = 4;
+    optional uint64 txSequence = 5;
+}
+
+message CommitTxResponse {
+    required FSStatusCode statusCode = 1;
+}
+
 service MdsService {
     // fs interface
     rpc CreateFs(CreateFsRequest) returns (CreateFsResponse);
@@ -180,6 +215,9 @@ service MdsService {
     rpc DeleteFs(DeleteFsRequest) returns (DeleteFsResponse);
     rpc AllocateS3Chunk(AllocateS3ChunkRequest) returns (AllocateS3ChunkResponse);
     rpc ListClusterFsInfo (ListClusterFsInfoRequest) returns (ListClusterFsInfoResponse);
+
+    rpc GetLatestTxId(GetLatestTxIdRequest) returns (GetLatestTxIdResponse);
+    rpc CommitTx(CommitTxRequest) returns (CommitTxResponse);
 
     // client lease
     rpc RefreshSession(RefreshSessionRequest) returns (RefreshSessionResponse);
--- a/curvefs/src/client/client_operator.cpp
+++ b/curvefs/src/client/client_operator.cpp
@@ -22,11 +22,13 @@
 
 #include <list>
 
+#include "src/common/uuid.h"
 #include "curvefs/src/client/client_operator.h"
 
 namespace curvefs {
 namespace client {
 
+using ::curve::common::UUIDGenerator;
 using ::curvefs::metaserver::DentryFlag;
 using ::curvefs::mds::topology::PartitionTxId;
 
@@ -35,6 +37,7 @@ using ::curvefs::mds::topology::PartitionTxId;
                << ", DebugString = " << DebugString();
 
 RenameOperator::RenameOperator(uint32_t fsId,
+                               const std::string& fsName,
                                uint64_t parentId,
                                std::string name,
                                uint64_t newParentId,
@@ -42,8 +45,10 @@ RenameOperator::RenameOperator(uint32_t fsId,
                                std::shared_ptr<DentryCacheManager> dentryManager,  // NOLINT
                                std::shared_ptr<InodeCacheManager> inodeManager,
                                std::shared_ptr<MetaServerClient> metaClient,
-                               std::shared_ptr<MdsClient> mdsClient)
+                               std::shared_ptr<MdsClient> mdsClient,
+                               bool enableParallel)
     : fsId_(fsId),
+      fsName_(fsName),
       parentId_(parentId),
       name_(name),
       newParentId_(newParentId),
@@ -57,11 +62,15 @@ RenameOperator::RenameOperator(uint32_t fsId,
       dentryManager_(dentryManager),
       inodeManager_(inodeManager),
       metaClient_(metaClient),
-      mdsClient_(mdsClient) {}
+      mdsClient_(mdsClient),
+      enableParallel_(enableParallel),
+      uuid_(),
+      sequence_(0) {}
 
 std::string RenameOperator::DebugString() {
     std::ostringstream os;
     os << "( fsId = " << fsId_
+       << ", fsName = " << fsName_
        << ", parentId = " << parentId_ << ", name = " << name_
        << ", newParentId = " << newParentId_ << ", newname = " << newname_
        << ", srcPartitionId = " << srcPartitionId_
@@ -72,7 +81,9 @@ std::string RenameOperator::DebugString() {
        << ", dstDentry = [" << dstDentry_.ShortDebugString() << "]"
        << ", prepare dentry = [" << dentry_.ShortDebugString() << "]"
        << ", prepare new dentry = [" << newDentry_.ShortDebugString() << "]"
-       << " )";
+       << ", enableParallel = " << enableParallel_
+       << ", uuid = " << uuid_
+       << ", sequence = " << sequence_ << ")";
     return os.str();
 }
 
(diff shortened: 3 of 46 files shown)


Wine93:
recheck

curvefs/src/mds/dlock/BUILD
```diff
@@ -0,0 +1,33 @@
+#
+#  Copyright (c) 2021 NetEase Inc.
```
SeanHai, line 2: 2022 check all files created.

curvefs/src/mds/dlock/dlock.cpp
```diff
@@ -0,0 +1,214 @@
+/*
+ *  Copyright (c) 2021 NetEase Inc.
```
SeanHai, line 2: ditto

curvefs/conf/client.conf
```diff
@@ -88,6 +88,10 @@ fuseClient.dCacheLruSize=65536
 fuseClient.enableICacheMetrics=true
 fuseClient.enableDCacheMetrics=true
 fuseClient.cto=true
+# You shoudle enable it when mount one filesystem to multi mountpoints,
+# it gurantee the consistent of file after rename, otherwise you shoudle
+# disable it for performance.
+fuseClient.enableMultiMountPointRename=false
```
wu-hanqing, line 94: if you set this config to false, should you prevent current filesystem mount again?

and also, should this config be consistent with fuseClient.cto? which means if cto is true, this config must be true.

curvefs/proto/mds.proto
```diff
@@ -169,6 +172,36 @@ message RefreshSessionResponse {
...
+message DLockValue {
+    required string owner = 1;
+    required uint64 expireTime = 2;
+}
+
+message GetLatestTxIdRequest {
+    optional bool lock = 1;
+    optional uint32 fsId = 2;
+    optional string fsName = 3;
+    optional string uuid = 4;
```
wu-hanqing, line 184: why all the fields is optional ?

curvefs/proto/mds.proto
```diff
@@ -181,6 +214,9 @@ service MdsService {
     rpc AllocateS3Chunk(AllocateS3ChunkRequest) returns (AllocateS3ChunkResponse);
     rpc ListClusterFsInfo (ListClusterFsInfoRequest) returns (ListClusterFsInfoResponse);
 
+    rpc GetLatestTxId(GetLatestTxIdRequest) returns (GetLatestTxIdResponse);
+    rpc CommitTx(CommitTxRequest) returns (CommitTxResponse);
```
wu-hanqing, line 218: if you move a rpc service, there will be a compatible issue.

curvefs/proto/topology.proto
```diff
@@ -471,7 +462,6 @@ service TopologyService {
     rpc CreatePartition(CreatePartitionRequest) returns (CreatePartitionResponse);
     rpc ListPartition(ListPartitionRequest) returns (ListPartitionResponse);
     rpc GetCopysetOfPartition(GetCopysetOfPartitionRequest) returns (GetCopysetOfPartitionResponse);
-    rpc CommitTx(CommitTxRequest) returns (CommitTxResponse);
```
wu-hanqing, line 474: ditto

curvefs/src/mds/dlock/dlock.cpp
```diff
@@ -0,0 +1,214 @@
...
+              << "total=" << total << ", step=" << step << ", key=" << key;
+    for (uint64_t elapsed = 0; elapsed < total; elapsed += step) {
+        {
+            NameLockGuard lg(nameLock_, name);
+            if (ReplaceOwner(name, owner, options_.ttlMs)) {
+                succ = true;
+                break;
+            }
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(step));
```
wu-hanqing, line 88: you should use `bthread_usleep`, because the caller is within a bthread.

curvefs/src/mds/fs_manager.cpp
```diff
@@ -619,5 +621,114 @@ FSStatusCode FsManager::ReloadMountedFsVolumeSpace() {
     return FSStatusCode::OK;
 }
 
+FSStatusCode FsManager::IncreaseFsTxSequence(const std::string& fsName,
```
wu-hanqing, line 624: I think it's better to extract rename-related things into one place because some of them are in topology, while some others are in fsinfo.

curvefs/conf/client.conf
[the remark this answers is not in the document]
Wine93: > I think the default value should be true, otherwiths error wiil be occured at defalt configure.

fixed.

curvefs/src/client/client_operator.cpp
```diff
@@ -163,6 +198,7 @@ CURVEFS_ERROR RenameOperator::RecordOldInodeInfo() {
 }
 
 CURVEFS_ERROR RenameOperator::PrepareRenameTx(
+    // TODO(@Wine93): added sequence
```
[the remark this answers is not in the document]
Wine93: > Already add and delete it?

fixed.

curvefs/src/client/client_operator.cpp
```diff
@@ -35,15 +37,18 @@ using ::curvefs::mds::topology::PartitionTxId;
                << ", DebugString = " << DebugString();
 
 RenameOperator::RenameOperator(uint32_t fsId,
+                               const std::string& fsname,
```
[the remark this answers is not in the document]
Wine93: > Already have fsId, why need fsname?

we need to lock the namespace of `fsName`, if we only pass fsId to `MDS`, we should ne more operation to get `fsName` by `fsId`.

> fsname to fsName

fixed

curvefs/proto/mds.proto
```diff
@@ -169,6 +172,36 @@ message RefreshSessionResponse {
...
+message GetLatestTxIdRequest {
+    optional bool lock = 1;
+    optional uint32 fsId = 2;
+    optional string fsName = 3;
+    optional string uuid = 4;
+}
+
+message GetLatestTxIdResponse {
+    required FSStatusCode statusCode = 1;
+    repeated topology.PartitionTxId txIds = 2;
```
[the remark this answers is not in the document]
Wine93: > Move message 'PartitionTxId' to mds.proto?

fixed.

curvefs/src/mds/dlock/BUILD
Wine93: > 2022 check all files created.

fixed.

curvefs/src/mds/dlock/dlock.cpp
Wine93: > ditto

fixed.

curvefs/src/client/rpcclient/metacache.cpp
```diff
@@ -49,6 +50,19 @@ void MetaCache::GetTxId(uint32_t partitionId, uint64_t *txId) {
     }
 }
 
+bool MetaCache::GetTxId(uint32_t fsId, uint64_t inodeId, uint32_t *partitionId,
+                        uint64_t *txId) {
+    for (const auto &partition : partitionInfos_) {
+        if (inodeId >= partition.start() && inodeId <= partition.end()) {
```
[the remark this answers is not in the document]
Wine93: > Should check fsId in partition, different fs has same inodeId.

The `MetaCache` object only belongs to one filesystem(fsId), the partition in `partitionInfos_` is also belongs to the one.

curvefs/src/mds/dlock/dlock.cpp
Wine93: > you should use `bthread_usleep`, because the caller is within a bthread.

fixed.

curvefs/src/client/common/config.h
```diff
@@ -66,6 +66,7 @@ struct ExcutorOpt {
     uint64_t minRetryTimesForceTimeoutBackoff = 5;
     uint64_t maxRetryTimesBeforeConsiderSuspend = 20;
     uint32_t batchLimit = 100;
+    bool enableRenameParallel = false;
```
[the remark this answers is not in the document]
Wine93: > I think the defalut value should be true.

I had set the default value of `fuseClient.enableMultiMountPointRename` to `true`, if we set the RPC configure item to true, it will cause confusion for people who invoke rpc, because we will get latest txid for every dentry related RPC 
 request when the  `enableRenameParallel ` to true.

curvefs/src/client/rpcclient/mds_client.cpp
```diff
@@ -474,6 +447,105 @@ MdsClientImpl::RefreshSession(const std::vector<PartitionTxId> &txIds,
     return ReturnError(rpcexcutor_.DoRPCTask(task, mdsOpt_.mdsMaxRetryMS));
 }
 
+FSStatusCode MdsClientImpl::GetLatestTxId(const GetLatestTxIdRequest& request,
+                                          GetLatestTxIdResponse* response) {
+    auto task = RPCTask {
+        mdsbasecli_->GetLatestTxId(request, response, cntl, channel);
```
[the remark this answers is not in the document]
Wine93: > Add mdsClientMetric_.GetLatestTxId metric.

fixed.

curvefs/conf/client.conf
Wine93: > if you set this config to false, should you prevent current filesystem mount again?
> 
> and also, should this config be consistent with fuseClient.cto? which means if cto is true, this config must be true.

i had set the default value of `fuseClient.enableMultiMountPointRename` to `true`, if user want to improve 
 the performance in single mount point, should disable it.

curvefs/proto/mds.proto
Wine93: > why all the fields is optional ?

there are two ways to invoke `CommitTX` RPC, one is lock, one is commit directly and the another one is commit with lock.

curvefs/proto/mds.proto
Wine93: > if you move a rpc service, there will be a compatible issue.

I will keep both of these 2 methods for compatibility.

curvefs/proto/topology.proto
Wine93: > ditto

fixed.

curvefs/src/mds/fs_manager.cpp
Wine93: > I think it's better to extract rename-related things into one place because some of them are in topology, while some others are in fsinfo.

i think it's inevitable, because the `txId` is belongs to partition, so we will store it in topology, but the `txSequence` is belongs to filesystem which is globally incremented, so we must store it in `fsinfo`.

curvefs/proto/mds.proto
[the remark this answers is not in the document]
SeanHai: Compatibility has been broken when and required field in proto. (For example txSequence txOwner in FsInfo)

curvefs/src/client/client_operator.cpp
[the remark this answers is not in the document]
SeanHai: The fsId is also unique. Can fsId be locked replace fsName？

curvefs/src/client/rpcclient/metacache.cpp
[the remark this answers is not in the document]
SeanHai: If fsId needn't delete it.

curvefs/src/client/common/config.h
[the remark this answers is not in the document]
SeanHai: If someone not set 'fuseClient.enableMultiMountPointRename', the value will be false and error may occur when rename at multile mount point.

curvefs/src/client/client_operator.cpp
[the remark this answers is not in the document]
Wine93: > The fsId is also unique. Can fsId be locked replace fsName？

not yet~ because other operator lock FS by fsname in `FsManager`:

```
FSStatusCode FsManager::DeleteFs(const std::string& fsName) {
    NameLockGuard lock(nameLock_, fsName);
   ...
}
```

curvefs/src/client/common/config.h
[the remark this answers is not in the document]
Wine93: > If someone not set 'fuseClient.enableMultiMountPointRename', the value will be false and error may occur when rename at multile mount point.

not yet, there is no effect for other RPC request.

curvefs/src/client/rpcclient/metacache.cpp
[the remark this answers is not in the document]
Wine93: > If fsId needn't delete it.

Maybe the fsId is needed for prevent get partition which not belongs to the `fsId`.

curvefs/proto/mds.proto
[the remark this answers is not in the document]
Wine93: > Compatibility has been broken when and required field in proto. (For example txSequence txOwner in FsInfo)

yep, i had convert it to option.

wu-hanqing (approved):

SeanHai (approved):

PR #1450, opened 2022-05-16
curvefs/topology: add fs inodeNum metric

Cyber-SiKu:

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/mds/topology/topology_metric.cpp
+++ b/curvefs/src/mds/topology/topology_metric.cpp
@@ -22,16 +22,21 @@
 
 #include "curvefs/src/mds/topology/topology_metric.h"
 
+#include <algorithm>
 #include <list>
 #include <memory>
 #include <set>
+#include <unordered_map>
+#include <utility>
 
 namespace curvefs {
 namespace mds {
 namespace topology {
 
+using FsMetricPtr = std::unique_ptr<FsMetric>;
 std::map<PoolIdType, PoolMetricPtr> gPoolMetrics;
 std::map<MetaServerIdType, MetaServerMetricPtr> gMetaServerMetrics;
+std::map<FsIdType, FsMetricPtr> gFsMetrics;
 
 void TopologyMetricService::UpdateTopologyMetrics() {
     // process metaserver
@@ -61,6 +66,7 @@ void TopologyMetricService::UpdateTopologyMetrics() {
         }
     }
 
+    std::unordered_map<FsIdType, uint64_t> fsId2InodeNum;
     // process pool
     std::vector<PoolIdType> pools = topo_->GetPoolInCluster();
     for (auto pid : pools) {
@@ -83,6 +89,14 @@ void TopologyMetricService::UpdateTopologyMetrics() {
         for (auto pit = partitions.begin(); pit != partitions.end(); ++pit) {
             totalInodeNum += pit->GetInodeNum();
             totalDentryNum += pit->GetDentryNum();
+            // update fs2inodeNum
+            auto fsId = pit->GetFsId();
+            auto it = fsId2InodeNum.find(fsId);
+            if (it == fsId2InodeNum.end()) {
+                fsId2InodeNum.emplace(fsId, pit->GetInodeNum());
+            } else {
+                it->second += pit->GetInodeNum();
+            }
         }
         it->second->inodeNum.set_value(totalInodeNum);
         it->second->dentryNum.set_value(totalDentryNum);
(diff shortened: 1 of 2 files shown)


curvefs/src/mds/topology/topology_metric.h
```diff
@@ -187,6 +187,22 @@ class TopologyMetricService {
...
+    bvar::Status<uint64_t> inodeNum_;
+
+    explicit FsMetric(FsIdType fsId)
+        : inodeNum_(kTopologyFsMetricPrefix,
+                    std::to_string(fsId) + "_inode_num", 0) {}
+};
+
+using FsMetricPtr = std::unique_ptr<FsMetric>;
+
+extern std::map<FsIdType, FsMetricPtr> gFsMetrics;
```
wu-hanqing, line 204: there is no need to put a declaration here because no other translation unit uses it.

curvefs/src/mds/topology/topology_metric.cpp
```diff
@@ -61,6 +62,11 @@ void TopologyMetricService::UpdateTopologyMetrics() {
         }
     }
 
+    // clear fs inodeNum metric
+    for (auto &it : gFsMetrics) {
+        it.second->inodeNum_.set_value(0);
+    }
+
```
wu-hanqing, line 69: it seems it's no sense.

curvefs/src/mds/topology/topology_metric.h
```diff
@@ -187,6 +187,22 @@ class TopologyMetricService {
...
+    // inode number
+
+    bvar::Status<uint64_t> inodeNum_;
+
+    explicit FsMetric(FsIdType fsId)
+        : inodeNum_(kTopologyFsMetricPrefix,
+                    std::to_string(fsId) + "_inode_num", 0) {}
+};
+
+using FsMetricPtr = std::unique_ptr<FsMetric>;
```
wu-hanqing, line 202: ditto, remove it.

curvefs/src/mds/topology/topology_metric.cpp
```diff
@@ -83,6 +89,15 @@ void TopologyMetricService::UpdateTopologyMetrics() {
...
             totalDentryNum += pit->GetDentryNum();
+            // update fs inodeNum metrics
+            auto fsId = pit->GetFsId();
+            auto it = gFsMetrics.find(fsId);
+            if (it == gFsMetrics.end()) {
+                FsMetricPtr fptr(new FsMetric(fsId));
+                it = gFsMetrics.emplace(fsId, std::move(fptr)).first;
+            }
+            it->second->inodeNum_.set_value(pit->GetInodeNum() +
+                                            it->second->inodeNum_.get_value());
```
wu-hanqing, line 100: I think you can use a map<FsId *fsid*, uint64_t *total inode number*> or unordered_map to record the intermedia value when looping through all the partitions, and after that, call set_value for each fs once.

curvefs/src/mds/topology/topology_metric.cpp
Cyber-SiKu: > it seems it's no sense.

If not set to zero, it will become cumulative

curvefs/src/mds/topology/topology_metric.h
Cyber-SiKu: > there is no need to put a declaration here because no other translation unit uses it.

fix

curvefs/src/mds/topology/topology_metric.h
Cyber-SiKu: > ditto, remove it.

fix

curvefs/src/mds/topology/topology_metric.cpp
Cyber-SiKu: > I think you can use a map<FsId _fsid_, uint64_t _total inode number_> or unordered_map to record the intermedia value when looping through all the partitions, and after that, call set_value for each fs once.

fix

cw123:
the ci is not pass

curvefs/src/mds/topology/topology_metric.h
```diff
@@ -187,6 +187,18 @@ class TopologyMetricService {
     TopologyOption option_;
 };
 
+struct FsMetric {
+ public:
+    const std::string kTopologyFsMetricPrefix = "topology_Fs_id_";
```
cw123, line 192: fs name也带上

Cyber-SiKu: > fs name也带上

fsname is not suitable, right? If there is an inodeNum fs, it is easy to be confused.

cw123 (approved):

wu-hanqing (approved):

Issue #1458, opened 2022-05-17
curvefs mount point exist check is invalid


cw123:
**Describe the bug (描述bug)**
现在curvefs的fs在挂载的时候会加上mount point是否已经exist的判断，以避免一个挂载点多次挂载。现在curvefs的mount point的挂载点加上了port信息，原来的判断mount point是否已经存在的方式就失效了。见下面的截图。
Now the fs of curvefs will add the judgment of whether the mount point has existed when it is mounted, so as to avoid multiple mounts of a mount point. Now that the mount point of the curvefs mount point is added with port information, the original method of judging whether the mount point already exists is invalid. See screenshot below.

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**
![image](https://user-images.githubusercontent.com/4277834/168782087-3a717867-d88c-401c-9d82-d5393fc6126c.png)
![image](https://user-images.githubusercontent.com/4277834/168782155-e66168a5-a9d9-4750-aa59-e3c6078ffa12.png)

PR #1472, opened 2022-05-19
curvefs: fix check mountpoint exist error

Cyber-SiKu:

Issue Number: #1458  

Problem Summary:

What's Changed:

1. Use message moupoint instead of string
2. fix monitor unit error

How it Works:

### Check List

### DIFF
--- a/curvefs/proto/mds.proto
+++ b/curvefs/proto/mds.proto
@@ -72,6 +72,12 @@ message FsDetail {
     optional common.S3Info s3Info = 2;
 }
 
+message Mountpoint {
+    required string hostname = 1;
+    required uint32 port = 2;
+    required string path = 3;
+}
+
 message FsInfo {
     required uint32 fsId = 1;
     required string fsName = 2;
@@ -80,7 +86,7 @@ message FsInfo {
     required uint64 capacity = 5;
     required uint64 blockSize = 6;
     required uint32 mountNum = 7;
-    repeated string mountpoints = 8;
+    repeated Mountpoint mountpoints = 8;
     required common.FSType fsType = 9;
     required FsDetail detail = 10;
     required bool enableSumInDir = 11;
@@ -111,7 +117,7 @@ message CreateFsResponse {
 
 message MountFsRequest {
     required string fsName = 1;
-    required string mountpoint = 2;
+    required Mountpoint mountpoint = 2;
 }
 
 message MountFsResponse {
@@ -121,7 +127,7 @@ message MountFsResponse {
 
 message UmountFsRequest {
     required string fsName = 1;
-    required string mountpoint = 2;
+    required Mountpoint mountpoint = 2;
 }
 
 message UmountFsResponse {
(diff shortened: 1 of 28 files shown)


Cyber-SiKu:
recheck

Cyber-SiKu:
recheck

curvefs/src/mds/fs_info_wrapper.cpp
```diff
@@ -33,23 +35,30 @@ namespace mds {
...
 using ::curvefs::common::S3Info;
 using ::curvefs::common::Volume;
-
-bool FsInfoWrapper::IsMountPointExist(const std::string& mp) const {
-    return std::find(fsInfo_.mountpoints().begin(), fsInfo_.mountpoints().end(),
-                     mp) != fsInfo_.mountpoints().end();
+using google::protobuf::util::MessageDifferencer;
+
+bool FsInfoWrapper::IsMountPointExist(const Mountpoint& mp) const {
+    return std::find_if(fsInfo_.mountpoints().begin(),
```
cw123, line 41: If hostname and path is same, but port is different, it will return exist ?

Cyber-SiKu: > If hostname and path is same, but port is different, it will return exist ?

Sorry, I changed a lot here, it has been fixed

curvefs/proto/mds.proto
```diff
@@ -80,7 +86,7 @@ message FsInfo {
     required uint64 capacity = 5;
     required uint64 blockSize = 6;
     required uint32 mountNum = 7;
-    repeated string mountpoints = 8;
+    repeated Mountpoint mountpoints = 8;
```
wu-hanqing, line 89: this will result in a compatible problem

curvefs/src/mds/mds_service.cpp
```diff
@@ -163,23 +166,23 @@ void MdsServiceImpl::MountFs(::google::protobuf::RpcController* controller,
     brpc::ClosureGuard doneGuard(done);
     brpc::Controller* cntl = static_cast<brpc::Controller*>(controller);
     std::string fsName = request->fsname();
-    std::string mount = request->mountpoint();
+    Mountpoint mount = request->mountpoint();
```
wu-hanqing, line 169: use reference to avoid creating temporary object

curvefs/src/mds/mds_service.cpp
```diff
@@ -191,20 +194,20 @@ void MdsServiceImpl::UmountFs(::google::protobuf::RpcController* controller,
...
```
wu-hanqing, line 197: ditto

Cyber-SiKu: > ditto

fix

curvefs/proto/mds.proto
Cyber-SiKu: > this will result in a compatible problem

Remounting solves it

wu-hanqing (approved):

cw123 (approved):

Issue #1427, opened 2022-05-12
curvefs getfattr fail and retry 1000000 times


cw123:
**Describe the bug (描述bug)**
在curvefs的挂载目录下执行getfattr花了23min之后失败。日志里面有重试100万次，然后失败。失败的截图及client相关日志见截图。
这个集群，已经有2000多万文件，规模比较大。是否和规模有关。

It failed after 23 minutes to execute getfattr in the mount directory of curvefs. There are 1 million retries in the log, and then it fails. See screenshots for failed screenshots and client-related logs.
This cluster already has more than 20 million files, and the scale is relatively large. Is it related to size.

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS: linux
Compiler:
branch: master
commit id: 5c985ffb

**Additional context/screenshots (更多上下文/截图)**
![image](https://user-images.githubusercontent.com/4277834/168034638-3aa96703-1b04-4a18-9b76-89972613aa48.png)
![image](https://user-images.githubusercontent.com/4277834/168034660-18952f5a-230f-417a-9f25-0a7ad65c2acb.png)

SeanHai:
尝试复现：再次跑这个命令getfattr -n curve.dir.rfbytes -d ILSVRC2012_img_val_dir
1、刚开始一直超时8s：http://10.166.16.62:7801/rpcz 看不到BatchGetInodeAttr请求
<img width="1322" alt="截屏2022-05-12 20 47 42" src="https://user-images.githubusercontent.com/13496900/168078377-4ea60892-34c1-4d87-88de-faca8ab971ab.png">

2、一段时间后，response返回后未携带inodeAttr；http://10.166.16.62:7801/rpcz 可以看到BatchGetInodeAttr请求
<img width="1103" alt="截屏2022-05-12 20 50 53" src="https://user-images.githubusercontent.com/13496900/168078775-a6282a1d-0b4b-4902-96e3-fd848520d196.png">

SeanHai:
接下来可测：
1、将该文件系统再挂载一个新的挂载点，再次测试看下结果。
2、如果1可复现，可在对应clent版本上加点日志，看下每次请求是否携带inodeId（首次发送肯定携带了，不然不会发送rpc，主要加在rpc重试过程，看下是否在重试过程中request发生变化）。
3、如果2可看到每次都携带了inodeId，则需要在metaserver端加日志，升级服务端再测试。

SeanHai:
[...]

PR #1474, opened 2022-05-20
curvefs/client: fix the data iteration error when rpc retry.

SeanHai:

Issue Number: #xxx 
#1427
Problem Summary:
The before logic will crash when rcp retry, the iter will forward when retry but this is not expected.
截屏2022-05-20 14 37 07

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/rpcclient/metaserver_client.cpp
+++ b/curvefs/src/client/rpcclient/metaserver_client.cpp
@@ -477,14 +477,14 @@ MetaStatusCode MetaServerClientImpl::GetInode(uint32_t fsId, uint64_t inodeid,
 bool GroupInodeIdByPartition(
     uint32_t fsId,
     std::shared_ptr<MetaCache> metaCache,
-    std::set<uint64_t> *inodeIds,
-    std::unordered_map<uint32_t, std::list<uint64_t>> *inodeGroups) {
-    for (const auto &it : *inodeIds) {
+    const std::set<uint64_t> &inodeIds,
+    std::unordered_map<uint32_t, std::vector<uint64_t>> *inodeGroups) {
+    for (const auto &it : inodeIds) {
         uint32_t pId = 0;
         if (metaCache->GetPartitionIdByInodeId(fsId, it, &pId)) {
             auto iter = inodeGroups->find(pId);
             if (iter == inodeGroups->end()) {
-                inodeGroups->emplace(pId, std::list<uint64_t>({it}));
+                inodeGroups->emplace(pId, std::vector<uint64_t>({it}));
             } else {
                 iter->second.push_back(it);
             }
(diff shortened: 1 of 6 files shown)


curvefs/src/client/rpcclient/metaserver_client.cpp
```diff
@@ -497,163 +497,185 @@ bool GroupInodeIdByPartition(
     return true;
 }
 
+bool MetaServerClientImpl::SplitRequestInodes(
+    uint32_t fsId,
+    const std::set<uint64_t> &inodeIds,
+    std::vector<std::list<uint64_t>> *inodeGroups) {
+    std::unordered_map<uint32_t, std::list<uint64_t>> groups;
```
wu-hanqing, line 504: for most cases, `std::vector` is always preferred, std::list/map/set are all node-based containers, they're not cache-friendly.

for `std::vector`, another tip is if you know the number of elements that it will store, you can call `vec.reserve` firstly to allocate enough memory.

curvefs/src/client/rpcclient/metaserver_client.cpp
```diff
@@ -497,163 +497,185 @@ bool GroupInodeIdByPartition(
...
+    bool ret = GroupInodeIdByPartition(fsId, metaCache_, inodeIds, &groups);
+    if (ret) {
+        for (const auto &it : groups) {
```
wu-hanqing, line 507:
```cpp
if (!ret) {
    return false;
}

for (...
```

curvefs/src/client/rpcclient/metaserver_client.cpp
```diff
@@ -497,163 +497,185 @@ bool GroupInodeIdByPartition(
...
+            request.set_poolid(poolID);
+            request.set_copysetid(copysetID);
+            request.set_partitionid(partitionID);
+            request.set_fsid(fsId);
+            request.set_appliedindex(
+                metaCache_->GetApplyIndex(CopysetGroupID(poolID,
+                                            copysetID)));
+            for (const auto &id : it) {
+                request.add_inodeid(id);
+            }
```
wu-hanqing, line 553: might be a shortcut, `*request.mutable_inode() = { it.begin(), it.end() }`

curvefs/src/client/rpcclient/metaserver_client.cpp
```diff
@@ -497,163 +497,185 @@ bool GroupInodeIdByPartition(
...
+            if (request.inodeid_size() <= 0) {
```
wu-hanqing, line 554: should you check whether `it` is empty or not at the beginning? otherwise L537 will be core dumped.

curvefs/src/client/rpcclient/metaserver_client.cpp
```diff
@@ -497,163 +497,185 @@ bool GroupInodeIdByPartition(
...
+            MetaStatusCode ret = response.statuscode();
             if (ret != MetaStatusCode::OK) {
-                attr->clear();
-                return ret;
+                LOG_IF(WARNING, ret != MetaStatusCode::NOT_FOUND)
+                    << "BatchGetInodeAttr failed, errcode = " << ret
+                    << ", errmsg = " << MetaStatusCode_Name(ret);
+            } else if (response.attr_size() > 0 &&
+                        response.has_appliedindex()) {
+                auto retAttr = response.attr();
```
wu-hanqing, line 578: avoid unnecessary copy, use reference here.

curvefs/src/client/rpcclient/metaserver_client.cpp
```diff
@@ -497,163 +497,185 @@ bool GroupInodeIdByPartition(
...
-                xattr->clear();
-                return ret;
+                LOG_IF(WARNING, ret != MetaStatusCode::NOT_FOUND)
+                    << "BatchGetXAttr failed, errcode = " << ret
+                    << ", errmsg = " << MetaStatusCode_Name(ret);
+            } else if (response.xattr_size() > 0 &&
+                        response.has_appliedindex()) {
+                auto retXattr = response.xattr();
+                for_each(retXattr.begin(), retXattr.end(),
+                    [&](XAttr &a) { xattr->push_back(a); });
```
wu-hanqing, line 659:
```cpp
auto* attrs = response.mutable_xattr();
xattr->insert(xattr->end(),
              std::make_move_iterator(attrs->begin()),
              std::make_move_iterator(attrs->end()));
```

might be a better way to just move xattrs in response to the destination.

curvefs/src/client/rpcclient/metaserver_client.cpp
```diff
@@ -497,163 +497,185 @@ bool GroupInodeIdByPartition(
...
-                                 << cntl->ErrorCode()
-                                 << ", error content:" << cntl->ErrorText()
-                                 << ", log id = " << cntl->log_id();
-                    return -cntl->ErrorCode();
-                }
+        uint64_t inodeId = *it.begin();
+        auto task = RPCTask {
+            metaserverClientMetric_->batchGetXattr.qps.count << 1;
+            LatencyUpdater updater(
+                &metaserverClientMetric_->batchGetXattr.latency);
```
wu-hanqing, line 620: the overall process is still serial, can we send requests parallelly?

curvefs/src/client/rpcclient/metaserver_client.cpp
SeanHai: fixed

curvefs/src/client/rpcclient/metaserver_client.cpp
SeanHai: fixed

curvefs/src/client/rpcclient/metaserver_client.cpp
SeanHai: fixed

curvefs/src/client/rpcclient/metaserver_client.cpp
SeanHai: fixed

curvefs/src/client/rpcclient/metaserver_client.cpp
SeanHai: fixed

cw123 (approved):

curvefs/src/client/rpcclient/metaserver_client.cpp
SeanHai: fixed

curvefs/src/client/rpcclient/metaserver_client.cpp
SeanHai: Yes, add a TODO and improve this in later pr.

wu-hanqing (approved):

Issue #1495, opened 2022-05-24
When DiskCacheType::Disable, S3ClientAdaptorImpl::BackGroundFlush is disabled


ilixiaocui:
**Describe the bug (描述bug)**
![image](https://user-images.githubusercontent.com/15689619/170000590-ecc23438-5ab9-4ecd-8bef-e437b0d0fe05.png)
![image](https://user-images.githubusercontent.com/15689619/170000612-b4072b06-a050-4f63-8673-5fbfb53e028e.png)

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch: master, release2.1, release2.2
commit id:

**Additional context/screenshots (更多上下文/截图)**

PR #1496, opened 2022-05-24
curvefs/client: Fix the problem of not opening diskcache datacache not flush

baijiaruo:

Issue Number: #1495  

Problem Summary:

What's Changed:

How it Works:

### Check List


curvefs/src/client/fuse_s3_client.cpp
```diff
@@ -71,7 +71,7 @@ CURVEFS_ERROR FuseS3Client::Init(const FuseClientOption &option) {
     } else {
         ret = s3Adaptor_->Init(opt.s3Opt.s3ClientAdaptorOpt, s3Client,
                                inodeManager_, mdsClient_, fsCacheManager,
-                               nullptr);
+                               nullptr, true);
```
ilixiaocui, line 74: 把默认值改为true吧
Change the default value to true

wuhongsong (approved):

ilixiaocui (approved):

Issue #1468, opened 2022-05-18
curvefs load copyset will stuck heartbeat


cw123:
**Describe the bug (描述bug)**
mds heart beat can not find leader
`W 2022-05-18T19:19:06.843816+0800   128 heartbeat_manager.cpp:233] leader not found, poolid: 1, copysetid: 8`

but copyset status is ok

> statusCode: TOPO_OK copysetInfo { poolId: 1 copysetId: 8 peers { id: 3 address: "10.166.16.60:6711:0" } peers { id: 6 address: "10.166.16.62:6710:0" } peers { id: 8 address: "10.166.16.61:6711:0" } epoch: 2 leaderPeer { id: 3 address: "10.166.16.60:6711:0" } }
> -status:
> status: COPYSET_OP_STATUS_SUCCESS copysetStatus { state: 1 peer { address: "10.166.16.60:6711:0" } leader { address: "10.166.16.60:6711:0" } readonly: false term: 3 committedIndex: 75058029 knownAppliedIndex: 75058029 pendingIndex: 0 pendingQueueSize: 0 applyingIndex: 0 firstIndex: 75014914 lastIndex: 75058029 diskIndex: 75058029 epoch: 2 }
> status: COPYSET_OP_STATUS_SUCCESS copysetStatus { state: 4 peer { address: "10.166.16.61:6711:0" } leader { address: "10.166.16.60:6711:0" } readonly: false term: 3 committedIndex: 75058029 knownAppliedIndex: 75058029 pendingIndex: 0 pendingQueueSize: 0 applyingIndex: 0 firstIndex: 75010129 lastIndex: 75058029 diskIndex: 75058029 epoch: 2 }
> status: COPYSET_OP_STATUS_SUCCESS copysetStatus { state: 4 peer { address: "10.166.16.62:6710:0" } leader { address: "10.166.16.60:6711:0" } readonly: false term: 3 committedIndex: 75058029 knownAppliedIndex: 75058029 pendingIndex: 0 pendingQueueSize: 0 applyingIndex: 0 firstIndex: 75016478 lastIndex: 75058029 diskIndex: 75058029 epoch: 2 }

and recover is scheduling

> I 2022-05-18T19:40:45.113906+0800 89 recoverScheduler.cpp:34] recoverScheduler begin.
[...]

PR #1505, opened 2022-05-25
curvefs fix load copyset stuck heartbeat #1468

cw123:

Issue Number: #1468

Problem Summary: curvefs fix load copyset stuck heartbeat #1468

When copyset is loading, it will stuck heartbeat when get partition info list. I make some changes.
1. When copyset is loading, heartbeat do not send partition info in copyset.
2. When copyset get partition, use try lock instead of lock.
3. Do not use heartbeat partition num to update topo.

### DIFF
--- a/curvefs/proto/heartbeat.proto
+++ b/curvefs/proto/heartbeat.proto
@@ -37,7 +37,8 @@ message CopySetInfo {
     required common.Peer leaderPeer = 5;
     repeated common.PartitionInfo partitionInfoList = 6;
     optional ConfigChangeInfo configChangeInfo = 7;
-    // required CopySetState state = 8;
+    optional bool isCopysetLoading = 8;
+    // required CopySetState state = 9;
 };
 
 message ConfigChangeInfo {
--- a/curvefs/src/mds/topology/topology_item.h
+++ b/curvefs/src/mds/topology/topology_item.h
@@ -473,7 +473,6 @@ class CopySetInfo {
           copySetId_(UNINITIALIZE_ID),
           leader_(UNINITIALIZE_ID),
           epoch_(0),
-          partitionNum_(0),
           hasCandidate_(false),
           candidate_(UNINITIALIZE_ID),
           dirty_(false),
@@ -484,7 +483,6 @@ class CopySetInfo {
           copySetId_(id),
           leader_(UNINITIALIZE_ID),
           epoch_(0),
-          partitionNum_(0),
           hasCandidate_(false),
           candidate_(UNINITIALIZE_ID),
           dirty_(false),
@@ -496,7 +494,7 @@ class CopySetInfo {
           leader_(v.leader_),
           epoch_(v.epoch_),
           peers_(v.peers_),
-          partitionNum_(v.partitionNum_),
+          partitionIds_(v.partitionIds_),
           hasCandidate_(v.hasCandidate_),
           candidate_(v.candidate_),
           dirty_(v.dirty_),
@@ -511,7 +509,7 @@ class CopySetInfo {
         leader_ = v.leader_;
         epoch_ = v.epoch_;
         peers_ = v.peers_;
-        partitionNum_ = v.partitionNum_;
+        partitionIds_ = v.partitionIds_;
         hasCandidate_ = v.hasCandidate_;
         candidate_ = v.candidate_;
         dirty_ = v.dirty_;
@@ -551,13 +549,7 @@ class CopySetInfo {
 
     bool SetCopySetMembersByJson(const std::string &jsonStr);
 
-    uint64_t GetPartitionNum() const { return partitionNum_; }
-
-    void SetPartitionNum(u_int64_t number) { partitionNum_ = number; }
-
-    void AddPartitionNum() { partitionNum_ += 1; }
-
-    void ReducePartitionNum() { partitionNum_ -= 1; }
+    uint64_t GetPartitionNum() const { return partitionIds_.size(); }
 
     bool HasCandidate() const { return hasCandidate_; }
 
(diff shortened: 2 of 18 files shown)


curvefs/src/metaserver/heartbeat.cpp
```diff
@@ -185,9 +185,20 @@ void Heartbeat::BuildCopysetInfo(curvefs::mds::heartbeat::CopySetInfo *info,
...
-        info->add_partitioninfolist()->CopyFrom(it);
+    if (!isLoading) {
+        std::list<PartitionInfo> partitionInfoList;
+        bool ret = copyset->GetPartitionInfoList(&partitionInfoList);
+        if (ret) {
+            for (auto& it : partitionInfoList) {
+                info->add_partitioninfolist()->CopyFrom(it);
+            }
+        } else {
+            info->set_iscopysetloading(true);
```
wu-hanqing, line 200: log a warning message?

curvefs/src/metaserver/metastore.cpp
```diff
@@ -175,16 +175,28 @@ MetaStatusCode MetaStoreImpl::DeletePartition(
...
+        rwLock_.Unlock();
+        return true;
+    } else {
+        LOG(WARNING) << "metastore GetPartitionInfoList fail, it fail to get"
+                        " the rwLock_";
+        return false;
     }
-    return partitionInfoList;
+
+    return true;
```
wu-hanqing, line 199: it's redundant

curvefs/proto/heartbeat.proto
```diff
@@ -37,7 +37,8 @@ message CopySetInfo {
     required common.Peer leaderPeer = 5;
     repeated common.PartitionInfo partitionInfoList = 6;
     optional ConfigChangeInfo configChangeInfo = 7;
-    // required CopySetState state = 8;
+    optional bool isCopysetLoading = 8;
```
wu-hanqing, line 40: this optional field will introduce problems when the version of MDS and Metaserver are mismatching

curvefs/src/metaserver/copyset/copyset_node.cpp
```diff
@@ -351,10 +352,12 @@ int CopysetNode::on_snapshot_load(braft::SnapshotReader* reader) {
     LOG(INFO) << "Copyset " << name_ << " begin to load snapshot from '"
               << reader->get_path() << "'";
 
+    isLoading_.store(true, std::memory_order_release);
```
wu-hanqing, line 355:
```cpp
struct CopysetLoadingGuard {
    CopysetLoadingGuard(std::atomic& flag) : flag(flag) {
           flag.store(true, ...);
    }

    ~CopysetLoadingGuard() {
          flags.stroe(false, ....);
    }


    std::atomic& flags;
};
```

in this way, you don't have to manually set `isLoading_` to false in every return point.

curvefs/src/mds/topology/topology_item.h
```diff
@@ -548,7 +550,7 @@ class CopySetInfo {
 
     uint64_t GetPartitionNum() const { return partitionNum_; }
 
-    void SetPartitionNum(u_int64_t number) { partitionNum_ = number; }
+    void SetPartitionNum(uint64_t number) { partitionNum_ = number; }
```
SeanHai, line 553: The partitionNum_ field is needn't anymore, remove it and related functions.

curvefs/src/metaserver/copyset/copyset_node.cpp
SeanHai: agree

curvefs/src/metaserver/copyset/copyset_node.cpp
```diff
@@ -540,8 +544,22 @@ void CopysetNode::ListPeers(std::vector<Peer>* peers) const {
...
 
-std::list<PartitionInfo> CopysetNode::GetPartitionInfoList() {
-    return metaStore_->GetPartitionInfoList();
+// if copyset is loading, return false;
+// if copyset is not loading, and metastore returns fales, retry;
+// if copyset is not loading, and metastore returns true, return true and get
+// partition info list success.
+bool CopysetNode::GetPartitionInfoList(
+                    std::list<PartitionInfo> *partitionInfoList) {
+    while (true) {
```
SeanHai, line 553: Does the infinite loop logic is ok when metaStore_->GetPartitionInfoList failed？

cw123: Under normal circumstances, one retry can be successful.

curvefs/src/metaserver/heartbeat.cpp
cw123: done

curvefs/src/metaserver/metastore.cpp
cw123: done

curvefs/proto/heartbeat.proto
cw123: done

curvefs/src/metaserver/copyset/copyset_node.cpp
cw123: done

curvefs/src/mds/topology/topology_item.h
cw123: done

cw123:
recheck

curvefs/src/mds/topology/topology_item.h
```diff
@@ -497,6 +497,7 @@ class CopySetInfo {
           epoch_(v.epoch_),
           peers_(v.peers_),
           partitionNum_(v.partitionNum_),
+          partitionIds_(v.partitionIds_),
```
SeanHai, line 500: The partitionNum_ is not needed. In GetPartitionNum can return partitionIds_.size().

cw123: If this is removed, the software will have problems with backward compatibility.

cw123: done, i have removed partitionNum_.

SeanHai (approved):

wu-hanqing (approved):

PR #1509, opened 2022-05-25
the Client Permission 

fansehep:

Issue Number: #xxx 

Problem Summary:

What's Changed: 

How it Works:

### Check List

### DIFF
--- a/curvefs/conf/client.conf
+++ b/curvefs/conf/client.conf
@@ -119,6 +119,8 @@ volume.bitmapallocator.small_alloc_proportion=0.2
 volume.blockgroup.allocate_once=4
 
 #### s3
+# this is for test. if s3.fakeS3=true, all data will be discarded
+s3.fakeS3=false
 # the max size that fuse send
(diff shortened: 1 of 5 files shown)


src/mds/nameserver2/file_record.cpp
```diff
@@ -215,5 +215,15 @@ bool FileRecordManager::FindFileMountPoint(
...
+bool FileRecordManager::SureFileWriterLive(
+    const std::string& fileName, const std::string& ClientIP, uint32_t ClientPort) {
+    auto iter = fileRecords_[fileName];
+    butil::EndPoint ep;
+    int rc = butil::str2endpoint(clientIP.c_str(), ClientPort, &ep);
+    if(iter[ep].IsTimeout() == true) {
+        return true;
+    }
+    return false;
+}
```
wu-hanqing, line 227: file record manager's state is volatile, it only keeps file records in memory, so if MDS failover, the new elected leader MDS doesn't have any file records.

Issue #1503, opened 2022-05-25
The partiton number is out of range in one copyset


SeanHai:
**Describe the bug (描述bug)**
The partition limit of one copyset is 256 default, but created more the 256.

cluster: massive cluster

for i in {1..13};do curvefs_tool list-partition -fsId=$i | grep copysetId;done | awk '{arr[$2]++};END{for(i in arr){print i,arr[i]}}' | sort -nk1
<img width="80" alt="截屏2022-05-25 11 21 54" src="https://user-images.githubusercontent.com/13496900/170173558-50d1a97a-f73f-468b-b1aa-10520f4c2540.png">

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

Issue #1501, opened 2022-05-25
mdtest quits halfway through


ilixiaocui:
**Describe the bug (描述bug)**
sudo mdtest -C -F -L -z 4 -b 10 -I 10000 -d /mnt/fakes3_2/100_million_small_files/mdtest3
<img width="1627" alt="image" src="https://user-images.githubusercontent.com/15689619/170160566-f6e7dacf-6faf-4ffc-b53f-7a959861ad17.png">

client error log
![image](https://user-images.githubusercontent.com/15689619/170160922-3170e48c-92a0-4307-95ed-bf3508e17939.png)

metaserver error log
![image](https://user-images.githubusercontent.com/15689619/170160885-9cbe7d2a-26b1-471d-8cd9-0fb548c22ebe.png)

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch: master with no diskcache and with fakeS3
commit id:

**Additional context/screenshots (更多上下文/截图)**

Wine93:
According to the client logs we found that that some inode and dentry which should belong to the same partition but sended request to different partitions: 

```
CreateInode :  inodeId: 686037139456 (copysetId: 161 partitionId: 40931) -> SUCCESS
CreateDentry:  inodeId: 686037139456 parentInodeId: 1142863953920 name: mdtest_tree.138 type: TYPE_DIRECTORY channel: 10.166.16.60:7802 (copysetId: 267 partitionId: 68158) -> SUCCESS
GetDentry   :  parentInodeId: 686037139456 name: "mdtest_tree.1381" (copysetId: 165 partitionId: 40930) -> NOT_FOUND
CreateInode :  inodeId: 340929806336 (copysetId: 84 partitionId: 20358) -> SUCCESS
CreateDentry:  inodeId: 340929806336 parentInodeId: 686037139456 name: mdtest_tree.1381 type: TYPE_DIRECTORY Channel: 10.166.16.61:7801 (copysetId: 165, partitionId: 40930) -> NOT_FOUND
```

then we list all partitions and found that two different partitions with the same inode range:
```
$ curvefs_tool list-partition -fsId=1

...
partitionInfoList {
  fsId: 1
  poolId: 1
  copysetId: 165
  partitionId: 40930
  start: 686037139456
  end: 686053916671
  txId: 0
  status: READWRITE
  inodeNum: 0
  dentryNum: 1
}
partitionInfoList {
  fsId: 1
[...]

PR #1511, opened 2022-05-26
curvefs/mds: fix create partition error at parallel case

SeanHai:

Issue Number: #xxx 
#1501 #1503

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/mds/topology/topology_manager.cpp
+++ b/curvefs/src/mds/topology/topology_manager.cpp
@@ -632,6 +632,9 @@ void TopologyManager::CreatePartitions(const CreatePartitionRequest *request,
     auto partitionInfoList = response->mutable_partitioninfolist();
     response->set_statuscode(TopoStatusCode::TOPO_OK);
 
+    // get lock and avoid multiMountpoint create concurrently
+    NameLockGuard lock(createPartitionMutex_, std::to_string(fsId));
+
     while (partitionInfoList->size() < count) {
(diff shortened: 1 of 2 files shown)


curvefs/src/mds/topology/topology_manager.cpp
```diff
@@ -632,6 +632,9 @@ void TopologyManager::CreatePartitions(const CreatePartitionRequest *request,
     auto partitionInfoList = response->mutable_partitioninfolist();
     response->set_statuscode(TopoStatusCode::TOPO_OK);
 
+    // get lock and avoid multiMountpoint create concurrently
+    NameLockGuard lock(createPartitionMutex_, std::to_string(fsId));
+
     while (partitionInfoList->size() < count) {
```
cw123, line 638: > uint32_t pNumber = topology_->GetPartitionNumberOfFs(fsId);
> uint64_t idStart = pNumber * option_.idNumberInPartition;
> uint64_t idEnd = (pNumber + 1) * option_.idNumberInPartition - 1;

if delete partition, the start and end will duplicate.
fix it or add a todo here.

SeanHai: This will fix at another pr later.

cw123 (approved):

wu-hanqing (approved):

SeanHai:
recheck

PR #1520, opened 2022-05-30
add perf test scripts

YunhuiChen:

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

Modify the following parameters in the script as required：
  size = '50G'   # test file size
    testpath = '/home/curvefs'   # test path
    filenum = 5 # Example Test the number of concurrent files in the path

### DIFF
--- a/robot/perf_test.py
+++ b/robot/perf_test.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import os
+import time
+
+import subprocess
+import time
+import sys
+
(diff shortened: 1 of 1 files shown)


robot/perf_test.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env python
```
ilixiaocui, line 1: 文件头新增一下备注，本脚本用于测试什么？如何测试？
Add a note to the file header, what is this script used for testing? How to test?

Issue #1526, opened 2022-05-31
Disk read cache elimination of curvefs requires increasing file count limit


ilixiaocui:
**Describe the bug (描述bug)**
**There are two problems here, one is that for the file system, the number of files is limited：**
10:41:27.073 /var/log/messages: May 31 10:41:23 prit2-curve-chunk-node17 kernel: [4650529.004965] EXT4-fs warning (device sdd): ext4_dx_add_entry:2228: inode #100139012: comm curve-fuse: Directory index full!

This inode is the read cache directory with nearly 3000w files below.

**Another problem is that there will be an LRU Cache in the memory to manage the file names in the read cache, and this cache occupies a lot.**
curl localhost:9012/vars  | grep diskcache
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 64893  100 64893    0     0 diskcache_cache_bytes : 822775312-:--:--     0
 7diskcache_cache_count : 27074348
1diskcache_cache_hit : 0
5diskcache_cache_miss : 0
6k      0 --:--:-- --:--:-- --:--:-- 7921k

<img width="327" alt="image" src="https://user-images.githubusercontent.com/15689619/171118115-a9cfd3a2-5cd6-4fc7-bdf6-57f69ca9ba67.png">

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch: master
commit id:

**Additional context/screenshots (更多上下文/截图)**

PR #1530, opened 2022-05-31
curvefs/client: add file limit can cache

wuhongsong:

Issue Number: #1526  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/conf/client.conf
+++ b/curvefs/conf/client.conf
@@ -183,6 +183,8 @@ diskCache.safeRatio=70
 diskCache.threads=5
 # the max size disk cache can use
 diskCache.maxUsableSpaceBytes=107374182400
+# the max files that can cache
+diskCache.maxFileNums=1000000
 # the max time system command can run
 diskCache.cmdTimeoutSec=300
 # directory of disk cache
--- a/curvefs/src/client/s3/disk_cache_manager.cpp
+++ b/curvefs/src/client/s3/disk_cache_manager.cpp
@@ -82,6 +82,7 @@ int DiskCacheManager::Init(S3Client *client,
     safeRatio_ = option.diskCacheOpt.safeRatio;
     cacheDir_ = option.diskCacheOpt.cacheDir;
     maxUsableSpaceBytes_ = option.diskCacheOpt.maxUsableSpaceBytes;
+    maxFileNums_ = option.diskCacheOpt.maxFileNums;
     cmdTimeoutSec_ = option.diskCacheOpt.cmdTimeoutSec;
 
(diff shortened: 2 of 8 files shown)


baijiaruo (approved):

curvefs/conf/client.conf
```diff
@@ -183,6 +183,8 @@ diskCache.safeRatio=70
 diskCache.threads=5
 # the max size disk cache can use
 diskCache.maxUsableSpaceBytes=107374182400
+# the max files that can cache
+diskCache.maxFileNums=3000000
```
ilixiaocui, line 187: Whether this value can be judged according to the type of the current file system, rather than fixed in the configuration file

curvefs/src/client/s3/disk_cache_manager.cpp
```diff
@@ -328,6 +330,16 @@ bool DiskCacheManager::IsDiskCacheSafe() {
     return false;
 }
 
+bool DiskCacheManager::IsExceedFileNums() {
+    uint64_t fileNums = cachedObjName_->Size();
+    if (fileNums >= maxFileNums_) {
+        VLOG(6) << "file nums is: " << fileNums
```
ilixiaocui, line 336: This log is printed once as the error level, which can be alarmed. The specific value can be viewed through metric

ilixiaocui: cachedObjName_ is an LRU, why not do constraints through LRU?

curvefs/conf/client.conf
wuhongsong: the limit of the file system is too large, so it may cost many memory.  and we cannot get  the limit directly

curvefs/src/client/s3/disk_cache_manager.cpp
wuhongsong: todo: https://github.com/opencurve/curve/issues/1534

wuhongsong:
recheck

ilixiaocui:
I tested the case where the LRU size is limited to 10w and found that it will exceed
diskcache_cache_count 100617
diskcache_cache_count 100014

wuhongsong:
> I tested the case where the LRU size is limited to 10w and found that it will exceed diskcache_cache_count 100617 diskcache_cache_count 100014 diskcache_cache_count 100560 diskcache_cache_count 100440

the limit is soft, but we can active trim.

wuhongsong:
recheck

wuhongsong:
recheck

PR #1537, opened 2022-06-02
curvefs: fix filetype inode error in metric

Cyber-SiKu:
The filetype2inode information is stored in partitioninfo, and the
restart of the metaserver will cause a rebuild, so the number will be
abnormal

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/metaserver/inode_manager.cpp
+++ b/curvefs/src/metaserver/inode_manager.cpp
@@ -70,6 +70,7 @@ MetaStatusCode InodeManager::CreateInode(uint64_t inodeId,
     }
 
     newInode->CopyFrom(inode);
+    ++(*type2InodeNum_)[inode.type()];
     VLOG(9) << "CreateInode success, inode = " << inode.ShortDebugString();
     return MetaStatusCode::OK;
 }
@@ -93,7 +94,7 @@ MetaStatusCode InodeManager::CreateRootInode(const InodeParam &param) {
                    << ", ret = " << MetaStatusCode_Name(ret);
         return ret;
     }
-
+    ++(*type2InodeNum_)[inode.type()];
     VLOG(9) << "CreateRootInode success, inode: " << inode.ShortDebugString();
     return MetaStatusCode::OK;
 }
@@ -206,6 +207,15 @@ MetaStatusCode InodeManager::GetXAttr(uint32_t fsId, uint64_t inodeId,
 MetaStatusCode InodeManager::DeleteInode(uint32_t fsId, uint64_t inodeId) {
     VLOG(6) << "DeleteInode, fsId = " << fsId << ", inodeId = " << inodeId;
     NameLockGuard lg(inodeLock_, GetInodeLockName(fsId, inodeId));
+    InodeAttr attr;
+    MetaStatusCode retGetAttr =
+        inodeStorage_->GetAttr(Key4Inode(fsId, inodeId), &attr);
+    if (retGetAttr != MetaStatusCode::OK) {
+        VLOG(9) << "GetInodeAttr fail, fsId = " << fsId
+                   << ", inodeId = " << inodeId
+                   << ", ret = " << MetaStatusCode_Name(retGetAttr);
+    }
+
     MetaStatusCode ret = inodeStorage_->Delete(Key4Inode(fsId, inodeId));
     if (ret != MetaStatusCode::OK) {
         LOG(ERROR) << "DeleteInode fail, fsId = " << fsId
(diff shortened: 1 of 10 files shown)


curvefs/src/metaserver/inode_manager.cpp
```diff
@@ -200,6 +201,8 @@ MetaStatusCode InodeManager::GetXAttr(uint32_t fsId, uint64_t inodeId,
 
 MetaStatusCode InodeManager::DeleteInode(uint32_t fsId, uint64_t inodeId) {
     VLOG(6) << "DeleteInode, fsId = " << fsId << ", inodeId = " << inodeId;
+    InodeAttr attr;
```
cw123, line 204: move it to lock range

curvefs/src/metaserver/inode_manager.cpp
```diff
@@ -306,8 +309,8 @@ MetaStatusCode InodeManager::UpdateInode(const UpdateInodeRequest& request,
     }
 
     if (needAddTrash) {
-        trash_->Add(old->fsid(), old->inodeid(), old->dtime());
-        ++(*deletedNum);
+        trash_->Add(old.fsid(), old.inodeid(), old.dtime());
+        --type2InodeNum_[old.type()];
```
cw123, line 313: delete inode也减了一次，会不会重复减
delete inode is also reduced once, will it be reduced repeatedly?

curvefs/src/metaserver/partition.cpp
```diff
@@ -386,10 +381,6 @@ bool Partition::Clear() {
         LOG(ERROR) << "Clear dentry storage failed";
         return false;
     }
-    partitionInfo_.set_inodenum(0);
```
cw123, line 389: why delete this code

curvefs/src/metaserver/inode_manager.cpp
Cyber-SiKu: > move it to lock range

There is lock in GetInodeAttr, moving to lock range will cause deadlock

curvefs/src/metaserver/inode_manager.cpp
Cyber-SiKu: > delete inode也减了一次，会不会重复减 delete inode is also reduced once, will it be reduced repeatedly?

这里的删除是交给trash后台扫描删除，而trash是直接使用InodeStorage来清理数据，没有包含InodeManager，所以不会调用deletinode。

The deletion here is handed over to trash to scan and delete in the background, and trash directly uses InodeStorage to clean up data. It does not include InodeManager, so deleteinode will not be called.

Cyber-SiKu:
recheck

cw123 (approved):

Cyber-SiKu:
recheck

Cyber-SiKu:
recheck

Cyber-SiKu:
recheck

Cyber-SiKu:
recheck

Cyber-SiKu:
recheck

wuhongsong:
recheck

wuhongsong:
recheck

Cyber-SiKu:
recheck

curvefs/src/metaserver/partition.cpp
```diff
@@ -404,10 +393,9 @@ uint32_t Partition::GetPartitionId() const {
 }
 
 PartitionInfo Partition::GetPartitionInfo() {
-    auto partitionInfo = partitionInfo_;
-    partitionInfo.set_inodenum(GetInodeNum());
-    partitionInfo.set_dentrynum(GetDentryNum());
-    return partitionInfo;
+    partitionInfo_.set_inodenum(GetInodeNum());
```
wu-hanqing, line 396: it seems the caller will GetInodeNum and GetDentryNum after get the partition.

https://github.com/opencurve/curve/blob/feba0744a602a39fddd9addcaf51f96937b793cd/curvefs/src/metaserver/metastore_fstream.cpp#L250

curvefs/test/metaserver/inode_manager_test.cpp
```diff
@@ -72,7 +73,9 @@ class InodeManagerTest : public ::testing::Test {
         auto inodeStorage = std::make_shared<InodeStorage>(
             kvStorage_, nameGenerator, 0);
         auto trash = std::make_shared<TrashImpl>(inodeStorage);
-        manager = std::make_shared<InodeManager>(inodeStorage, trash);
+        map_ = std::make_shared<FileType2InodeNumMap>();
```
wu-hanqing, line 76: `map` can be a data member rather than a pointer, and give a more meaningful name

curvefs/test/metaserver/s3compact/s3compact_test.cpp
```diff
@@ -90,7 +91,9 @@ class S3CompactTest : public ::testing::Test {
         inodeStorage_ =
             std::make_shared<InodeStorage>(kvStorage_, nameGenerator, 0);
         trash_ = std::make_shared<TrashImpl>(inodeStorage_);
-        inodeManager_ = std::make_shared<InodeManager>(inodeStorage_, trash_);
+        map_ = std::make_shared<FileType2InodeNumMap>();
```
wu-hanqing, line 94: ditto

Cyber-SiKu:
recheck

wu-hanqing (approved):

Issue #1529, opened 2022-05-31
Unbalanced usage of disk space by metaserver


ilixiaocui:
**Describe the bug (描述bug)**
In the 7*24-hour test, it was found that the disk space occupied by the metaserver on different nodes is quite different：
root@prit2-curve-chunk-node15:/data/curvefs# du -sh metaserver*
140G	metaserver0
116G	metaserver1
92G	metaserver2

root@prit2-curve-chunk-node16:/data/curvefs# du -sh metaserver*
70G	metaserver0
76G	metaserver1
75G	metaserver2

root@prit2-curve-chunk-node17:/data/curvefs# du -sh metaserver*
55G	metaserver0
105G	metaserver1
73G	metaserver2

metaserver对磁盘的占用很不均衡

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

cw123:
![image](https://user-images.githubusercontent.com/4277834/171174335-5aeb73ed-00c5-4ca2-af38-b0aaaec138d6.png)

cw123:
当前fs为partition选择copyset是随机选择的，当fs个数不多，或者fs间使用不均衡的时候，会出现使用率不均衡。
schedule的资源调度有个范围，当资源的使用率相差30%的时候，才会触发调度。当前这个集群，配置文件磁盘配置了2TB，内存配置了30GB。当磁盘使用量相差超过600GB或者内存使用量相差超过9GB的时候，会触发调度。
图中，在下午4点的时候，内存使用量相差超过了9GB，确实触发了调度。
所以当前这个这个环境的资源使用情况是正常的，符合设计预期的。
![image](https://user-images.githubusercontent.com/4277834/171174997-3ba5d7ba-8988-46c0-9d00-6855e9aa2d6c.png)

cw123:
<html>
<body>


  | 3 partition | 12 partition
-- | -- | --

[...]

PR #1559, opened 2022-06-09
curvefs:modify mds.conf

cw123:

Issue Number: #1529 

Problem Summary:

What's Changed:

1、原来的patition 格式默认是3，这个数字太小了。这个会导致一个fs的元数据只能分布在3个copyset中。现在把这个数字改成和集群的初始copyset一致，可以让fs的元数据初始分布更均匀。
2、调低了partition中的inode的个数，让fs尽量使用更多的partition去让元数据分布的更均衡。
3、调低了copyset中的partition的个数，避免一个copyset管理太多的元数据。
4、集群默认的copyset个数调整为3的倍数。

1. The original patition format defaults to 3, which is too small. This will cause the metadata of an fs to only be distributed in 3 copies. Now change this number to be consistent with the initial copy of the cluster, which can make the initial distribution of fs metadata more uniform.
2. Reduced the number of inodes in the partition, and let fs use as many partitions as possible to make the metadata distribution more balanced.
3. Reduced the number of partitions in the copy to avoid managing too much metadata in one copy.
4. The default number of copies in the cluster is adjusted to a multiple of 3.

How it Works:

### Check List

### DIFF
--- a/curvefs/conf/mds.conf
+++ b/curvefs/conf/mds.conf
@@ -44,16 +44,16 @@ leader.electionTimeoutMs=0
 #
 # time interval flush data to db
 mds.topology.TopologyUpdateToRepoSec=60
-# max partition number in copyset 2^8
-mds.topology.MaxPartitionNumberInCopyset=256
-# id number in each partition 2^24 [0, 2^24-1]
-mds.topology.IdNumberInPartition=16777216
+# max partition number in copyset 2^7
+mds.topology.MaxPartitionNumberInCopyset=128
+# id number in each partition 2^20 [0, 2^20-1]
+mds.topology.IdNumberInPartition=1048676
 # initial copyset number in cluster
-mds.topology.InitialCopysetNumber=10
+mds.topology.InitialCopysetNumber=12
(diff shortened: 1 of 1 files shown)


curvefs/conf/mds.conf
```diff
@@ -44,16 +44,16 @@ leader.electionTimeoutMs=0
...
+mds.topology.IdNumberInPartition=1048676
 # initial copyset number in cluster
-mds.topology.InitialCopysetNumber=10
+mds.topology.InitialCopysetNumber=12
 # min avaiable copyset num in cluster
-mds.topology.MinAvailableCopysetNum=10
+mds.topology.MinAvailableCopysetNum=12
 # default create partition number 3
-mds.topology.CreatePartitionNumber=3
+mds.topology.CreatePartitionNumber=12
```
ilixiaocui, line 56: Note the reason for the modification in the commit message

cw123:
recheck

SeanHai (approved):

ilixiaocui (approved):

PR #1570, opened 2022-06-13
metaserver: set apply queue and bthread worker thread name

wu-hanqing:
Support set thread name which will display by `pstree` and `top`, and also set apply queue and bthread worker name in metaserver.

Here is some illustration,

image

`top` command will show the process's resource usage, and `top -H -p ${pid}` will display every threads' resource usage.

If thread's name is not specified, we will see this

image

, every thread has the same name, so we can't distinguish them.

After this patch, bthread worker and apply queue's thread name is specified, so we can get more information from `top -H -p ${pid}`

image

### DIFF
--- a/curvefs/src/metaserver/copyset/apply_queue.h
+++ b/curvefs/src/metaserver/copyset/apply_queue.h
@@ -26,6 +26,7 @@
 #include <atomic>
 #include <cstdint>
 #include <memory>
+#include <string>
 #include <thread>
 #include <utility>
 #include <vector>
@@ -38,9 +39,12 @@ namespace curvefs {
 namespace metaserver {
 namespace copyset {
 
+class CopysetNode;
+
 struct ApplyQueueOption {
     uint32_t workerCount = 1;
     uint32_t queueDepth = 1;
+    CopysetNode* copysetNode = nullptr;
 };
 
 class CURVE_CACHELINE_ALIGNMENT ApplyQueue {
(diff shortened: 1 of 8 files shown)


wu-hanqing:
recheck

wu-hanqing:
recheck

wu-hanqing:
recheck

wu-hanqing:
recheck

SeanHai:
You can give the result after modify, it's better to do a comparison with before.

curvefs/src/metaserver/copyset/apply_queue.h
```diff
@@ -63,8 +67,11 @@ class CURVE_CACHELINE_ALIGNMENT ApplyQueue {
     void StartWorkers();
 
     struct TaskWorker {
-        explicit TaskWorker(size_t cap)
-            : running(false), worker(), tasks(cap) {}
+        TaskWorker(size_t cap, std::string workerName)
```
Cyber-SiKu, line 70: Why remove explicit

wu-hanqing: in most cases, `explicit` should only mark single argument constructor

wu-hanqing:
> You can give the result after modify in pr description, it's better to do a comparison with before.

I added some explanation

SeanHai (approved):

Cyber-SiKu (approved):

Issue #1631, opened 2022-06-24
copy not migrated after stopping metaserver


YunhuiChen:
**Describe the bug (描述bug)**

copy not migrated after stopping metaserver

**To Reproduce (复现方法)**
1、run vdbench and mdtest
2、stop metaserver1 copyset migrated
3、start metaserver1 
4、 stop metaserver2 copyset migrated

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

wu-hanqing:
<img width="2299" alt="image" src="https://user-images.githubusercontent.com/10327590/175485660-3b3f96e5-d7e9-4efd-8b74-833d348d1078.png">

it seems when we delete a copyset the corresponding rocksdb instance is not closed, so when create the same copyset again, open rocksdb failed.

PR #1634, opened 2022-06-24
metaserver: close metadata storage when stopping copyset

wu-hanqing:

Issue Number: #1631

Problem Summary: see #1631 

What's Changed:

How it Works:

### Check List


wu-hanqing:
recheck

curvefs/src/metaserver/storage/rocksdb_storage.h
```diff
@@ -183,10 +183,10 @@ class RocksDBStorage : public KVStorage, public StorageTransaction {
     void InitDbOptions();
 
  private:
-    bool inited_;
+    bool inited_ = false;
```
Wine93, line 186: why init these members in declare?

wu-hanqing: just give them a default value, because if you don't initialize them in constructor, the value is unspecified, and some assert will fail.

SeanHai (approved):

Wine93 (approved):
LGTM!

Issue #1627, opened 2022-06-23
Random write performance drops below 100 after a few minutes


YunhuiChen:
**Describe the bug (描述bug)**
Random write performance drops below 100 after a few minutes

**To Reproduce (复现方法)**
1、fio -directory=/home/nbs/failover/test3 -ioengine=psync -rw=randwrite -bs=4k -direct=1 -group_reporting=1 -fallocate=none -time_based=1 -name=1.txt -numjobs=1 -nrfiles=1 -size=100G
2、
![image](https://user-images.githubusercontent.com/67582932/175259546-558fe59c-f973-4267-bf14-8f856ba7b7d3.png)

nbs@prit2-curve-chunk-node12:~/failover$ fio -directory=/home/nbs/failover/test3 -ioengine=psync -rw=randwrite -bs=4k -direct=1 -group_reporting=1 -fallocate=none -time_based=1 -name=1.txt -numjobs=1 -nrfiles=1 -size=100G
fio: time_based requires a runtime/timeout setting
1.txt: (g=0): rw=randwrite, bs=4K-4K/4K-4K/4K-4K, ioengine=psync, iodepth=1
fio-2.16
Starting 1 process
1.txt: Laying out IO file(s) (1 file(s) / 102400MB)
Jobs: 1 (f=1): [w(1)] [4.1% done] [0KB/144KB/0KB /s] [0/36/0 iops] [eta 05h:46m:05s]    
Jobs: 1 (f=1): [w(1)] [4.4% done] [0KB/180KB/0KB /s] [0/45/0 iops] [eta 15h:37m:44s]

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

PR #1646, opened 2022-06-27
curvefs/client: fix problem of perf decline

wuhongsong:

Issue Number: #1627  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/s3/disk_cache_manager.cpp
+++ b/curvefs/src/client/s3/disk_cache_manager.cpp
@@ -67,7 +67,7 @@ DiskCacheManager::DiskCacheManager(std::shared_ptr<PosixWrapper> posixWrapper,
     // cannot limit the size,
     // because cache is been delete must after upload to s3
     cachedObjName_ = std::make_shared<
-      LRUCache<std::string, bool>>(0,
+      SglLRUCache<std::string>>(0,
         std::make_shared<CacheMetrics>("diskcache"));
 }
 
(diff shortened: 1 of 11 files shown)


curvefs/src/client/s3/disk_cache_manager.cpp
```diff
@@ -380,18 +380,19 @@ void DiskCacheManager::TrimCache() {
...
             // then read will fail when do cache read,
             // and then it cannot load the file from S3.
             // so read is fail.
-            if (ret == 0) {
+            while (ret == 0) {
                 VLOG(1) << "do not remove this disk file"
                         << ", file has not been uploaded to S3."
                         << ", file is: " << cacheKey;
-                continue;
+                bthread_usleep(1000);
```
xu-chaojie, line 394: why use bthread_usleep？ this function is running in bthread？

wuhongsong:
when file is deleted in cachewrite ,then we will call `cachedObjName_->Put(fileName, false)` ，and it will put the obj to the head of the list, so the  funciton GetLast will cost many  times， and it will  block flush because it will call `diskCacheManager_->AddCache(name)` .

curvefs/src/client/s3/disk_cache_manager.cpp
```diff
@@ -159,13 +159,13 @@ int DiskCacheManager::ClearReadCache(const std::list<std::string> &files) {
 
 void DiskCacheManager::AddCache(const std::string name,
   bool cacheWriteExist) {
-    cachedObjName_->Put(name, cacheWriteExist);
+    cachedObjName_->Put(name);
     VLOG(9) << "cache size is: " << cachedObjName_->Size();
 }
 
 bool DiskCacheManager::IsCached(const std::string name) {
     bool exist;
```
xu-chaojie, line 167: This variable is no longer used.

src/common/lru_cache.h
```diff
@@ -551,6 +559,26 @@ void SglLRUCache<K, KeyTraits>::Put(const K &key) {
...
+    ll_.push_back(key);
+    cache_[key] = --ll_.end();
+    size_++;
+    if (cacheMetrics_ != nullptr) {
+        cacheMetrics_->UpdateAddToCacheCount();
+        cacheMetrics_->UpdateAddToCacheBytes(KeyTraits::CountBytes(key));
+    }
+    return true;
+}
+
```
baijiaruo, line 581: I think the objname of the current diskcache no longer belongs to lru, it is recommended to implement this module by yourself

wuhongsong: ok， i have add todo in the code, will edit later.

baijiaruo (approved):

xu-chaojie (approved):

Issue #1678, opened 2022-07-01
Failed to create file system when mds switch


YunhuiChen:
**Describe the bug (描述bug)**

Failed to create file system when mds switch

**To Reproduce (复现方法)**
1、Create a filesystem in a loop with curvefs_tool
2、stop mds
3、[error]
send create fs request to mds: 10.166.24.58:6700 failed, errorcode= 112, error text: [E111]Fail to connect Socket{id=0 addr=10.166.24.58:6700} (0x0x7f4a6ad681c0): Connection refused [R1][E112]Not connected to 10.166.24.58:6700 yet, server_id=0 [R2][E112]Not connected to 10.166.24.58:6700 yet, server_id=0 [R3][E112]Not connected to 10.166.24.58:6700 yet, server_id=0
send create fs request to mds: 10.166.24.59:6700 failed, errorcode= 112, error text: [E111]Fail to connect Socket{id=2 addr=10.166.24.59:6700} (0x0x7f4a6ad68640): Connection refused [R1][E112]Not connected to 10.166.24.59:6700 yet, server_id=2 [R2][E112]Not connected to 10.166.24.59:6700 yet, server_id=2 [R3][E112]Not connected to 10.166.24.59:6700 yet, server_id=2
send create fs request to mds: 10.166.24.35:6700 failed, errorcode= 112, error text: [E111]Fail to connect Socket{id=4 addr=10.166.24.35:6700} (0x0x7f4a6ad68ac0): Connection refused [R1][E112]Not connected to 10.166.24.35:6700 yet, server_id=4 [R2][E112]Not connected to 10.166.24.35:6700 yet, server_id=4 [R3][E112]Not connected to 10.166.24.35:6700 yet, server_id=4

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

SeanHai:
The curvefs_tool use the buildin retry framework of brpc which is simple and can not deal with most situation.
You can see the docs: https://github.com/apache/incubator-brpc/blob/master/docs/cn/client.md#%E9%87%8D%E8%AF%95

In this case, it didn't trigger retry at all. So curvefs_tool rpc retry policy may need improve @Cyber-SiKu

Cyber-SiKu:
[...]

PR #1692, opened 2022-07-04
curvefs/tool: Adjust the logic of how to retry

Cyber-SiKu:
Send rpc requests repeatedly.

Issue Number: #1678  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/tools/curvefs_tool.h
+++ b/curvefs/src/tools/curvefs_tool.h
@@ -23,6 +23,7 @@
 #define CURVEFS_SRC_TOOLS_CURVEFS_TOOL_H_
 
 #include <brpc/channel.h>
+#include <bthread/bthread.h>
 #include <google/protobuf/message.h>
 #include <google/protobuf/service.h>
 #include <json/json.h>
@@ -48,6 +49,7 @@ DECLARE_string(confPath);
 DECLARE_uint32(rpcTimeoutMs);
 DECLARE_uint32(rpcRetryTimes);
 DECLARE_uint32(rpcStreamIdleTimeoutMs);
+DECLARE_uint32(rpcRetryIntervalUs);
 
 namespace curvefs {
 namespace tools {
@@ -210,7 +212,6 @@ class CurvefsToolRpc : public CurvefsTool {
         uint32_t failHostNumber = 0;
         bool ret = false;
         for (const std::string& host : hostsAddr_) {
-            SetController();
             brpc::ChannelOptions channelOpt;
             if (isStreaming_) {
                 // set stream rpc client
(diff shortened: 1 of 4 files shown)


Cyber-SiKu:
recheck

curvefs/src/tools/curvefs_tool.h
```diff
@@ -232,9 +232,16 @@ class CurvefsToolRpc : public CurvefsTool {
...
             // it will crash in there
-            service_stub_func_(controller_.get(), &requestQueue_.front(),
-                               response_.get());
-            if (controller_->Failed()) {
+            uint32_t i = 0;
+            for (i = 0; i <= FLAGS_rpcRetryTimes; ++i) {
+                controller_->Reset();
+                service_stub_func_(controller_.get(), &requestQueue_.front(),
+                                   response_.get());
+                if (!controller_->Failed()) {
```
wu-hanqing, line 240: it's better to check the error code of the controller, otherwise, you will do meaningless retry if the server is quitted or never started.

you can see this for some help

https://github.com/opencurve/curve/blob/ba87f7dba7dfccb4b78ab9bca54e1c261783c41a/src/client/mds_client.cpp#L125

curvefs/src/tools/curvefs_tool.h
```diff
@@ -232,9 +232,16 @@ class CurvefsToolRpc : public CurvefsTool {
...
+                    break;
+                }
```
SeanHai, line 242: After failed you should check the retcode and check whether need retry and need sleep for a while before retry.

It needn't retry In some cases, you can refer to :https://github.com/opencurve/curve/pull/1684/files

curvefs/src/tools/curvefs_tool.h
```diff
@@ -265,7 +272,6 @@ class CurvefsToolRpc : public CurvefsTool {
 
     virtual void SetController() {
         controller_->set_timeout_ms(FLAGS_rpcTimeoutMs);
-        controller_->set_max_retry(FLAGS_rpcRetryTimes);
```
wu-hanqing, line 268: You can keep it, and just set max retry times to 0. The framework will retry 3 times by default, but, I think it's also unnecessary.

Cyber-SiKu: > You can keep it, and just set max retry times to 0. The framework will retry 3 times by default, but, I think it's also unnecessary.

fix

curvefs/src/tools/curvefs_tool.h
Cyber-SiKu: > it's better to check the error code of the controller, otherwise, you will do meaningless retry if the server is quitted or never started.
> 
> you can see this for some help
> 
> https://github.com/opencurve/curve/blob/ba87f7dba7dfccb4b78ab9bca54e1c261783c41a/src/client/mds_client.cpp#L125

fix

curvefs/src/tools/curvefs_tool.h
Cyber-SiKu: > After failed you should check the retcode and check whether need retry and need sleep for a while before retry.
> 
> It needn't retry In some cases, you can refer to :https://github.com/opencurve/curve/pull/1684/files

fix
This is not multi-threaded, and sleep should not be needed

curvefs/src/tools/curvefs_tool.h
```diff
@@ -232,9 +232,24 @@ class CurvefsToolRpc : public CurvefsTool {
...
-            service_stub_func_(controller_.get(), &requestQueue_.front(),
-                               response_.get());
-            if (controller_->Failed()) {
+            uint32_t i = 0;
+            for (i = 0; i <= FLAGS_rpcRetryTimes; ++i) {
+                controller_->Reset();
+                service_stub_func_(controller_.get(), &requestQueue_.front(),
+                                   response_.get());
+                int32_t retCode = controller_->ErrorCode();
+                if (retCode == -EHOSTDOWN || retCode == -ECONNRESET ||
```
SeanHai, line 241: Here the code are positive number.

wu-hanqing (approved):

Cyber-SiKu:
recheck

SeanHai (approved):

Issue #1102, opened 2022-02-24
Error log optimization


ilixiaocui:
Following error log need optimize. Only the logs that need human operation and maintenance are defined as the Error level.

```
报警项:ccs.chunkserver日志监控,报警内容:cs=11, count=2, E 2022-02-23T15:06:15.071630+0800 1984056 io_tracker.cpp:296] file [\/nova\/a8cf72cc-9f1a-4296-8d62-2550a62212ab]  IO Error  OpType   Read  offset   8591355904  length   4096  E 2022-02-23T15:06:15.071630+0800 1984056 io_tracker.cpp:296] file [\/nova\/a8cf72cc-9f1a-4296-8d62-2550a62212ab]  IO Error  OpType   Read  offset   8591355904  length   4096  W 2022-02-23T15:06:15.071630+0800 1983065 service_helper.cpp:290] GetLeader failed from 10.194.49.18:8219  logicpool id   2  copyset id   884  proxy id   75  error   [E125]Operation canceled @10.194.49.18:8219  W 2022-02-23T15:06:15.286691+0800 1984052 chunk_closure.cpp:371] Read not exists  logicpool id   2  copyset id   884  chunk id   21633561  offset   1421312  length   4096  sn   1  source info   empty  status 7  retried times   1  IO id   164963  request id   164963  remote side   10.194.49.21:8218  W 2022-02-23T15:06:15.457401+0800 1984051 chunk_closure.cpp:371] Read not exists  logicpool id   2  copyset id   884  chu
````

```
[...]

PR #1694, opened 2022-07-04
curvebs client,common : Error log optimization

xu-chaojie:

Issue Number: #1102  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/src/client/io_tracker.cpp
+++ b/src/client/io_tracker.cpp
@@ -505,7 +505,7 @@ void IOTracker::Done() {
     } else {
         MetricHelper::IncremUserEPSCount(fileMetric_, type_);
         if (type_ == OpType::READ || type_ == OpType::WRITE) {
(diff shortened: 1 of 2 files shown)


src/client/io_tracker.cpp
```diff
@@ -505,7 +505,7 @@ void IOTracker::Done() {
     } else {
         MetricHelper::IncremUserEPSCount(fileMetric_, type_);
         if (type_ == OpType::READ || type_ == OpType::WRITE) {
-            LOG(ERROR) << "file [" << fileMetric_->filename << "]"
+            LOG(WARNING) << "file [" << fileMetric_->filename << "]"
```
wu-hanqing, line 508: I think is not a good idea to lower this log level to WARNING, because when this error is printed, it means a real error happens and this error will return to the upper-layer application.

ilixiaocui:
cc @xu-chaojie 
Please resubmit the pr to the new master branch if needed.

Issue #1699, opened 2022-07-06
curvefs/client: the data in the read disk cache is more larger than the actual data shard


ilixiaocui:
**Describe the bug (描述bug)**
Write a few hundred bytes of data，S3Info is 955 byte:
<img width="567" alt="image" src="https://user-images.githubusercontent.com/15689619/177456028-a6a2ed16-55b0-4873-9d55-ef7037cae22c.png">

Then use a new mount point to read this data，we get 4M data:
<img width="550" alt="image" src="https://user-images.githubusercontent.com/15689619/177456164-d259147e-7c4c-4ffd-9455-f410c7603c25.png"> 

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS: 
Compiler:
branch: master、all releases
commit id:

**Additional context/screenshots (更多上下文/截图)**

baijiaruo:
![image](https://user-images.githubusercontent.com/7495307/177456388-3ef21e49-e5a6-40be-a1b9-d49e63e3019a.png)
When pre-reading, I did not look at the actual size of the file, but directly pre-read a block size of 4M.

PR #1703, opened 2022-07-06
curvefs/client: fix data in read disk cache is more larger than the a…

baijiaruo:
…ctual data

Signed-off-by: baijiaruo <baijiaruo@126.com>

Issue Number: #1699  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/s3/client_s3_cache_manager.h
+++ b/curvefs/src/client/s3/client_s3_cache_manager.h
@@ -315,7 +315,8 @@ class FileCacheManager {
     int ReadFromS3(const std::vector<S3ReadRequest> &requests,
                             std::vector<S3ReadResponse> *responses,
                             uint64_t fileLen);
-    void PrefetchS3Objs(std::vector<std::string> prefetchObjs);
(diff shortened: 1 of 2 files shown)


wuhongsong (approved):

baijiaruo:
image
image
Write amplification problem solved

curvefs/src/client/s3/client_s3_cache_manager.h
```diff
@@ -315,7 +315,8 @@ class FileCacheManager {
     int ReadFromS3(const std::vector<S3ReadRequest> &requests,
                             std::vector<S3ReadResponse> *responses,
                             uint64_t fileLen);
-    void PrefetchS3Objs(std::vector<std::string> prefetchObjs);
+    void
+    PrefetchS3Objs(std::vector<std::pair<std::string, uint64_t>> prefetchObjs);
```
ilixiaocui, line 319: const std::vector<std::pair<std::string, uint64_t>> &prefetchObjs

ilixiaocui (approved):

Issue #1708, opened 2022-07-07
curvefs/client: need metric of diskcache size usage


ilixiaocui:
**Describe the bug (描述bug)**
Now we don't know the cache disk usage from the metric

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

PR #1709, opened 2022-07-07
curvefs/client: add metric for diskcache usedbytes

wuhongsong:

Issue Number: #1708  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/metric/client_metric.h
+++ b/curvefs/src/client/metric/client_metric.h
@@ -262,10 +262,13 @@ struct DiskCacheMetric {
 
     std::string fsName;
     InterfaceMetric writeS3;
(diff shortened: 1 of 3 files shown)


curvefs/src/client/metric/client_metric.h
```diff
@@ -262,10 +262,13 @@ struct DiskCacheMetric {
 
     std::string fsName;
     InterfaceMetric writeS3;
+    bvar::Status<uint64_t> diskUsedBytes;
```
ilixiaocui, line 265: Use bvar::Adder directly?

wuhongsong: just using existing variables

ilixiaocui (approved):

Issue #1672, opened 2022-06-30
A metaserver machine failure needs to wait for the migration before adding a new machine


YunhuiChen:
**Describe the bug (描述bug)**
A metaserver machine failure needs to wait for the migration before adding a new machine

**To Reproduce (复现方法)**
1、stop one metaserver
2、add new metaserver
![image](https://user-images.githubusercontent.com/67582932/176607870-bd92a8be-61ca-4eb9-93ff-ab1698894722.png)

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

PR #1719, opened 2022-07-08
Cherry pick #1693 to v2.3

cw123:

Issue Number: #1672
Problem Summary: If a metaserver disk is broken, the new disk can not add to cluster to replace the old one. Now I change the logic. If the old metaserver is in offline status, the new metaserver can add to cluster use the old ip and old port.

What's Changed:

How it Works:

### Check List


cw123:
merge to wrong branch

Issue #1428, opened 2022-05-12
curvefs+bs:The fs side needs to judge the available capacity of the bs volume. When the written data is equal to or greater than the volume capacity, an io error will be returned


YunhuiChen:
**Describe the bug (描述bug)**
The fs side needs to judge the available capacity of the bs volume. If it exceeds the available capacity, an error will be returned.

**To Reproduce (复现方法)**

![image](https://user-images.githubusercontent.com/67582932/168036147-73706ab4-a8b7-43c5-baca-4467ce348b77.png)
![image](https://user-images.githubusercontent.com/67582932/168036195-5df958cd-49c4-412d-90f9-99e0e2e62aac.png)

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

PR #1732, opened 2022-07-11
curve-fuse: return NOSPACE when volume space used up

wu-hanqing:

Issue Number: #1428

Problem Summary:

What's Changed:

How it Works:

### Check List


wu-hanqing:
recheck

wu-hanqing:
recheck

curvefs/src/volume/common.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <ostream>
+#include <string>
+
+#include "curvefs/src/volume/utils.h"
+
+namespace curvefs {
+namespace volume {
+
+std::ostream& operator<<(std::ostream& os, const AllocateHint& hint) {
```
SeanHai, line 33: Implement this in AllocateHint maybe more clear.

wu-hanqing: it's not a core function, so I think it's ok to put it here. And, put so many logic in header file will slow down compilation and linking.

wu-hanqing:
recheck

wu-hanqing:
recheck

SeanHai (approved):

Issue #1048, opened 2022-01-21
truncate may result in data corruption


h0hmj:
**Describe the bug (描述bug)**
situation:

Assume:
each s3chunkinfo's end == next s3chunkinfo's begin, no overlap
each s3chunkinfo size = 1 for id 0~7
s3compact threshold = 5

State 0: len 8
chunkIndex 0
s3info chunkid [0,1,2,3]
chunkIndex 1
s3info chunkid [4,5,6,7]

Action 1:
truncate -s 1
will change inode len only

State 1: len 1
chunkIndex 0
s3info chunkid [0,1,2,3]
chunkIndex 1
s3info chunkid [4,5,6,7]

Action 2:
truncate -s 8
will add new zero s3chunkinfo to list with same chunkid  <- root cause
chunkid is not unique globally

State 2: len 8
chunkIndex 0
s3info chunkid [0,1,2,3,8]
chunkIndex 1
s3info chunkid [4,5,6,7,8]

Action 3:
meet s3compact threshold
merge all s3info to one with the largest chunkid, and write s3

State 3
chunkIndex 0
s3info chunkid [8]
chunkIndex 1
s3info chunkid [8]

current s3objname rule is
fsid + inodeid + chunkid + blockindex_in_a_chunk + compaction

metadata in State 3 will get same s3objname
result in data corruption

**Expected behavior (期望行为)**

chunkid should be unique, only used by one s3chunkinfo, rather than different s3chunkinfo share the same chunkid

ps: block index in s3obj naming rule is not meaningful. we cannot get data location from name. Using offset of the file will be better, we can calculate chunkindex and blockindex. offset = chunkindex*chunksize + blockindex*blocksize

h0hmj:
ping @baijiaruo @ilixiaocui @Cyber-SiKu

baijiaruo:
This problem still exists, and mds needs to provide an interface to obtain multiple chunkids

PR #1739, opened 2022-07-13
curvefs: fix truncate file size becomes larger, chunkid duplicate que…

baijiaruo:
…stion

Signed-off-by: baijiaruo <baijiaruo@126.com>

Issue Number: #1048  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/proto/mds.proto
+++ b/curvefs/proto/mds.proto
@@ -155,12 +155,13 @@ message DeleteFsResponse {
 
 message AllocateS3ChunkRequest {
     required uint32 fsId = 1;
+    optional uint32 chunkIdNum = 2;
 }
 
 message AllocateS3ChunkResponse {
-    // status code, default value is FSStatusCode::UNKNOWN_ERROR
+   // status code, default value is FSStatusCode::UNKNOWN_ERROR
    required FSStatusCode statusCode = 1;
-   required uint64 chunkId = 2;
+   required uint64 beginChunkId = 2;
 }
 
 message ListClusterFsInfoRequest {
--- a/curvefs/src/client/rpcclient/base_client.cpp
+++ b/curvefs/src/client/rpcclient/base_client.cpp
@@ -130,12 +130,13 @@ void MDSBaseClient::ListPartition(uint32_t fsID,
     stub.ListPartition(cntl, &request, response, nullptr);
 }
 
-void MDSBaseClient::AllocS3ChunkId(uint32_t fsId,
+void MDSBaseClient::AllocS3ChunkId(uint32_t fsId, uint32_t idNum,
                                    AllocateS3ChunkResponse* response,
                                    brpc::Controller* cntl,
                                    brpc::Channel* channel) {
     AllocateS3ChunkRequest request;
     request.set_fsid(fsId);
+    request.set_chunkidnum(idNum);
 
     curvefs::mds::MdsService_Stub stub(channel);
     stub.AllocateS3Chunk(cntl, &request, response, nullptr);
--- a/curvefs/src/client/s3/client_s3_adaptor.cpp
+++ b/curvefs/src/client/s3/client_s3_adaptor.cpp
@@ -197,20 +197,23 @@ CURVEFS_ERROR S3ClientAdaptorImpl::Truncate(Inode *inode, uint64_t size) {
         uint64_t index = offset / chunkSize_;
         uint64_t chunkPos = offset % chunkSize_;
         uint64_t n = 0;
-        uint64_t chunkId;
+        uint64_t beginChunkId;
         FSStatusCode ret;
         uint64_t fsId = inode->fsid();
-        ret = AllocS3ChunkId(fsId, &chunkId);
+        uint32_t chunkIdNum = len / chunkSize_ + 1;
+        ret = AllocS3ChunkId(fsId, chunkIdNum, &beginChunkId);
         if (ret != FSStatusCode::OK) {
             LOG(ERROR) << "Truncate alloc s3 chunkid fail. ret:" << ret;
             return CURVEFS_ERROR::INTERNAL;
         }
+        uint64_t chunkId = beginChunkId;
         while (len > 0) {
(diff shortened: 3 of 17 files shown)


curvefs/proto/mds.proto
```diff
@@ -155,12 +155,13 @@ message DeleteFsResponse {
 
 message AllocateS3ChunkRequest {
     required uint32 fsId = 1;
+    required uint64 chunkIdNum = 2;
```
wu-hanqing, line 158: The required field will lead to incompatible with previous version. A feasible way is to make this field optional, and if client-side sets this field then the server-side allocates more chunk ids, and if not, server-side only allocates one chunk id.
 
And `uint32` is already enough for this field.

curvefs/src/client/rpcclient/base_client.cpp
```diff
@@ -130,12 +130,13 @@ void MDSBaseClient::ListPartition(uint32_t fsID,
     stub.ListPartition(cntl, &request, response, nullptr);
 }
 
-void MDSBaseClient::AllocS3ChunkId(uint32_t fsId,
+void MDSBaseClient::AllocS3ChunkId(uint32_t fsId, uint64_t idNum,
```
wu-hanqing, line 133: ditto, `uint32_t`

curvefs/test/mds/chunkid_allocator_test.cpp
```diff
@@ -104,7 +104,7 @@ TEST_F(ChunkIdAllocatorTest, test_get_chunkIds_1001) {
     set<uint64_t> chunkids;  // record get chunkids
     for (int i = 0; i < times; ++i) {
         uint64_t tmp = 0;
-        chunkIdAllocator_->GenChunkId(&tmp);
+        chunkIdAllocator_->GenChunkId(1, &tmp);
```
wu-hanqing, line 107: it's better to add test case for allocating more chunks at once.

curvefs/src/client/s3/client_s3_adaptor.cpp
```diff
@@ -197,20 +197,23 @@ CURVEFS_ERROR S3ClientAdaptorImpl::Truncate(Inode *inode, uint64_t size) {
         uint64_t index = offset / chunkSize_;
         uint64_t chunkPos = offset % chunkSize_;
         uint64_t n = 0;
-        uint64_t chunkId;
+        uint64_t beginChunkId;
         FSStatusCode ret;
         uint64_t fsId = inode->fsid();
-        ret = AllocS3ChunkId(fsId, &chunkId);
+        uint64_t chunkIdNum = len / chunkSize_ + 1;
+        ret = AllocS3ChunkId(fsId, chunkIdNum, &beginChunkId);
```
wu-hanqing, line 204: Add a test case. A good practice is adding a test case to reproduce issue #1048 firstly, and after your modification, the test should be passed.

wu-hanqing (changes requested):

curvefs/src/mds/chunkid_allocator.cpp
```diff
@@ -26,20 +26,24 @@
...
 namespace mds {
 
-int ChunkIdAllocatorImpl::GenChunkId(uint64_t *chunkId) {
+int ChunkIdAllocatorImpl::GenChunkId(uint64_t idNum, uint64_t *chunkId) {
     int ret = 0;
     // will unloack at destructor
     ::curve::common::WriteLockGuard guard(nextIdRWlock_);
 
-    if (nextId_ > lastId_ || CHUNKIDINITIALIZE == nextId_) {
+    if (nextId_ + idNum > lastId_ || CHUNKIDINITIALIZE == nextId_) {
```
Cyber-SiKu, line 34: nextId_ + idNum - 1 > lastId_ ? 
latestid_ indicates the last assignable id, and netxid indicates the next assignable id. So the current number of assignable ids is latestid - nextid +1.

curvefs/src/mds/chunkid_allocator.cpp
```diff
@@ -26,20 +26,24 @@
...
         // the chunkIds is exhausted || first to get chunkId
-        ret = AllocateBundleIds(bundleSize_);
+        uint64_t allocSize = nextId_ + idNum - lastId_;
```
Cyber-SiKu, line 36: allocSize = ((nextId_ + idNum - lastId_)/bundleSize_ +1)*bundleSize_ ?

Otherwise, the next application will still cause resource exhaustion to trigger allocation

curvefs/test/mds/chunkid_allocator_test.cpp
Cyber-SiKu: Agree, and preferably some limit value or random value, and judge whether the returned value is correct each time.

curvefs/proto/mds.proto
baijiaruo: fixed

curvefs/test/mds/chunkid_allocator_test.cpp
baijiaruo: fixed

curvefs/src/mds/mds_service.cpp
```diff
@@ -275,8 +275,8 @@ void MdsServiceImpl::AllocateS3Chunk(
     brpc::ClosureGuard guard(done);
     VLOG(9) << "start to allocate chunkId.";
 
-    uint64_t chunkId = 0;
-    int stat = chunkIdAllocator_->GenChunkId(&chunkId);
+    uint64_t  chunkId = 0;
+    int stat = chunkIdAllocator_->GenChunkId(request->chunkidnum(), &chunkId);
```
wu-hanqing, line 279: should you test whether the request has chunkidnum, otherwise, it will return 0, or just give the optional field a default value in the message definition

wu-hanqing (changes requested):

Cyber-SiKu (approved):

baijiaruo:
recheck

baijiaruo:
recheck

PR #1757, opened 2022-07-20
curvefs/client: eliminate s3ChunkInfo when too large and get merged one from metaserver

SeanHai:

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


SeanHai:
recheck

SeanHai:
recheck

curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -118,6 +118,14 @@ CURVEFS_ERROR InodeCacheManagerImpl::GetInode(uint64_t inodeId,
                     << inodeId << " opencount is 0";
             iCache_->Remove(inodeId);
         } else {
+            curve::common::UniqueLock lgGuard = out->GetUniqueLock();
```
ilixiaocui, line 121: Refreshing directly on the path of getinode will reduce performance. You can consider getinode triggered and refresh asynchronously.

curvefs/src/client/inode_wrapper.h
```diff
@@ -363,6 +377,21 @@ class InodeWrapper : public std::enable_shared_from_this<InodeWrapper> {
...
 
+    bool NeedRefreshData() {
+        uint64_t size = GetS3ChunkSize();
+        if (size >= maxDataSize_ &&
+            ::curve::common::TimeUtility::GetTimeofDaySec()
+            - lastRefreshTime_ > refreshDataInterval_) {
+            VLOG(6) << "EliminateLargeS3ChunkInfo size = " << size
+                    << ", max = " << maxDataSize_
+                    << ", inodeId = " << inode_.inodeid();
+            inode_.clear_s3chunkinfomap();
```
ilixiaocui, line 388: Why is the infomap of the inode emptied in the judgment function?RefreshS3ChunkInfo will handle old chunkinfomap.

curvefs/src/client/inode_cache_manager.cpp
SeanHai: If refresh s3ChunkInfo asynchronously, you need combine the s3ChunkInfo with cached one which may increased.
The combine is not easy and this process will also block io if write at same time.

curvefs/src/client/inode_wrapper.h
SeanHai: fixed

curvefs/src/client/inode_wrapper.h
```diff
@@ -363,6 +377,19 @@ class InodeWrapper : public std::enable_shared_from_this<InodeWrapper> {
 
     CURVEFS_ERROR RefreshVolumeExtent();
 
+    bool NeedRefreshData() {
+        uint64_t size = GetS3ChunkSize();
+        if (size >= maxDataSize_ &&
+            ::curve::common::TimeUtility::GetTimeofDaySec()
+            - lastRefreshTime_ > refreshDataInterval_) {
```
wu-hanqing, line 384: reverse these two conditions, `GetS3ChunkSize` is time-consuming.

curvefs/src/client/inode_wrapper.h
```diff
@@ -80,28 +82,40 @@ class InodeWrapper : public std::enable_shared_from_this<InodeWrapper> {
...
           isNlinkValid_(true),
           metaClient_(metaClient),
           s3ChunkInfoMetric_(s3ChunkInfoMetric),
           openCount_(0),
-          dirty_(false) {
+          dirty_(false),
+          baseMaxDataSize_(maxDataSize),
+          maxDataSize_(maxDataSize),
+          refreshDataInterval_(refreshDataInterval),
+          lastRefreshTime_(0) {
```
wu-hanqing, line 98: it's better to set it to current timestamp

curvefs/src/client/inode_wrapper.h
```diff
@@ -80,28 +82,40 @@ class InodeWrapper : public std::enable_shared_from_this<InodeWrapper> {
...
```
wu-hanqing, line 118: ditto

curvefs/src/client/inode_wrapper.h
```diff
@@ -363,6 +377,19 @@ class InodeWrapper : public std::enable_shared_from_this<InodeWrapper> {
 
     CURVEFS_ERROR RefreshVolumeExtent();
 
+    bool NeedRefreshData() {
+        uint64_t size = GetS3ChunkSize();
```
wu-hanqing, line 381: Can we cache the number of s3info of current inode? so, we don't have to calculate the size every time.

For example, when we fetch chunkinfo map from metaserver, we can calculate an initial value, and in `AppendS3ChunkInfo` adds 1 to the value. So, `GetS3ChunkSize` can directly returns the cached value.

This logic can also apply to `GetS3ChunkInfoAddSize`

curvefs/src/client/inode_wrapper.h
SeanHai: fixed

curvefs/src/client/inode_wrapper.h
SeanHai: fixed

curvefs/src/client/inode_wrapper.h
SeanHai: fixed

curvefs/src/client/inode_wrapper.h
SeanHai: fixed, recorded in inodeWrapper and update when append and refresh

wu-hanqing (approved):

ilixiaocui (approved):

Issue #1582, opened 2022-06-16
The openCount may miss when eliminated from icache and overflow when release after reload


SeanHai:
**Describe the bug (描述bug)**
releated code: https://github.com/opencurve/curve/blob/1b69419bbedb76c73336bd2d2528627db8a9cc03/curvefs/src/client/inode_wrapper.cpp#L330

<img width="521" alt="截屏2022-06-16 14 43 01" src="https://user-images.githubusercontent.com/13496900/174008424-9eee6806-c50d-45ed-9b75-b1a0bc1634f1.png">

// one
The 'openCount_' will not be persistented when inode eliminated form icache, the  'openCount_' will overflow and 'InodeOpenStatusChange::CLOSE' status will not update to metaserver anymore at below case:

A file be opened and eliminated before release, the inode will get from metaserver when release, the 'openCount_' will lost and inited to 0. So the 'openCount_' will overflow when release(0--).

Test on the curvefs:
1. set icache limit to 2 in client.yaml
2. create file1 file2 file3
3. open file1 file2 file3
4. close file1, it occurs.

<img width="776" alt="截屏2022-06-16 11 00 11" src="https://user-images.githubusercontent.com/13496900/174010854-ca1d49bc-c845-4baf-8433-d630fa8bbe65.png">

// two
A file be opened and eliminated before release, another process open it again and will get inode from metaserver and opencount is 0 now, so it will update 'InodeOpenStatusChange::OPEN' status to metaserver. The opencount is messed in metaserver.

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

ilixiaocui:
opencount is removed by pr: https://github.com/opencurve/curve/pull/1771

Issue #1463, opened 2022-05-18
Multiple Mount If the CTO function is inconsistent, the mount point data of the CTO function will be overwritten


YunhuiChen:
**Describe the bug (描述bug)**
Multiple Mount If the CTO function is inconsistent, the mount point data of the CTO function will be overwritten

1、curvebs2 close cto
volumeSize: 107374182400
volumeName: /test3
volumeUser: test
#VolumePassword: 123
container_image: harbor.cloud.netease.com/curve/curvefs:release2.2
log_dir: /data/logs/curvefs
client.loglevel: 8
core_dir: /corefile
container_pid: host
fuseClient.cto: false
2、curvebs3 open cto
![image](https://user-images.githubusercontent.com/67582932/168973621-7f84b9ea-4392-44b0-8baf-f71dda7b8c1c.png)

**To Reproduce (复现方法)**
![image](https://user-images.githubusercontent.com/67582932/168973441-c8729bc2-2dc9-4073-a397-9d800732aad6.png)

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

ilixiaocui:
All mount points on the same file system are either enableCto, or they do not guarantee consistency. fix by pr: https://github.com/opencurve/curve/pull/1771

YunhuiChen:
![image](https://user-images.githubusercontent.com/67582932/185567195-64727bb4-0c49-4976-a7bd-a7f1e66181ff.png)

Issue #1213, opened 2022-03-25
Description Echo data is corrupted after being transferred to multiple mounted files


YunhuiChen:
**Describe the bug (描述bug)**

**To Reproduce (复现方法)**
1、mount test1 and test10
2、vdbench test test1
3、run test.sh
#/bin/bash
for i in `seq 1 100`;do 
	echo "test $i" >> /home/nbs/failover/test1/cyh
	cat /home/nbs/failover/test1/cyh
	cat /home/nbs/failover/test10/cyh
	sleep 0.5
	echo "test $i" >> /home/nbs/failover/test10/cyh
	cat /home/nbs/failover/test1/cyh
	cat /home/nbs/failover/test10/cyh
	sleep 0.5
done
4、cyh is :
![image](https://user-images.githubusercontent.com/67582932/160079316-23eacc6a-9c73-4aaa-a989-fed37f92422d.png)

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
curve-mds:
curve-chunkserver:
curve-snapshotcloneserver:
curve-sdk:
nebd:
curve-nbd:

**Additional context/screenshots (更多上下文/截图)**

ilixiaocui:
FuseOpRelease is asynchronously

<img width="808" alt="image" src="https://user-images.githubusercontent.com/15689619/160326345-ff421c28-4ae1-44d5-9ac4-e0b37d6010c2.png">

Solution: same as https://github.com/opencurve/curve/issues/1197
Let FuseOpFlush handle the logic in FuseOpRelease.

YunhuiChen:
This is normal in the Curvebs + Curvefs scenario

![image](https://user-images.githubusercontent.com/67582932/168982278-13f1fc7f-008b-40ad-b214-89febe3ae0d9.png)

ilixiaocui:
mountpoint1: 
<img width="1228" alt="image" src="https://user-images.githubusercontent.com/15689619/180411051-e91d0bed-daf2-495c-9a9b-ddc62c66ee43.png">
mountpoint2:
<img width="1227" alt="image" src="https://user-images.githubusercontent.com/15689619/180411084-71965faa-6b92-44aa-ba00-dac2e80a64be.png">

ilixiaocui:
Interface calls corresponding to open and close：
```
I 2022-07-28T10:14:00.800768+0800 1435874 fuse_client.cpp:776] FuseOpGetAttr ino = 1
[...]

PR #1771, opened 2022-07-29
curvefs/client: fix cto bugs

ilixiaocui:
Signed-off-by: ilixiaocui <ilixiaocui@163.com>

related to issue: 1213/1582/1463
1. All mount points on the same file system are either enableCto, or they do not guarantee consistency.
2. If enableCto, attr and dentry cache will auto set invalid.
3. Consistency of multi-mount nodes under delete operation is not guaranteed. If one mount point deletes a file,
the behavior of the other mount point for all operations on that file is unknown.

Issue Number: #1213 #1582 #1463 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/inode_cache_manager.cpp
+++ b/curvefs/src/client/inode_cache_manager.cpp
@@ -107,107 +107,120 @@ class TrimICacheAsyncDone : public MetaServerClientDone {
     std::shared_ptr<InodeCacheManagerImpl> inodeCacheManager_;
 };
 
-CURVEFS_ERROR InodeCacheManagerImpl::GetInode(uint64_t inodeId,
-    std::shared_ptr<InodeWrapper> &out) {
+#define GET_INODE_REMOTE(FSID, INODEID, OUT, STREAMING)                        \
+    MetaStatusCode ret = metaClient_->GetInode(FSID, INODEID, OUT, STREAMING); \
+    if (ret != MetaStatusCode::OK) {                                           \
+        LOG_IF(ERROR, ret != MetaStatusCode::NOT_FOUND)                        \
+            << "metaClient_ GetInode failed, MetaStatusCode = " << ret         \
+            << ", MetaStatusCode_Name = " << MetaStatusCode_Name(ret)          \
+            << ", inodeid = " << INODEID;                                      \
+        return MetaStatusCodeToCurvefsErrCode(ret);                            \
+    }
+
+#define PUT_INODE_CACHE(INODEID, INODEWRAPPER)                                 \
+    std::shared_ptr<InodeWrapper> eliminatedOne;                               \
+    bool eliminated = iCache_->Put(INODEID, INODEWRAPPER, &eliminatedOne);     \
+    if (eliminated) {                                                          \
+        VLOG(3) << "GetInode eliminate one inode, ino: "                       \
+                << eliminatedOne->GetInodeId()                                 \
+                << ", iCache does not evict inodes via put interface ";        \
+        assert(0);                                                             \
+    }
+
+#define REFRESH_DATA_REMOTE(OUT, STREAMING)                                    \
+    CURVEFS_ERROR rc = RefreshData(OUT, STREAMING);                            \
+    if (rc != CURVEFS_ERROR::OK) {                                             \
+        return rc;                                                             \
+    }
+
+CURVEFS_ERROR
+InodeCacheManagerImpl::GetInode(uint64_t inodeId,
+                                std::shared_ptr<InodeWrapper> &out) {
     NameLockGuard lock(nameLock_, std::to_string(inodeId));
+    // get inode from cache
     bool ok = iCache_->Get(inodeId, &out);
     if (ok) {
-        // if enableCto, we need and is unopen, we need reload from metaserver
-        if (curvefs::client::common::FLAGS_enableCto && !out->IsOpen()) {
-            VLOG(6) << "InodeCacheManagerImpl, GetInode: enableCto and inode: "
-                    << inodeId << " opencount is 0";
-            iCache_->Remove(inodeId);
-        } else {
-            curve::common::UniqueLock lgGuard = out->GetUniqueLock();
-            if (out->NeedRefreshData()) {
-                CURVEFS_ERROR rc = out->RefreshS3ChunkInfo();
-                if (rc != CURVEFS_ERROR::OK) {
-                    LOG(ERROR) << "RefreshS3ChunkInfo failed, retCode = " << rc;
-                    return rc;
-                }
-            }
-            return CURVEFS_ERROR::OK;
-        }
+        curve::common::UniqueLock lgGuard = out->GetUniqueLock();
+        REFRESH_DATA_REMOTE(out, out->NeedRefreshData());
+        return CURVEFS_ERROR::OK;
     }
 
+    // get inode from metaserver
     Inode inode;
-    bool streaming;
-
-    MetaStatusCode ret2 = metaClient_->GetInode(
-        fsId_, inodeId, &inode, &streaming);
-
-    if (ret2 != MetaStatusCode::OK) {
-        LOG_IF(ERROR, ret2 != MetaStatusCode::NOT_FOUND)
-            << "metaClient_ GetInode failed, MetaStatusCode = " << ret2
-            << ", MetaStatusCode_Name = " << MetaStatusCode_Name(ret2)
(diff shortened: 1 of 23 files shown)


ilixiaocui:
recheck

ilixiaocui:
recheck

ilixiaocui:
recheck

ilixiaocui:
recheck

ilixiaocui:
recheck

curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -107,107 +107,120 @@ class TrimICacheAsyncDone : public MetaServerClientDone {
...
-            if (out->NeedRefreshData()) {
-                CURVEFS_ERROR rc = out->RefreshS3ChunkInfo();
-                if (rc != CURVEFS_ERROR::OK) {
-                    LOG(ERROR) << "RefreshS3ChunkInfo failed, retCode = " << rc;
-                    return rc;
-                }
-            }
-            return CURVEFS_ERROR::OK;
-        }
+        curve::common::UniqueLock lgGuard = out->GetUniqueLock();
```
SeanHai, line 143: Need refreshInode here If cto enabled?

curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -107,107 +107,120 @@ class TrimICacheAsyncDone : public MetaServerClientDone {
...
     bool ok = iCache_->Get(inodeId, &inodeWrapper);
     if (ok) {
-        if (curvefs::client::common::FLAGS_enableCto &&
-            !inodeWrapper->IsOpen()) {
-            iCache_->Remove(inodeId);
-        } else {
-            inodeWrapper->GetInodeAttrLocked(out);
-            return CURVEFS_ERROR::OK;
-        }
+        inodeWrapper->GetInodeAttrLocked(out);
```
SeanHai, line 206: Need refreshInodeAttr here If cto enabled?

curvefs/src/mds/fs_info_wrapper.cpp
```diff
@@ -75,15 +75,32 @@ FsInfoWrapper::FsInfoWrapper(const ::curvefs::mds::CreateFsRequest* request,
...
                         fsInfo_.mountpoints().end(),
-                        [mp](const Mountpoint& mountPoint) {
+                        [mp](const Mountpoint &mountPoint) {
                             return mp.path() == mountPoint.path() &&
                                    mp.hostname() == mountPoint.hostname();
                         }) != fsInfo_.mountpoints().end();
 }
 
+bool FsInfoWrapper::IsMountPointConflict(const Mountpoint &mp) const {
+    bool cto = (fsInfo_.mountpoints_size() ? false : mp.cto());
```
cw123, line 88: what does this cto mean?
What is the relationship between mountpoints_size and cto

curvefs/src/metaserver/trash.cpp
```diff
@@ -123,11 +123,7 @@ bool TrashImpl::NeedDelete(const TrashItem &item) {
                      << ", ret = " << MetaStatusCode_Name(ret);
         return false;
     }
-    if (inode.has_openmpcount() && inode.openmpcount() > 0) {
-            return false;
-    } else {
-        return ((now - item.dtime) >= options_.expiredAfterSec);
-    }
+    return ((now - item.dtime) >= options_.expiredAfterSec);
```
cw123, line 126: if delete opencount, when deleting an open file, what should trash do here?

curvefs/src/mds/fs_info_wrapper.cpp
```diff
@@ -75,15 +75,32 @@ FsInfoWrapper::FsInfoWrapper(const ::curvefs::mds::CreateFsRequest* request,
...
+                     [&](const Mountpoint &mountPoint) {
+                         if (mountPoint.has_cto() && mountPoint.cto()) {
+                             cto = true;
+                         }
+
+                         return mp.path() == mountPoint.path() &&
+                                mp.hostname() == mountPoint.hostname();
+                     }) != fsInfo_.mountpoints().end();
+
+    return exist || (!cto && mp.cto()) || (cto && !mp.cto());
```
SeanHai, line 101: Here is not more clear and can add some code comments.

curvefs/src/mds/fs_info_wrapper.cpp
ilixiaocui: On a mount point, there are only two ways to mount multiple clients. One is that all mount points are all enableCto and the other is that all mount points are not enableCto.

The cto variable here is to record, whether the current mount point of the file system has cto enabled. For the initialization of this variable, if the file system already has a mount point, set it to false, and assign it later when traversing; if the file system has no mount point, set it to be the same as the current mount point.

curvefs/src/metaserver/trash.cpp
ilixiaocui: When the file arrives in the trash, it must be deleted by the user, so it is good to delete it directly. The current curvefs does not guarantee the correctness of one writing and one deleting in the case of multiple mounts.

As for what to keep the file after the user deletes it by mistake, it depends on what the recycle bin does.

curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -107,107 +107,120 @@ class TrimICacheAsyncDone : public MetaServerClientDone {
     std::shared_ptr<InodeCacheManagerImpl> inodeCacheManager_;
 };
 
-CURVEFS_ERROR InodeCacheManagerImpl::GetInode(uint64_t inodeId,
-    std::shared_ptr<InodeWrapper> &out) {
+#define GET_INODE_REMOTE(FSID, INODEID, OUT, STREAMING)                        \
+    MetaStatusCode ret = metaClient_->GetInode(FSID, INODEID, OUT, STREAMING); \
+    if (ret != MetaStatusCode::OK) {                                           \
+        LOG_IF(ERROR, ret != MetaStatusCode::NOT_FOUND)                        \
```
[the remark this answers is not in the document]
ilixiaocui: fix

curvefs/src/client/inode_cache_manager.cpp
ilixiaocui: As mentioned in [issu](https://github.com/opencurve/curve/issues/1213)，when enableCto, fuseClient.attrTimeOut and  fuseClient.entryTimeOut will be set to 0. So opening the file involves calling: 
```
1. FuseOpGetAttr ino = 1
2. FuseOpLookup parent: 1, name: test62
3. FuseOpGetAttr ino = 7340752
4. FuseOpOpen, ino: 7340752
```
Among them, only 2 and 3 calls will return attr, so just update length and s3info in one of the interface calls. This modification selection is updated in `FuseOpGetAttr`

curvefs/src/client/inode_cache_manager.cpp
ilixiaocui: ditto

curvefs/src/mds/fs_info_wrapper.cpp
ilixiaocui: fix

curvefs/src/metaserver/trash.cpp
ilixiaocui: The recycle bin only gives users a chance to retrieve files under specified conditions, and does not guarantee that they can be retrieved.

curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -107,107 +107,120 @@ class TrimICacheAsyncDone : public MetaServerClientDone {
...
+            << "metaClient_ GetInode failed, MetaStatusCode = " << ret         \
+            << ", MetaStatusCode_Name = " << MetaStatusCode_Name(ret)          \
+            << ", inodeid = " << inodeId;                                      \
```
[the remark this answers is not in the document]
ilixiaocui: fix

cw123 (approved):

curvefs/src/mds/fs_info_wrapper.cpp
SeanHai: so can use "return exist || cto != mp.cto()" instead ?

ilixiaocui: fix

SeanHai (approved):

ilixiaocui:
recheck

PR #1815, opened 2022-08-14
update: optimize LRUCache memory usage (#1815)

fansehep:

Issue Number: #xxx 

Problem Summary:
  Old Design will make the Key construct function call twice.

What's Changed:
 I change the Item key to be a key pointer, std::unordered_map will save the key, so just make the Item.key point to the std::unordered_map<>::iterator->key. 
  By the way, **This LRUCache is not standard**. It's bad to change a standard datastruct for business needs. We can custom a  datastruct for business needs. Maybe some other file just need esay interface such as get or put. It don't need like this "LRUCache" provide some function like GetLast().
How it Works:

### Check List

### DIFF
--- a/src/common/lru_cache.h
+++ b/src/common/lru_cache.h
@@ -139,8 +139,10 @@ template <typename K,  typename V,
     typename ValueTraits = CacheTraits<V>>
 class LRUCache : public LRUCacheInterface<K, V> {
  public:
+    // to less one Key construct
+    // make the Item key be a pointer
     struct Item {
-        K key;
+        const K* key;
         V value;
     };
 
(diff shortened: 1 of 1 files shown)


src/common/lru_cache.h
```diff
@@ -139,8 +139,10 @@ template <typename K,  typename V,
     typename ValueTraits = CacheTraits<V>>
 class LRUCache : public LRUCacheInterface<K, V> {
  public:
+    // to less one Key construct
+    // make the Item key be a pointer
     struct Item {
-        K key;
+        K* key;
```
wu-hanqing, line 145: `const K* key` ?

fansehep: fix

ilixiaocui:
Please merge historical commits into one.

fansehep:
> Please merge historical commits into one.

so but there is only one? right?

fansehep:
> > Please merge historical commits into one.
> 
> so but there is only one? right?

fansehep:
recheck

PR #1823, opened 2022-08-16
curvefs/mds: check fsname before create fs

Cyber-SiKu:
1. skip check s3info if fs has been created
2. add check fsname in create fs
3. add check fsname in toolsv-2

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


Cyber-SiKu:
recheck

curvefs/src/mds/fs_manager.cpp
```diff
@@ -253,11 +254,18 @@ FSStatusCode FsManager::CreateFs(const ::curvefs::mds::CreateFsRequest* request,
     const auto& fsType = request->fstype();
     const auto& detail = request->fsdetail();
 
+    // check fsname
+    if (!std::regex_match(fsName.cbegin(), fsName.cend(),
+                          std::regex(FSNAME_PATTERN))) {
```
wu-hanqing, line 259: Does creating a regex object have runtime cost? If so, it's better to make it a static object, because the pattern is fixed.

curvefs/test/mds/fs_manager_test2.cpp
```diff
@@ -406,5 +404,53 @@ TEST_F(FsManagerTest2, createHybridFs) {
     EXPECT_EQ(FsStatus::INITED, resultInfo.status());
 }
 
+TEST_F(FsManagerTest2, createFsNameInvalid) {
```
wu-hanqing, line 407: suggestion: you can extract fsname check procedure into a free function(neither a member function nor static member function), so, you can directly test fsname check procedure directly rather than writing long meaningless code.

wuhongsong (approved):

wu-hanqing (approved):

PR #1843, opened 2022-08-19
update: use modern style rewrite the uncopyable and fix its' destructor be normal

fansehep:

Issue Number: #xxx 

Problem Summary: The class of Uncopyable is not mordern style and bad.
The old design make this class disconstruct function be virtual 😮‍💨 . 
There are about **20+ class** extends this class in our project. 
So there is a big virtual table in this uncopybale class.

What's Changed:
1. remove the uncopyable class disconstruct func virtual.

How it Works: 

1. I think some coding needs re-review. 😢 

### Check List

### DIFF
--- a/nebd/src/common/uncopyable.h
+++ b/nebd/src/common/uncopyable.h
@@ -28,10 +28,10 @@ namespace common {
 
 class Uncopyable {
  protected:
-    Uncopyable() = default;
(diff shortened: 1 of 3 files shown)


wu-hanqing:
recheck

wu-hanqing (approved):

ilixiaocui (approved):

Issue #1842, opened 2022-08-18
The newly mounted fuse is stuck


YunhuiChen:
**Describe the bug (描述bug)**
The newly mounted fuse is stuck

![image](https://user-images.githubusercontent.com/67582932/185364421-ab725b4a-ed18-424c-ae3f-033254cf483b.png)

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

PR #1844, opened 2022-08-19
curvefs/client: use separate thread to call SetDiskInitUsedBytes(#1842)

wuhongsong:
SetDiskInitUsedBytes may takes a long time, so use a separate thread to do this.
otherwise it will block FuseOpInit

Issue Number: #1842

### DIFF
--- a/curvefs/src/client/fuse_client.h
+++ b/curvefs/src/client/fuse_client.h
@@ -326,7 +326,7 @@ class FuseClient {
         pToken = strtok_r(const_cast<char*>(srcStr.c_str()),
           const_cast<char*>(delimiter.c_str()), &pSave);
         if (nullptr == pToken) {
-            VLOG(3) << "whsdel lookpath end";
+            VLOG(6) << "del lookpath end";
             return;
         }
         splitPath->push_back(pToken);
@@ -334,10 +334,10 @@ class FuseClient {
             pToken = strtok_r(NULL, const_cast<char*>(
               delimiter.c_str()), &pSave);
             if (nullptr == pToken) {
-                VLOG(3) << "whsdel lookpath end";
+                VLOG(6) << "del lookpath end";
                 break;
             }
-            VLOG(9) << "whsdel pToken is:" << pToken
+            VLOG(9) << "del pToken is:" << pToken
                     << "pSave:" << pSave;
             splitPath->push_back(pToken);
         }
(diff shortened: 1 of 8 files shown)


wuhongsong:
recheck

curvefs/src/client/s3/disk_cache_manager.cpp
```diff
@@ -450,8 +450,11 @@ void DiskCacheManager::InitMetrics(const std::string &fsName) {
     cacheWrite_->InitMetrics(metric_);
     cacheRead_->InitMetrics(metric_);
     // this function move to here from init，
-    // Otherwise, you can't get the original metric
-    SetDiskInitUsedBytes();
+    // Otherwise, you can't get the original metric.
+    // SetDiskInitUsedBytes may takes a long time,
+    // so use a separate thread to do this.
+    Thread t(&DiskCacheManager::SetDiskInitUsedBytes, this);
+    t.detach();
```
xu-chaojie, line 457: It is not a good idea to use a detach thread，if thead exec function SetDiskInitUsedBytes after DiskCacheManager destruct，it will cause an exception.

ilixiaocui: +1

xu-chaojie (approved):

curvefs/src/client/s3/disk_cache_manager.cpp
```diff
@@ -450,8 +457,11 @@ void DiskCacheManager::InitMetrics(const std::string &fsName) {
     cacheWrite_->InitMetrics(metric_);
     cacheRead_->InitMetrics(metric_);
     // this function move to here from init，
-    // Otherwise, you can't get the original metric
-    SetDiskInitUsedBytes();
+    // Otherwise, you can't get the original metric.
+    // SetDiskInitUsedBytes may takes a long time,
+    // so use a separate thread to do this.
+    diskInitThread_ = std::thread(
```
xu-chaojie, line 463: use curve::common::Thread

wuhongsong: fix

curvefs/src/client/fuse_client.h
```diff
@@ -326,18 +326,18 @@ class FuseClient {
...
+            VLOG(3) << "del lookpath end";
             return;
         }
         splitPath->push_back(pToken);
         while (true) {
             pToken = strtok_r(NULL, const_cast<char*>(
               delimiter.c_str()), &pSave);
             if (nullptr == pToken) {
-                VLOG(3) << "whsdel lookpath end";
+                VLOG(3) << "del lookpath end";
```
ilixiaocui, line 337: VLOG(6) ?

wuhongsong:
recheck

wuhongsong:
recheck

ilixiaocui (approved):

PR #1860, opened 2022-08-25
curvefs/client: use separate thread to call SetDiskInitUsedBytes(#1842)

wuhongsong:
SetDiskInitUsedBytes may takes a long time, so use a separate thread to do this.
otherwise it will block FuseOpInit

Issue Number: #1842   

Problem Summary:

What's Changed:

How it Works:

### Check List


wuhongsong:
recheck

opencurveadmin (approved):

xu-chaojie (approved):

Issue #1841, opened 2022-08-18
Failed to umount warmup mount point


YunhuiChen:
**Describe the bug (描述bug)**
Failed to umount warmup mount point

**To Reproduce (复现方法)**

sudo curve fs warmup add /home/nbs/failover/test6

nbs@prit2-curve-chunk-node12:~$ curveadm umount /home/nbs/failover/test6  --host node12
Umount FileSystem: [ERROR]  
  + host=node12  mountPoint=/home/nbs/failover/test6 [0/1] [ERROR]  
---
Error-Code: 600000
Error-Description: execute command timed out
Error-Clue: wait container stop failed (docker stop CONTAINER)
How to Solve:
  * Website: https://github.com/opencurve/curveadm/wiki/errno#600000
  * Log: /home/nbs/.curveadm/logs/curveadm-2022-08-18_17-17-15.log
  * WeChat: opencurve_bot
**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

Wine93:
The bug fixed for curveadm `v0.1.4`, maybe you can retry it again.

Issue #1833, opened 2022-08-17
curvefs: warm up bugs


ilixiaocui:
**Describe the bug (描述bug)**
1. There is a 1.5TB disk, and other configurations of the client are: `diskCache.maxUsableSpaceBytes: 1539316278886, diskCache.fullRatio=90, diskCache.safeRatio=70` 

- I have 1.1TB data to warm up，but only warm up 861GB.
- Than set `diskCache.fullRatio=95, diskCache.safeRatio=90`, no data load.

2. When warm up run, `curveadm umount` timeout

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch: release2.4
commit id:

**Additional context/screenshots (更多上下文/截图)**

PR #1862, opened 2022-08-25
curvefs/client: fix warmup bug(#1841)

wuhongsong:
if all objs has cached, then the function will block

Issue Number: #1841  

Issue Number: #1833

### DIFF
--- a/curvefs/src/client/fuse_s3_client.cpp
+++ b/curvefs/src/client/fuse_s3_client.cpp
@@ -304,7 +304,8 @@ void FuseS3Client::WarmUpAllObjs(
             context->cb = cb;
             s3Adaptor_->GetS3Client()->DownloadAsync(context);
         }
-        cond.Wait();
+        if (pendingReq.load())
+            cond.Wait();
     }
 }
 


curvefs/src/client/fuse_s3_client.cpp
```diff
@@ -304,7 +304,8 @@ void FuseS3Client::WarmUpAllObjs(
             context->cb = cb;
             s3Adaptor_->GetS3Client()->DownloadAsync(context);
         }
-        cond.Wait();
+        if (pendingReq.load())
```
xu-chaojie, line 307: It seems useless, please explain why？

xu-chaojie (approved):

curvefs/src/client/fuse_s3_client.cpp
wuhongsong: if all objs has cached, then the function will block

wuhongsong:
recheck

wuhongsong:
recheck

wuhongsong:
recheck

YunhuiChen:
recheck

wuhongsong:
recheck

wuhongsong:
recheck

ilixiaocui (approved):

PR #1868, opened 2022-08-28
curvefs/client: fixed GetLatestTxId rpc request without fsid which le…

Wine93:
…ads to some mountpoints use stale txid to find dentry and not found.

Test Case 1: fixed version
---

* mount `/fs1` to two different mount point (`/root/mount/mnt1`、`/root/mount/mnt2`)
* create files on `/root/mount/mnt1`
```
$ touch 1 2 3
```

* list files on `/root/mount/mnt2` for invoke get all partition firstly

*  rename files on `/root/mount/mnt1`
```
$ mv 1 1.back
$ mv 2 2.back
$ mv 3 3.back
```
* list files on `/root/mount/mnt2`

```
$ ls 
1.back  2.back  3.back
```

* it works well

image

Test Case 2: old version
---

* mount `/fs1` to two different mount point (`/root/mount/mnt1`、`/root/mount/mnt2`)
* create files on `/root/mount/mnt1`
```
$ touch 1 2 3
```

* list files on `/root/mount/mnt2` for invoke get all partition firstly

*  rename files on `/root/mount/mnt1`
```
$ mv 1 1.back
$ mv 2 2.back
$ mv 3 3.back
```
* list files on `/root/mount/mnt2`

```
$ ls 
3
```

* files lost

image

### DIFF
--- a/curvefs/src/client/rpcclient/mds_client.cpp
+++ b/curvefs/src/client/rpcclient/mds_client.cpp
@@ -486,6 +486,7 @@ MdsClientImpl::RefreshSession(const std::vector<PartitionTxId> &txIds,
 FSStatusCode MdsClientImpl::GetLatestTxId(const GetLatestTxIdRequest& request,
                                           GetLatestTxIdResponse* response) {
     auto task = RPCTask {
+        VLOG(3) << "GetLatestTxId [request]: " << request.DebugString();
         mdsClientMetric_.getLatestTxId.qps.count << 1;
         LatencyUpdater updater(&mdsClientMetric_.getLatestTxId.latency);
         mdsbasecli_->GetLatestTxId(request, response, cntl, channel);
@@ -509,6 +510,8 @@ FSStatusCode MdsClientImpl::GetLatestTxId(const GetLatestTxIdRequest& request,
             LOG(WARNING) << "GetLatestTxId fail, errcode = " << rc
                          << ", errmsg = " << FSStatusCode_Name(rc);
         }
+
+        VLOG(3) << "GetLatestTxId [response]: " << response->DebugString();
         return rc;
     };
 
@@ -518,6 +521,7 @@ FSStatusCode MdsClientImpl::GetLatestTxId(const GetLatestTxIdRequest& request,
 
 FSStatusCode MdsClientImpl::CommitTx(const CommitTxRequest& request) {
     auto task = RPCTask {
+        VLOG(3) << "CommitTx [request]: " << request.DebugString();
         mdsClientMetric_.commitTx.qps.count << 1;
         LatencyUpdater updater(&mdsClientMetric_.commitTx.latency);
         CommitTxResponse response;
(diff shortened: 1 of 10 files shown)


curvefs/src/mds/fs_manager.cpp
```diff
@@ -808,6 +808,11 @@ FSStatusCode FsManager::GetFsTxSequence(const std::string& fsName,
 void FsManager::GetLatestTxId(const GetLatestTxIdRequest* request,
                               GetLatestTxIdResponse* response) {
     std::vector<PartitionTxId> txIds;
+    if (!request->has_fsid()) {
+        response->set_statuscode(FSStatusCode::PARAM_ERROR);
```
ilixiaocui, line 812: log error here.

Wine93: > log error here.

fixed

curvefs/src/mds/mds_service.cpp
```diff
@@ -325,15 +325,19 @@ void MdsServiceImpl::GetLatestTxId(
...
 }
 
 void MdsServiceImpl::CommitTx(::google::protobuf::RpcController* controller,
                               const CommitTxRequest* request,
                               CommitTxResponse* response,
                               ::google::protobuf::Closure* done) {
     brpc::ClosureGuard guard(done);
+    VLOG(3) << "CommitTx [request]: " << request->DebugString();
     fsManager_->CommitTx(request, response);
+    VLOG(3) << "CommitTx [response]: " << request->DebugString();
```
SeanHai, line 340: VLOG(6) and log in key point is enough

curvefs/src/client/rpcclient/mds_client.cpp
```diff
@@ -542,15 +546,18 @@ FSStatusCode MdsClientImpl::CommitTx(const CommitTxRequest& request) {
...
     // for rpc error or get lock failed/timeout, we will retry until success
     return ReturnError(rpcexcutor_.DoRPCTask(task, 0));
 }
 
-FSStatusCode MdsClientImpl::GetLatestTxId(std::vector<PartitionTxId>* txIds) {
+FSStatusCode MdsClientImpl::GetLatestTxId(uint32_t fsId,
+                                          std::vector<PartitionTxId>* txIds) {
     GetLatestTxIdRequest request;
     GetLatestTxIdResponse response;
+    request.set_fsid(fsId);
```
SeanHai, line 560: Change to required in proto if fs is required.

SeanHai (approved):

curvefs/src/mds/mds_service.cpp
Wine93: > VLOG(6) and log in key point is enough

we shoule use `VLOG(3)` to let log for troubleshooting more easy.

curvefs/src/client/rpcclient/mds_client.cpp
Wine93: > Change to required in proto if fs is required.

maybe we should use `optional` for compatibility.

ilixiaocui (approved):

Issue #1854, opened 2022-08-23
open/ read stuck after warmup


YunhuiChen:
**Describe the bug (描述bug)**
open/ read stuck after warmup

**To Reproduce (复现方法)**
mount fs to test5 and test6
1、fio --name=big-file-multi-write-2 --directory=/home/nbs/failover/test5 --rw=write --refill_buffers --bs=1M --size=50G --numjobs=1 --end_fsync=1 -fallocate=none
2、mv big-file-multi-write-2.0.0
3、sudo curve fs warmup add /home/nbs/failover/test6/warmuptest/
4、touch renametest
5、echo 111 >  renametest
6、read stuck at test6

![image](https://user-images.githubusercontent.com/67582932/186118607-4cb1290e-4797-4a62-a69b-d33251779651.png)

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

cw123:
![image](https://user-images.githubusercontent.com/4277834/186303345-a9a58a00-070b-4d1d-871f-db099ba646b0.png)
![image](https://user-images.githubusercontent.com/4277834/186303358-53c4ba57-2f04-4286-bb71-2abe94af1bdb.png)
If ReadFromS3 read data from s3 fail, and the error code is s3obj not exist, it will retry forever and stuck at mount point.

h0hmj:
the main loop of s3 sdk was stuck by AsyncRequestInflightBytesThrottle

main loop:
1. make_request, bytes = base + size1 < throttle
2. request failed (maybe 404), goto callback
3. call back phase 1: bytes = base - size1 = base, then notify all
( another stuck request get the chance, bytes = base + size2)
4. call back phase 2: bytes = base + size1 + size2 > throttle
5. request in cb was stuck, and stuck the main loop

```
Thread 26 (Thread 0x7f79f55ff700 (LWP 1972183)):
#0  pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x00007f7a0ba7e50c in std::condition_variable::wait(std::unique_lock<std::mutex>&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
[...]

PR #1874, opened 2022-08-30
avoid deadlock when calling s3async callback, issue #1854

h0hmj:
Signed-off-by: h0hmj <h0hmjcn@gmail.com>

Issue Number: #1854 

Problem Summary:

What's Changed:

add a work queue and a thread to handle async s3 request

How it Works:

move throttle check outside main loop, async s3 request will not block main loop

### Check List

### DIFF
--- a/src/common/s3_adapter.cpp
+++ b/src/common/s3_adapter.cpp
@@ -180,9 +180,12 @@ void S3Adapter::Init(const S3AdapterOption& option) {
         option.maxAsyncRequestInflightBytes == 0
             ? UINT64_MAX
             : option.maxAsyncRequestInflightBytes));
+    asyncCallbackTP_.Start(option.asyncThreadNum);
 }
 
 void S3Adapter::Deinit() {
+    // wait for all previous async requests finished
+    asyncCallbackTP_.Stop();
     // delete s3client in s3adapter
     if (clientCfg_ != nullptr) {
         Aws::Delete<Aws::S3Crt::ClientConfiguration>(clientCfg_);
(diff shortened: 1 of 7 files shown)


h0hmj:
recheck

src/common/s3_adapter.cpp
```diff
@@ -332,6 +335,11 @@ int S3Adapter::PutObject(const Aws::String &key, const std::string &data) {
 */
 
 void S3Adapter::PutObjectAsync(std::shared_ptr<PutObjectAsyncContext> context) {
+    asyncRequestTP_.Enqueue(&S3Adapter::PutObjectAsyncHelper, this, context);
```
wuhongsong, line 338: can cause excessive memory usage？

wu-hanqing: I have the same question

h0hmj: yes, i will change it, seems that just put callbacks into queue is enough.

h0hmj:
let callbacks run in a separate thread rather than the event main loop to avoid deadlock.
num of callback ~= num of request (limited by inflight bytes)

src/common/s3_adapter.h
```diff
@@ -332,6 +333,19 @@ class S3Adapter {
         std::condition_variable cond_;
     };
 
+    class AsyncCallbackThreadPool : public TaskThreadPool<> {
```
wu-hanqing, line 336: it seems this derived class is unnecessary, we can add an extra parameter to control whether run all tasks when task pool stopped.

```
    virtual void ThreadFunc() {
        while (running_.load(std::memory_order_acquire)
|| (drain && !queue.empty())
) {
            Task task(Take());
            /* ThreadPool 退出的时候，queue 为空，那么会返回无效的 task */
            if (task) {
                task();
            }
        }
    }
```

src/common/s3_adapter.cpp
```diff
@@ -340,13 +343,18 @@ void S3Adapter::PutObjectAsync(std::shared_ptr<PutObjectAsyncContext> context) {
...
-         originCallback](const std::shared_ptr<PutObjectAsyncContext>& ctx) {
+         originCallback](const std::shared_ptr<PutObjectAsyncContext> ctx) {
             inflightBytesThrottle_->OnComplete(ctx->bufferSize);
             ctx->cb = originCallback;
             ctx->cb(ctx);
         };
+    auto asyncCallback =
+        [this,
+         throttledCallback](const std::shared_ptr<PutObjectAsyncContext>& ctx) {
+            asyncCallbackTP_.Enqueue(std::move(throttledCallback), ctx);
```
wu-hanqing, line 356: can we just do the Enqueue in L378 rather than wrap another call back?

h0hmj: yes, i'll put it in handle function

wu-hanqing (approved):

Issue #1850, opened 2022-08-21
inode's timestamp is not fixed when encountered metaserver failover/restart


wu-hanqing:
Inode's timestamp(atime/ctime/mtime) is initialized in metaserver when creating an inode, but this has two problems:
1. The create inode request is executed in three replicas separately, which means the inode in three replicas may have a different timestamp, so if raft group's leader transferred, we will see timestamp changed automatically.
2. If a create inode request is not included in raft snapshot, the request will execute again, which also results in the inode's timestamp changed automatically.

And this should not meet the POSIX standard.

https://github.com/opencurve/curve/blob/6b7d5dd2d94ce487be81af7015f6a6ab623d1b7e/curvefs/src/metaserver/inode_manager.cpp#L101

fansehep:
@wu-hanqing Close it. 😄

PR #1883, opened 2022-09-01
curvefs/metaserver: fix inode create time inconsistent

fansehep:

Issue Number: #1850 

Problem Summary: see issue

What's Changed: add fuse client rpc to the three replicas tv_sec(sec) and tv_nsec(ns).

How it Works:

### Check List

### DIFF
--- a/curvefs/proto/metaserver.proto
+++ b/curvefs/proto/metaserver.proto
@@ -242,6 +242,12 @@ message CreateInodeRequest {
     required uint64 parent = 10;
     optional uint64 rdev = 11;
     optional string symlink = 12;   // TYPE_SYM_LINK only
+    optional Time create = 13;
+}
+
+message Time {
+    required uint64 sec = 1;
+    required int64 nsec = 2;
 }
 
 message CreateInodeResponse {
@@ -258,6 +264,7 @@ message CreateRootInodeRequest {
     required uint32 uid = 5;
     required uint32 gid = 6;
     required uint32 mode = 7;
+    optional Time create = 8;
 }
 
 message CreateRootInodeResponse {
--- a/curvefs/src/metaserver/inode_manager.cpp
+++ b/curvefs/src/metaserver/inode_manager.cpp
@@ -27,6 +27,7 @@
 #include <list>
 #include <unordered_set>
 #include <utility>
+#include <ctime>
 
 #include "curvefs/proto/metaserver.pb.h"
 #include "curvefs/src/common/define.h"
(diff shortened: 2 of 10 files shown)


fansehep:
recheck

curvefs/src/metaserver/metastore.cpp
```diff
@@ -453,6 +453,8 @@ MetaStatusCode MetaStoreImpl::CreateInode(const CreateInodeRequest* request,
     param.type = request->type();
     param.parent = request->parent();
     param.rdev = request->rdev();
+    param.tv_sec = request->createsec();
+    param.tv_nsec = request->creatensec();
```
wu-hanqing, line 457: Consider backward compatibility, even though is a bug, because we cannot upgrade all the components at once.

For most cases, we upgrade server-side components first, then the client-side.

So, in this case, when we upgrade metaserver to a newer version, these two fields are missing if the older client sends create inode requests which mean `tv_sec` and `tv_nsec` will be 0, then we end up with another bug.

A proper way is to test whether the request has `createsec` and `creatensec`, if it has, then we use these two timestamps, otherwise, we fallback to previous buggy version.

And for `InodeParam`, we can use `absl::optional` to wrap a timespec, like
```cpp
struct InodeParam {
   // other fields
   absl::optional<timespec> timestamp;
};
```

And in here, we can do this
```cpp
if (request->has_createsec() && request->has_creatensec()) {
    param.timestamp.emplace(request->createsec(), request->creatensec());
}
```

curvefs/proto/metaserver.proto
```diff
@@ -242,6 +242,8 @@ message CreateInodeRequest {
     required uint64 parent = 10;
     optional uint64 rdev = 11;
     optional string symlink = 12;   // TYPE_SYM_LINK only
+    optional uint64 createsec = 13;
```
wu-hanqing, line 245: `CreateRootInodeRequest` also has the same problem.

curvefs/proto/metaserver.proto
```diff
@@ -242,6 +242,8 @@ message CreateInodeRequest {
...
+    optional uint64 creatensec = 14;
```
wu-hanqing, line 246:
```proto
message Time {
    required uint64_t sec = 1;
    required int64_t nsec = 2;
}

message CreateInodeRequest {
    // other fields
    optional Time create;
}
```

fansehep: > ```proto
> message Time {
>     required uint64_t sec = 1;
>     required int64_t nsec = 2;
> }
> 
> message CreateInodeRequest {
>     // other fields
>     optional Time create;
> }
> ```

Great Design!
fix.

curvefs/proto/metaserver.proto
fansehep: > `CreateRootInodeRequest` also has the same problem.

fix

curvefs/src/metaserver/metastore.cpp
fansehep: > Consider backward compatibility, even though is a bug, because we cannot upgrade all the components at once.
> 
> For most cases, we upgrade server-side components first, then the client-side.
> 
> So, in this case, when we upgrade metaserver to a newer version, these two fields are missing if the older client sends create inode requests which mean `tv_sec` and `tv_nsec` will be 0, then we end up with another bug.
> 
> A proper way is to test whether the request has `createsec` and `creatensec`, if it has, then we use these two timestamps, otherwise, we fallback to previous buggy version.
> 
> And for `InodeParam`, we can use `absl::optional` to wrap a timespec, like
> 
> ```c++
> struct InodeParam {
>    // other fields
>    absl::optional<timespec> timestamp;
> };
> ```
> 
> And in here, we can do this
> 
> ```c++
> if (request->has_createsec() && request->has_creatensec()) {
>     param.timestamp.emplace(request->createsec(), request->creatensec());
> }
> ```

fix.

wu-hanqing (approved):

curvefs/src/metaserver/inode_manager.cpp
```diff
@@ -112,14 +113,23 @@ void InodeManager::GenerateInodeInternal(uint64_t inodeId,
...
 
-    struct timespec now;
-    clock_gettime(CLOCK_REALTIME, &now);
-    inode->set_mtime(now.tv_sec);
-    inode->set_mtime_ns(now.tv_nsec);
-    inode->set_atime(now.tv_sec);
-    inode->set_atime_ns(now.tv_nsec);
-    inode->set_ctime(now.tv_sec);
-    inode->set_ctime_ns(now.tv_nsec);
+    if (param.timestamp.has_value()) {
```
ilixiaocui, line 116: Add unit tests here

fansehep:
recheck

fansehep:
recheck

fansehep:
recheck

fansehep:
recheck

fansehep:
It seems CI is not unstable.

ilixiaocui (approved):

Issue #1851, opened 2022-08-22
Warmup 50G files written sequentially, the cache capacity exceeds 70G 


YunhuiChen:
**Describe the bug (描述bug)**

**To Reproduce (复现方法)**
mount fs1 to test1 and test3
1、fio --name=big-file-multi-write --directory=/home/nbs/failover/test1 --rw=write --refill_buffers --bs=1M --size=50G --numjobs=1 --end_fsync=1 -fallocate=none
2、sudo curve fs warmup add /home/nbs/failover/test3/big-file-multi-write.0.0
![image](https://user-images.githubusercontent.com/67582932/185888126-4f2d4799-d454-40f2-b885-8f816de9aff6.png)
![image](https://user-images.githubusercontent.com/67582932/185888195-e5dd3c1e-9359-42df-b4bf-7be469d47157.png)

17918 * 4m/1024 = 70G 
**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

ilixiaocui:
cc @wuhongsong 
Please continue to follow up and fix this problem.

wuhongsong:
> cc @wuhongsong Please continue to follow up and fix this problem.

There is no good solution to this possible data redundancy problem.

Cyber-SiKu:
![image](https://user-images.githubusercontent.com/15775037/211981967-779895f0-8174-4723-9c42-a62407c21760.png)

![image](https://user-images.githubusercontent.com/15775037/211982034-ec5bf868-7a8e-4d39-80b9-3d91e3ca9432.png)

PR #1893, opened 2022-09-05
curvefs/client:fix warmup and read bug(#1851)

wuhongsong:
warmup will get old obj(compact not done) in the scene overwrite;
read data is not full if length is larger than blocksize.

Issue Number: #1851

### DIFF
--- a/curvefs/src/client/fuse_s3_client.cpp
+++ b/curvefs/src/client/fuse_s3_client.cpp
@@ -132,7 +132,6 @@ void FuseS3Client::BackGroundFetch() {
 }
 
 void FuseS3Client::fetchDataEnqueue(fuse_ino_t ino) {
-    VLOG(9) << "fetchDataEnqueue start: " << ino;
     auto task = [this, ino]() {
         std::shared_ptr<InodeWrapper> inodeWrapper;
         CURVEFS_ERROR ret = inodeManager_->GetInode(ino, inodeWrapper);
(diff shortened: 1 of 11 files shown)


wuhongsong:
recheck

wuhongsong:
recheck

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -572,6 +597,15 @@ int FileCacheManager::ReadFromS3(const std::vector<S3ReadRequest> &requests,
                                << ", len: " << n << ", ret is: " << ret;
                     return ret;
                 }
+                if (warmup) {
+                    ret = s3ClientAdaptor_->GetDiskCacheManager()->
+                      WriteReadDirect(name, response.GetDataBuf() +
+                        readOffset + blockPos - objectOffset, n);
+                    if (ret < 0) {
+                        LOG_EVERY_SECOND(INFO) <<
```
wu-hanqing, line 605: warning log ?

cw123:
warm up is Warm up and normal read requests follow the same path. Will warm up affect normal read requests?

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -633,20 +668,21 @@ class AsyncPrefetchCallback {
...
         }
+
         {
             curve::common::LockGuard lg(fileCache->downloadMtx_);
             fileCache->downloadingObj_.erase(context->key);
         }
     }
 
  private:
+    bool needStore_;
```
cw123, line 685: where does this param use?

curvefs/src/client/fuse_s3_client.cpp
```diff
@@ -141,19 +140,18 @@ void FuseS3Client::fetchDataEnqueue(fuse_ino_t ino) {
...
-        google::protobuf::Map<uint64_t, S3ChunkInfoList> *s3ChunkInfoMap
-          = nullptr;
-        {
-            ::curve::common::UniqueLock lgGuard = inodeWrapper->GetUniqueLock();
-            s3ChunkInfoMap = inodeWrapper->GetChunkInfoMap();
-        }
-        if (nullptr == s3ChunkInfoMap ||
-          s3ChunkInfoMap->empty()) {
+        uint64_t fileSize = inodeWrapper->GetLength();
+        std::unique_ptr<char[]> buffer(new char[fileSize]);
```
cw123, line 144: if fileSize is too large, the new may fail

wuhongsong: there are maybe many files , and file maybe large too, so the new() for read()  may cause memory problem

wuhongsong:
there are maybe many files , and file maybe large too, so the new() for read() may cause memory problem

so close the pr

PR #1901, opened 2022-09-07
curvefs/client: fix update nlink error

SeanHai:

fix the update inode nlink logic, the previous logic will update nlink incorrect at:
未命名绘图

The fixed logic is every update of nlink to metaserver should refresh from metaserver first.

Issue Number: #1904  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/fuse_client.cpp
+++ b/curvefs/src/client/fuse_client.cpp
@@ -565,8 +565,8 @@ CURVEFS_ERROR FuseClient::FuseOpOpen(fuse_req_t req, fuse_ino_t ino,
     return ret;
 }
 
-CURVEFS_ERROR FuseClient::UpdateParentInodeMCTimeAndInvalidNlink(
-    fuse_ino_t parent, FsFileType type) {
+CURVEFS_ERROR FuseClient::UpdateParentMCTimeAndNlink(
+    fuse_ino_t parent, FsFileType type, NlinkChange nlink) {
     std::shared_ptr<InodeWrapper> parentInodeWrapper;
     auto ret = inodeManager_->GetInode(parent, parentInodeWrapper);
     if (ret != CURVEFS_ERROR::OK) {
@@ -580,9 +580,10 @@ CURVEFS_ERROR FuseClient::UpdateParentInodeMCTimeAndInvalidNlink(
         parentInodeWrapper->UpdateTimestampLocked(kModifyTime | kChangeTime);
 
         if (FsFileType::TYPE_DIRECTORY == type) {
-            parentInodeWrapper->InvalidateNlink();
+            parentInodeWrapper->UpdateNlinkLocked(nlink);
         }
     }
+    inodeManager_->ShipToFlush(parentInodeWrapper);
     return CURVEFS_ERROR::OK;
 }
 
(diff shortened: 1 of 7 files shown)


SeanHai:
recheck

SeanHai:
recheck

curvefs/src/client/fuse_client.h
```diff
@@ -348,8 +348,8 @@ class FuseClient {
 
     virtual void FlushData() = 0;
 
-    CURVEFS_ERROR UpdateParentInodeMCTimeAndInvalidNlink(
-        fuse_ino_t parent, FsFileType type);
+    CURVEFS_ERROR UpdateParentMCTimeAndNlink(
+        fuse_ino_t parent, FsFileType type, int32_t nlink);
```
wu-hanqing, line 352: As for the third argument `nlink`, I only saw you passed either `1` or `-1`, does other values exist, like 2/3 or -2/-3? And if not, a proper way is using an enum to represent 1 or -1.

curvefs/src/client/fuse_client.cpp
```diff
@@ -580,7 +580,7 @@ CURVEFS_ERROR FuseClient::UpdateParentInodeMCTimeAndInvalidNlink(
         parentInodeWrapper->UpdateTimestampLocked(kModifyTime | kChangeTime);
 
         if (FsFileType::TYPE_DIRECTORY == type) {
-            parentInodeWrapper->InvalidateNlink();
+            parentInodeWrapper->UpdateNlinkLocked(nlink);
```
wu-hanqing, line 583: Is there a bug here? we changed the timestamp and nlink of an inode, but we don't call `ShipToFlush` which means the updated fields don't persist to metaserver?

curvefs/src/client/inode_wrapper.h
```diff
@@ -312,6 +262,14 @@ class InodeWrapper : public std::enable_shared_from_this<InodeWrapper> {
         return s3ChunkInfoAdd_.empty();
     }
 
+    uint32_t GetNlinkLocked() {
+        return inode_.nlink();
+    }
+
+    void UpdateNlinkLocked(uint32_t nlink) {
+        inode_.set_nlink(inode_.nlink() + nlink);
```
wu-hanqing, line 270: Should we mark inode dirty and also set corresponding field in dirtyAttr?

curvefs/src/client/inode_wrapper.h
```diff
@@ -52,13 +52,11 @@ constexpr int kAccessTime = 1 << 0;
 constexpr int kChangeTime = 1 << 1;
 constexpr int kModifyTime = 1 << 2;
 
-#define REFRESH_NLINK_IF_NEED               \
+#define REFRESH_NLINK                       \
```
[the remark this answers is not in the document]
SeanHai: fixed

curvefs/src/client/fuse_client.cpp
SeanHai: The update of nlink is done in metaserver and client at same time, the nlink in client is just to show, it need refresh first from metaserver when update to metaserver.

The update of time will set inode to dirty and flush backgroud

curvefs/src/client/fuse_client.h
SeanHai: fixed

SeanHai: fixed

Wine93 (approved):
LGTM!

wu-hanqing (approved):

PR #1928, opened 2022-09-20
curvefs/client: fix throttle bug

wuhongsong:
burst is 0 but burstlenth is non-zero,so the function of SetLimit not work default.

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


ilixiaocui (approved):

wu-hanqing (approved):

wuhongsong:
recheck

ilixiaocui (approved):

PR #1939, opened 2022-09-27
curvefs: fix batch get inodeattr when readdir

SeanHai:

1. write some data to file and the inode attr in icache not sync to metaserver.
2. readdir will batch get inode attr from metasever and cached in vfs. But the length of inode is older.
3. read the data new written will failed.

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/inode_cache_manager.cpp
+++ b/curvefs/src/client/inode_cache_manager.cpp
@@ -206,8 +206,8 @@ CURVEFS_ERROR InodeCacheManagerImpl::GetInodeAttr(uint64_t inodeId,
 
 CURVEFS_ERROR InodeCacheManagerImpl::BatchGetInodeAttr(
     std::set<uint64_t> *inodeIds,
-    std::list<InodeAttr> *attr) {
-    // get some inode in icache
+    std::list<InodeAttr> *attrs) {
+    // get some inode attr in icache
     for (auto iter = inodeIds->begin(); iter != inodeIds->end();) {
         std::shared_ptr<InodeWrapper> inodeWrapper;
         NameLockGuard lock(nameLock_, std::to_string(*iter));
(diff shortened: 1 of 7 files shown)


curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -237,21 +238,36 @@ CURVEFS_ERROR InodeCacheManagerImpl::BatchGetInodeAttr(
...
+    // get some inode attr in icache
+    for (auto iter = inodeIds->begin(); iter != inodeIds->end();) {
+        std::shared_ptr<InodeWrapper> inodeWrapper;
+        NameLockGuard lock(nameLock_, std::to_string(*iter));
+        bool ok = iCache_->Get(*iter, &inodeWrapper);
+        if (ok) {
+            InodeAttr tmpAttr;
+            inodeWrapper->GetInodeAttr(&tmpAttr);
+            attrs->emplace(*iter, tmpAttr);
+            iter = inodeIds->erase(iter);
```
wuhongsong, line 258: The return value is ambiguous in different c++ version，so i think It's best not to use it directly

wuhongsong (approved):

curvefs/src/client/inode_cache_manager.cpp
```diff
@@ -237,21 +238,36 @@ CURVEFS_ERROR InodeCacheManagerImpl::BatchGetInodeAttr(
...
-        return CURVEFS_ERROR::OK;
-    }
-
     NameLockGuard lg(asyncNameLock_, std::to_string(parentId));
     bool ok  = iAttrCache_->Get(parentId, attrs);
     if (ok) {
         return CURVEFS_ERROR::OK;
     }
 
+    // get some inode attr in icache
```
wu-hanqing, line 249: Should we get attrs from `iCache_` first, then `iAttrCache_`? In other words, can `iAttrCache_` has outdated attributes in the same scenario?

SeanHai:
recheck

curvefs/src/client/inode_cache_manager.cpp
SeanHai: fixed

wu-hanqing (approved):

Issue #1838, opened 2022-08-18
After curveadm clean metaserver, the interface needs to be set to offline


YunhuiChen:
**Describe the bug (描述bug)**
After curveadm clean metaserver, the interface needs to be set to offline
1、stop 
2、clean
3、修改stop的metaserver的数据路径,提交修改
4、deploy

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

cw123:
@SeanHai will modify the pr later

wuhongsong:
@SeanHai

PR #1943, opened 2022-09-28
curvefs: support set metaserver offline

SeanHai:

Issue Number: #xxx 
#1838 

Problem Summary:

What's Changed:

How it Works:

### Check List


curvefs/src/mds/topology/topology.cpp
```diff
@@ -536,6 +536,23 @@ bool TopologyImpl::GetMetaServer(const std::string &hostIp, uint32_t port,
...
+bool TopologyImpl::IsMetaServerReRegistered(MetaServerIdType msId) {
+    MetaServer ms;
+    if (GetMetaServer(msId, &ms)) {
+        ReadLockGuard rlockMetaServerMap(metaServerMutex_);
+        for (auto it = metaServerMap_.begin(); it != metaServerMap_.end();
+            it++) {
+            ReadLockGuard rlockMetaServer(it->second.GetRWLockRef());
+            if (it->second.GetInternalIp() == ms.GetInternalIp() &&
+                it->second.GetInternalPort() == ms.GetInternalPort() &&
+                it->second.GetId() != ms.GetId()) {
```
Cyber-SiKu, line 548: What happens: 
ip:port is the same, but msid is different

PR #1989, opened 2022-10-19
curvefs/client: fix bug that s3 deadlock

wuhongsong:

Issue Number: #1854


wuhongsong:
recheck

wuhongsong:
recheck

src/common/s3_adapter.cpp
```diff
@@ -478,7 +489,12 @@ void S3Adapter::GetObjectAsync(std::shared_ptr<GetObjectAsyncContext> context) {
         throttle_->Add(true, context->len);
     }
 
-    inflightBytesThrottle_->OnStart(context->len);
+    if (inflightBytesThrottle_->OnStart(context->len) < 0) {
+        context->retCode = -1;
+        context->retryIntervalUS = retryIntervalUS_;
+        asyncCallbackTP_.Enqueue(context->cb, this, context);
```
h0hmj, line 495: too many throttled requests will make callback queue too large?

h0hmj: put/get 10 objects async requests from the same function, a for loop

before: 1st success, 2nd throttled (on start cond_.wait), 3\~10 blocked
now: 1st success, 2\~10 throttled but still put callbacks to queue


if for loop is too large, such as download all warmup objs (>100,000 maybe), queue size maybe too large? how about the memory consumption? need more test data.

ilixiaocui:
Modify and post the test results later. Or note the cause of the deadlock in the release.

wuhongsong:
the new aws sdk has been revert

Issue #2033, opened 2022-11-04
[compile] update the braft vesion


cw123:
**Describe the feature you'd like:**

curve uses braft as the consensus protocol, the braft version we are currently using is too old. Hope to be able to upgrade braft to a newer version.

The external dependencies used by curve are documented in https://github.com/opencurve/curve/blob/master/WORKSPACE. The currently used version of braft is:

curve使用braft作为一致性协议，我们当前使用的braft版本太老了。希望能够把braft的升级到一个比较新的稳定版本。

curve使用的外部依赖库记录在 https://github.com/opencurve/curve/blob/master/WORKSPACE 这个文件中。其中braft的当前使用的版本为：

```
git_repository(
    name = "com_github_baidu_braft",
    remote = "https://github.com/baidu/braft",
    commit = "e255c0e4b18d1a8a5d484d4b647f41ff1385ef1e",
)
```

Replace the old commit with the new version of the commit, and then `make build` compiles the entire curve project, some compilation errors may occur. You need to find a way to solve these compilation problems and complete the version replacement of braft.

使用新版本的commit替换这个旧的commit，然后`make build`编译整个curve工程，可能会出现一些编译错误。你需要想办法解决这些编译问题，完成braft的版本替换。

**Teachability, Documentation, Adoption, Migration Strategy:**

Build compilation environment：https://github.com/opencurve/curve/blob/master/docs/cn/build_and_run.md.

PR #2063, opened 2022-11-10
[compile] update the braft vesion

tangwz:

Issue Number: #2033 

Problem Summary:

Replace the old commit with the new version of the commit, and solve some build problem.

What's Changed:

How it Works:

Side effects(Breaking backward compatibility? Performance regression?): None

### Check List

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -34,7 +34,7 @@ bazel_skylib_workspace()
 git_repository(
     name = "com_github_baidu_braft",
     remote = "https://github.com/baidu/braft",
-    commit = "e255c0e4b18d1a8a5d484d4b647f41ff1385ef1e",
+    commit = "d0277bf2aea66908d1fd7376d191bd098371966e",
 )
 
 bind(
@@ -116,7 +116,7 @@ bind(
 
 http_archive(
     name = "com_github_google_leveldb",
-    build_file = "@com_github_apache_brpc//:leveldb.BUILD",
+    build_file = "@com_github_brpc_brpc//:leveldb.BUILD",
     strip_prefix = "leveldb-a53934a3ae1244679f812d998a4f16f2c7f309a6",
     urls = ["https://github.com/google/leveldb/archive/a53934a3ae1244679f812d998a4f16f2c7f309a6.tar.gz"],
 )
@@ -127,7 +127,7 @@ bind(
 )
 
 git_repository(
-    name = "com_github_apache_brpc",
+    name = "com_github_brpc_brpc",
     remote = "https://github.com/apache/incubator-brpc",
     commit = "1b9e00641cbec1c8803da6a1f7f555398c954cb0",
     patches = ["//:thirdparties/brpc/brpc.patch"],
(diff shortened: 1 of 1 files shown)


src/chunkserver/raftlog/curve_segment_log_storage.cpp
```diff
@@ -422,7 +422,7 @@ int CurveSegmentLogStorage::append_entry(const braft::LogEntry* entry) {
 }
 
 int CurveSegmentLogStorage::append_entries(
-                    const std::vector<braft::LogEntry*>& entries) {
+                    const std::vector<braft::LogEntry*>& entries, braft::IOMetric* metric) {
```
ilixiaocui, line 425: Why do you need to add a parameter? Is the interface of the new version of brake replaced

ilixiaocui:
Thanks you for this pr!
1. Update the latest version of the master to the local through rebase
2 use git commit -s -a **** to satisfy DCO requirement

WORKSPACE
```diff
@@ -34,14 +34,20 @@ bazel_skylib_workspace()
 git_repository(
     name = "com_github_baidu_braft",
     remote = "https://github.com/baidu/braft",
-    commit = "e255c0e4b18d1a8a5d484d4b647f41ff1385ef1e",
+    commit = "351c74aca6f7d5436da0c2e06c93067659c31435",
```
wu-hanqing, line 37: it's better to use a stable version rather than a mainline version, and the latest stable is v1.1.2, https://github.com/baidu/braft/releases.

wu-hanqing (commented):
thanks for your contribution, 👍

src/chunkserver/raftlog/curve_segment_log_storage.cpp
tangwz: Virtual function `virtual int append_entries(const std::vector<LogEntry*>& entries, IOMetric* metric) = 0;`, so we must break this, or do not update so high version.

WORKSPACE
tangwz: I use the stable version v1.0.2. Because the newer version's function `append_entries` need add a parameter.

tangwz:
@wu-hanqing @ilixiaocui PTAL

wu-hanqing (approved):
LGTM.

@ilixiaocui Should we start a round of failover testing before or after merging this pr?

ilixiaocui:
> LGTM.
> 
> @ilixiaocui Should we start a round of failover testing before or after merging this pr?

It's a necessity.

ilixiaocui (dismissed):

WORKSPACE
wu-hanqing: > I use the stable version v1.0.2. Because the newer version's function `append_entries` need add a parameter.

Did you mean `SegmentLogStorage::append_entries` added another parameter `metric`? If so, I think we can just ignore this parameter, so that, we can use braft v1.1.2.

wu-hanqing: Your first version is acceptable,
image

ilixiaocui (approved):

Issue #2070, opened 2022-11-14
Async read need


wuhongsong:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

in function of ReadFromS3, the default is sync read

**Describe the solution you'd like (描述你期望的解决方法)**

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

YLShiJustFly:
Dear Song, Let me have a try.

wuhongsong:
@YLShiJustFly

Ziy1-Tan:
It seem that `ReadFromS3` is not existed, do you mind to update desc?

wuhongsong:
> It seem that `ReadFromS3` is not existed, do you mind to update desc?

yep，`FileCacheManager::ReadKVRequest` has replaced， so maybe we can use asynchronies io in it

PR #2080, opened 2022-11-16
add async in read from s3

YLShiJustFly:
Signed-off-by: YLShi <344515349@qq.com>

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/s3/client_s3_cache_manager.cpp
+++ b/curvefs/src/client/s3/client_s3_cache_manager.cpp
@@ -452,7 +452,7 @@ int FileCacheManager::ReadFromS3(const std::vector<S3ReadRequest> &requests,
     std::vector<S3ReadRequest>::const_iterator iter = requests.begin();
     std::atomic<uint64_t> pendingReq(0);
     curve::common::CountDownEvent cond(1);
-    bool async = false;
+    bool async = true;
     // first is chunkIndex, second is vector chunkPos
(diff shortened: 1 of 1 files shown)


wuhongsong:
recheck

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -532,14 +532,35 @@ int FileCacheManager::ReadFromS3(const std::vector<S3ReadRequest> &requests,
             uint64_t start = butil::cpuwide_time_us();
             if (async) {
```
wuhongsong, line 533: if async is enable, you can set（bool async = true;)

cw123:
ReadFromS3的实现中，还是默认的sync read。你的修改是解决 #2070 这个问题的吗？

In function of ReadFromS3, the default is still sync read. 
Does your change address #2070?

wuhongsong:
recheck

cw123 (commented):
1. most read logic of sync read between async are same,  extract common logic
2.  add some unit test case

cw123:
recheck

cw123:
recheck

ilixiaocui:
This pr has been opened for a long time, and the review comments have not been replied. Please resubmit the pr if necessary.

PR #2123, opened 2022-12-01
update bazelrc

fansehep:

Issue Number: #xxx 

Problem Summary: update ```.gitignore```.

What's Changed:

How it Works:

### Check List

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -3,3 +3,5 @@ build --copt -DHAVE_ZLIB=1 --copt -DGFLAGS_NS=google --copt -DUSE_BTHREAD_MUTEX
 build --cxxopt -Wno-error=format-security
 build:gcc7-later --cxxopt -faligned-new
 build --incompatible_blacklisted_protos_requires_proto_info=false
+build --copt=-fdiagnostics-color=always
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -121,8 +121,7 @@ thirdparties/rocksdb/rocksdb/
 thirdparties/rocksdb/*log
 thirdparties/rocksdb/*.tar.gz
 thirdparties/aws/*.tar.gz
-thirdparties/etcdclient/tmp/
-thirdparties/etcdclient/*.h
+thirdparties/etcdclient/*
```
ilixiaocui, line 124: Modifications to files other than `tmp`  under `thirdparties/etcdclient` cannot be ignored.

fansehep: We usually do not change the contents of this folder, instead when compiling the project, a lot of extra files are created in this directory.

fansehep: > Modifications to files other than `tmp` under `thirdparties/etcdclient` cannot be ignored.




here you can see it's inconvenient here. And someone use vscode git plugin. so ugly here.

fansehep:
recheck

PR #2136, opened 2022-12-06
Fix uint64_t type error

201341:
Signed-off-by: swj <1186093704@qq.com>

Issue Number: #xxx 

Problem Summary:
uint64_t is alway >=0, 

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/s3/client_s3_cache_manager.cpp
+++ b/curvefs/src/client/s3/client_s3_cache_manager.cpp
@@ -243,17 +243,18 @@ int FileCacheManager::Write(uint64_t offset, uint64_t length,
     uint64_t chunkPos = offset % chunkSize;
     uint64_t writeLen = 0;
     uint64_t writeOffset = 0;
+    int64_t len = length;
 
-    while (length > 0) {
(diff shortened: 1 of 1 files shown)


curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -243,17 +243,18 @@ int FileCacheManager::Write(uint64_t offset, uint64_t length,
     uint64_t chunkPos = offset % chunkSize;
     uint64_t writeLen = 0;
     uint64_t writeOffset = 0;
+    int64_t len = length;
```
ilixiaocui, line 246: why change uint64_t to int64_t

201341: the code use the condition (length > 0) to break the loop, however the uint64_t is always >= 0

ilixiaocui: > the code use the condition (length > 0) to break the loop, however the uint64_t is always >= 0 @ilixiaocui

`length = 0 `will break the loop.

PR #2165, opened 2022-12-14
[feat]curvefs: support memcached in container

Cyber-SiKu:
1. add support in entrypoint.sh
2. add memcached install in bash image
3. copy dir by module when making image, which increases the number of layers of the image, but reduces the burden of uploading

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

curvefs/docker/debian10/Dockerfile
```diff
@@ -1,11 +1,15 @@
 FROM opencurvedocker/curve-base:debian10
 ENV TZ=Asia/Shanghai
 RUN mkdir -p /curvefs /etc/curvefs /core /etc/curve
-COPY curvefs /curvefs
+COPY curvefs/client /curvefs
```
Wine93, line 4: I think only one OS Dockerfile needs to be kept for running curve in docker (maybe debian9 or debian10), it's too many to maintain.

curvefs/docker/debian10/entrypoint.sh
```diff
@@ -87,6 +87,14 @@ function prepare() {
             g_binary="$g_prefix/sbin/curve-fuse"
             g_start_args="--confPath $conf_path"
             ;;
+        monitor)
+            g_binary="python3"
+            g_start_args="target_json.py"
+            ;;
+        memcached)
+            g_prefix="/memcached"
```
Wine93, line 95: Whether memcache is part of the curvefs system (like snapshot clone server)? If so, i think place it to `/curvefs` directory is better.

curvefs/docker/debian10/Dockerfile
```diff
@@ -1,11 +1,15 @@
...
+COPY curvefs/mds /curvefs
+COPY curvefs/metaserver /curvefs
+COPY curvefs/monitor /curvefs
+COPY curvefs/tools /curvefs
 COPY entrypoint.sh /
 COPY curvefs/tools/sbin/curvefs_tool /usr/bin
-COPY libmemcached.so /usr/lib/
-COPY libmemcached.so.11 /usr/lib/
-COPY libhashkit.so.2 /usr/lib
+COPY libmemcached.so libmemcached.so.11 libhashkit.so.2 /usr/lib/
```
Wine93, line 13: 👍

curvefs/docker/debian10/Dockerfile
```diff
@@ -1,11 +1,15 @@
 FROM opencurvedocker/curve-base:debian10
 ENV TZ=Asia/Shanghai
 RUN mkdir -p /curvefs /etc/curvefs /core /etc/curve
-COPY curvefs /curvefs
+COPY curvefs/client /curvefs
+COPY curvefs/conf /curvefs
```
Wine93, line 5: Why use multi `COPY` commands to copy curvefs directory?

Cyber-SiKu: > Why use multi `COPY` commands to copy curvefs directory?

Most of the time, only one or two modules in client, mds or metaserver may be updated, but the entire /curvefs directory needs to be updated, so the entire directory needs to be updated when pushing or pulling.

curvefs/docker/debian10/Dockerfile
Cyber-SiKu: > I think only one OS Dockerfile needs to be kept for running curve in docker (maybe debian9 or debian10), it's too many to maintain.

removed

curvefs/docker/debian10/entrypoint.sh
Cyber-SiKu: > Whether memcache is part of the curvefs system (like snapshot clone server)? If so, i think place it to `/curvefs` directory is better.

Will memcached be used by other systems in the future? Isn't this bound to curvefs?

Cyber-SiKu:
cicheck

wu-hanqing (approved):

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Wine93 (approved):

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

PR #2170, opened 2022-12-17
refator: `std::exception` should be catched by reference

Ziy1-Tan:

Problem Summary:
`std::exception` should be passed by reference in catch clause, otherwise in later versions of gcc (eg: 8.3) this will be treated as an error.

What's Changed:

pass by reference.

How it Works:

### Check List


fansehep:
Can you paste the compile error?

Cyber-SiKu:
will fix in this [pr](https://github.com/opencurve/curve/pull/2165)

Cyber-SiKu:
fixed by this [pr](https://github.com/opencurve/curve/pull/2165)

Cyber-SiKu:
fixed by this [pr](https://github.com/opencurve/curve/pull/2165)

Issue #1583, opened 2022-06-16
Reduce unnecessary string creation when interact with rocksdb


wu-hanqing:

zhanghuidinah:
@ccmaxcc Are you having some trouble? If so, please let us know so that we can communicate together.

wuhongsong:
@ccmaxcc

liam-thunder:
Interested in CurveFS, is this PR still available to contribute? I can start with this PR to get familiar with CurveFS.

ilixiaocui:
> Interested in CurveFS, is this PR still available to contribute? I can start with this PR to get familiar with CurveFS.

@liam-thunder You are very welcome to contribute to this PR.

liam-thunder:
address this issue in https://github.com/opencurve/curve/pull/2183

Cyber-SiKu:
@liam-thunder Is this still going on?

liam-thunder:
> @liam-thunder Is this still going on?

Sorry for late response, this issue is blocked by E2E benchmark testing and performance analysis, and requiring some extra efforts. But I'm quite busy recently, will get back on this issue later.

Cyber-SiKu:
@liam-thunder Are you still going on?

liam-thunder:
> @liam-thunder Are you still going on?

Yes, I'm working on deploy a single node cluster to do the benchmark.

ilixiaocui:
> > @liam-thunder Is this still going on?
> 
> Sorry for late response, this issue is blocked by E2E benchmark testing and performance analysis, and requiring some extra efforts. But I'm quite busy recently, will get back on this issue later.


If you have any questions during the benchmarking process, please feel free to ask. You can also communicate in the WeChat group.

PR #2183, opened 2023-01-04
avoid memcpy when using RocksDB slice

liam-thunder:

Issue Number: [#1583](https://github.com/opencurve/curve/issues/1583)

What's Changed:
1. add a new interface to expose raw string buffer
2. remove useless Value() called from iterator
3.  use butil::IOBufAppender to improve append 

### Check List

### DIFF
--- a/curvefs/src/metaserver/metastore.cpp
+++ b/curvefs/src/metaserver/metastore.cpp
@@ -738,11 +738,18 @@ MetaStatusCode MetaStoreImpl::GetOrModifyS3ChunkInfo(
 
 void MetaStoreImpl::PrepareStreamBuffer(butil::IOBuf* buffer,
                                         uint64_t chunkIndex,
-                                        const std::string& value) {
+                                        absl::string_view value) {
     buffer->clear();
+    /*
     buffer->append(std::to_string(chunkIndex));
     buffer->append(":");
     buffer->append(value);
+    */
+    butil::IOBufAppender append_buffer;
+    append_buffer.append(std::to_string(chunkIndex));
+    append_buffer.push_back(':');
+    append_buffer.append(value.data(), value.size());
+    append_buffer.move_to(*buffer);
 }
 
 MetaStatusCode MetaStoreImpl::SendS3ChunkInfoByStream(
@@ -752,8 +759,8 @@ MetaStatusCode MetaStoreImpl::SendS3ChunkInfoByStream(
     Key4S3ChunkInfoList key;
     Converter conv;
     for (iterator->SeekToFirst(); iterator->Valid(); iterator->Next()) {
-        std::string skey = iterator->Key();
-        if (!conv.ParseFromString(skey, &key)) {
+        absl::string_view skey = iterator->Key();
+        if (!conv.ParseFromStringView(skey, &key)) {
             return MetaStatusCode::PARSE_FROM_STRING_FAILED;
         }
 
--- a/curvefs/src/metaserver/storage/iterator.h
+++ b/curvefs/src/metaserver/storage/iterator.h
@@ -27,6 +27,7 @@
 #include <vector>
 #include <memory>
 
+#include "absl/strings/string_view.h"
 #include "curvefs/src/metaserver/storage/common.h"
 
 namespace curvefs {
@@ -46,9 +47,9 @@ class Iterator {
 
     virtual void Next() = 0;
 
-    virtual std::string Key() = 0;
+    virtual absl::string_view Key() = 0;
 
-    virtual std::string Value() = 0;
+    virtual absl::string_view Value() = 0;
 
     virtual const ValueType* RawValue() const { return nullptr; }
 
@@ -92,11 +93,11 @@ class MergeIterator : public Iterator {
         FindCurrent();
     }
 
-    std::string Key() override {
+    absl::string_view Key() override {
         return current_->Key();
     }
 
-    std::string Value() override {
+    absl::string_view Value() override {
         return current_->Value();
     }
 
@@ -147,11 +148,11 @@ class ContainerIterator : public Iterator {
         iter_++;
     }
 
-    std::string Key() override {
+    absl::string_view Key() override {
         return iter_->first;
     }
 
-    std::string Value() override {
+    absl::string_view Value() override {
         return iter_->second;
     }
 
--- a/curvefs/src/metaserver/dentry_storage.cpp
+++ b/curvefs/src/metaserver/dentry_storage.cpp
@@ -379,8 +379,7 @@ MetaStatusCode DentryStorage::List(const Dentry& dentry,
     time.start();
     for (iterator->SeekToFirst(); iterator->Valid(); iterator->Next()) {
         seekTimes++;
-        std::string skey = iterator->Key();
-        std::string svalue = iterator->Value();
+        absl::string_view skey = iterator->Key();
         if (!StringStartWith(skey, sprefix)) {
             break;
         } else if (!iterator->ParseFromValue(&current)) {
--- a/curvefs/src/metaserver/inode_storage.cpp
+++ b/curvefs/src/metaserver/inode_storage.cpp
@@ -209,7 +209,7 @@ bool InodeStorage::GetAllInodeId(std::list<uint64_t>* ids) {
 
     Key4Inode key;
     for (iterator->SeekToFirst(); iterator->Valid(); iterator->Next()) {
-        if (!conv_.ParseFromString(iterator->Key(), &key)) {
+        if (!conv_.ParseFromStringView(iterator->Key(), &key)) {
             return false;
         }
         ids->push_back(key.inodeId);
@@ -372,12 +372,12 @@ MetaStatusCode InodeStorage::DelS3ChunkInfoList(
     }
 
     Key4S3ChunkInfoList key;
-    std::vector<std::string> key2del;
+    std::vector<absl::string_view> key2del;
     for (iterator->SeekToFirst(); iterator->Valid(); iterator->Next()) {
-        std::string skey = iterator->Key();
+        absl::string_view skey = iterator->Key();
         if (!StringStartWith(skey, sprefix)) {
             break;
-        } else if (!conv_.ParseFromString(skey, &key)) {
+        } else if (!conv_.ParseFromStringView(skey, &key)) {
             return MetaStatusCode::PARSE_FROM_STRING_FAILED;
         }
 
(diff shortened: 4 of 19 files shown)


curvefs/src/metaserver/storage/iterator.h
```diff
@@ -50,6 +50,8 @@ class Iterator {
 
     virtual std::string Value() = 0;
 
+    virtual void Value(const char** data, size_t* size) {}
```
wu-hanqing, line 53: Can we just return `absl::string_view` and remove previous `virtual std::string Value() = 0`?

And also, `virtual std::string Key() = 0`

curvefs/src/metaserver/metastore.cpp
```diff
@@ -737,12 +737,14 @@ MetaStatusCode MetaStoreImpl::GetOrModifyS3ChunkInfo(
 }
 
 void MetaStoreImpl::PrepareStreamBuffer(butil::IOBuf* buffer,
-                                        uint64_t chunkIndex,
-                                        const std::string& value) {
+                             uint64_t chunkIndex,
+                             const char* data, size_t data_size) {
```
wu-hanqing, line 741: ditto, if `absl::string_view` works, then we can use it.

wu-hanqing (commented):
Thanks for your contribution. 👍 

BTW, you can reply `cicheck` to trigger CI.

ilixiaocui:
image

Please commit by `git commit -s` to Signed-off

curvefs/src/metaserver/storage/iterator.h
liam-thunder: I suppose returning string_view from RocksDBStorageIterator is ok, but I'm not sure whether returning string_view from Value() is safe for other derived iterators, since their returned string's life time may not be longer than this string_view.

liam-thunder:
cicheck

liam-thunder:
> image
> 
> Please commit by `git commit -s` to Signed-off

I've forced pushed a signed-off commit, but this DCO check doesn't restart by replying 'cicheck'..

ilixiaocui:
> > image
> > Please commit by `git commit -s` to Signed-off
> 
> I've forced pushed a signed-off commit, but this DCO check doesn't restart by replying 'cicheck'..

I saw the corresponding commit message, and now manually set the pass.

liam-thunder:
cicheck

ilixiaocui:
cicheck

curvefs/src/metaserver/metastore.cpp
ilixiaocui: I have the same suggestion: https://github.com/abseil/abseil-cpp/blob/master/absl/strings/string_view.h

```
// A `string_view` provides a lightweight view into the string data provided by
// a `std::string`, double-quoted string literal, character array, or even
// another `string_view`. A `string_view` does *not* own the string to which it
// points, and that data cannot be modified through the view.
```

liam-thunder: > I have the same suggestion: https://github.com/abseil/abseil-cpp/blob/master/absl/strings/string_view.h
> 
> ```
> // A `string_view` provides a lightweight view into the string data provided by
> // a `std::string`, double-quoted string literal, character array, or even
> // another `string_view`. A `string_view` does *not* own the string to which it
> // points, and that data cannot be modified through the view.
> ```

string_view is definitely neater solution here, but I still have the same concern as I replayed above, using string_view as returned value of Value() is not safe for other derived iterators. As stated in string_view's comments:

```
// Due to lifetime issues, a `string_view` is sometimes a poor choice for a
// return value and usually a poor choice for a data member. If you do use a
// `string_view` this way, it is your responsibility to ensure that the object
// pointed to by the `string_view` outlives the `string_view`.
```

For RocksDBStorageIterator, since RocksDB has already guaranteed that its key & value will live as long as the iterator's life time, returning string_view is ok, but I don't see other iterators have such guarantee.

curvefs/src/metaserver/storage/iterator.h
wu-hanqing: I think the only exception is https://github.com/opencurve/curve/blob/9de6b45672d68ebfa7fe43a57b83e9b267f51f59/curvefs/src/metaserver/storage/storage_fstream.h#L237, it returns `std::string` from a temporary.

Maybe we can store the temporary as member variable and returns string_view from it?

liam-thunder: > I think the only exception is
> 
> https://github.com/opencurve/curve/blob/9de6b45672d68ebfa7fe43a57b83e9b267f51f59/curvefs/src/metaserver/storage/storage_fstream.h#L237
> 
> , it returns `std::string` from a temporary.
> Maybe we can store the temporary as member variable and returns string_view from it?

Yes, we could use a member variable to store every string returned from InternalKey to endure its life time. 
But changing Iterator's Key() & Value() return type will involve changes to various upper layer logics, if this is acceptable, I  can start to modify these logics. @wu-hanqing

wu-hanqing: > > I think the only exception is
> > https://github.com/opencurve/curve/blob/9de6b45672d68ebfa7fe43a57b83e9b267f51f59/curvefs/src/metaserver/storage/storage_fstream.h#L237
> > 
> > , it returns `std::string` from a temporary.
> > Maybe we can store the temporary as member variable and returns string_view from it?
> 
> Yes, we could use a member variable to store every string returned from InternalKey to endure its life time. But changing Iterator's Key() & Value() return type will involve changes to various upper layer logics, if this is acceptable, I can start to modify these logics. @wu-hanqing

I think it's acceptable.

liam-thunder:
@wu-hanqing @ilixiaocui 
Updated initial version using string_view, still have two missing pieces:
1. for StorageKey's derived class (Key4Inode, Prefix4AllInode, ...), their ParseFromString implementation uses StringToUl to convert string to unsigned long, but for string_view, I haven't found any solid implementation
https://github.com/opencurve/curve/blob/e0148a2503bc3660093119d5208b809c26ee6200/src/common/string_util.h#L72-L85
https://github.com/opencurve/curve/blob/e0148a2503bc3660093119d5208b809c26ee6200/curvefs/src/metaserver/storage/converter.cpp#L113-L118

3. To avoid memcpy here, SDel's interface needs modification, this will involves changes to BaseStorage and all its derived classes https://github.com/opencurve/curve/blob/1eadb36e8fb986987f1b6603302890acac32a251/curvefs/src/metaserver/inode_storage.cpp#L400-L405

Needs your advices on these two problems, thanks!

liam-thunder:
cicheck

wu-hanqing:
> @wu-hanqing @ilixiaocui Updated initial version using string_view, still have two missing pieces:
> 
> 1. for StorageKey's derived class (Key4Inode, Prefix4AllInode, ...), their ParseFromString implementation uses StringToUl to convert string to unsigned long, but for string_view, I haven't found any solid implementation
>    https://github.com/opencurve/curve/blob/e0148a2503bc3660093119d5208b809c26ee6200/src/common/string_util.h#L72-L85
>    
>    https://github.com/opencurve/curve/blob/e0148a2503bc3660093119d5208b809c26ee6200/curvefs/src/metaserver/storage/converter.cpp#L113-L118
> 2. To avoid memcpy here, SDel's interface needs modification, this will involves changes to BaseStorage and all its derived classes https://github.com/opencurve/curve/blob/1eadb36e8fb986987f1b6603302890acac32a251/curvefs/src/metaserver/inode_storage.cpp#L400-L405
> 
> Needs your advices on these two problems, thanks!

Both problems are tough. 

For the first one, `absl::from_chars` only supports floating-point currently, and `std::from_chars` support integer types but it's C++17 feature.

For the second one, should we change all function parameters from `const std::string & ` to `absl::string_view`? But it seems there is a lot of work.

I suggest we can keep the status quo.

@ilixiaocui What's your point?

ilixiaocui:
> > @wu-hanqing @ilixiaocui Updated initial version using string_view, still have two missing pieces:
> > 
> > 1. for StorageKey's derived class (Key4Inode, Prefix4AllInode, ...), their ParseFromString implementation uses StringToUl to convert string to unsigned long, but for string_view, I haven't found any solid implementation
> >    https://github.com/opencurve/curve/blob/e0148a2503bc3660093119d5208b809c26ee6200/src/common/string_util.h#L72-L85
> >    
> >    https://github.com/opencurve/curve/blob/e0148a2503bc3660093119d5208b809c26ee6200/curvefs/src/metaserver/storage/converter.cpp#L113-L118
> > 2. To avoid memcpy here, SDel's interface needs modification, this will involves changes to BaseStorage and all its derived classes https://github.com/opencurve/curve/blob/1eadb36e8fb986987f1b6603302890acac32a251/curvefs/src/metaserver/inode_storage.cpp#L400-L405
> > 
> > Needs your advices on these two problems, thanks!
> 
> Both problems are tough.
> 
> For the first one, `absl::from_chars` only supports floating-point currently, and `std::from_chars` support integer types but it's C++17 feature.
> 
> For the second one, should we change all function parameters from `const std::string & ` to `absl::string_view`? But it seems there is a lot of work.
> 
> I suggest we can keep the status quo.
> 
> @ilixiaocui What's your point?

Indeed trouble. @liam-thunder Are you interested in testing the delay ratio of this interface on the current calling chain? If this part has a large proportion of delay in the call chain, it is still very meaningful to modify. If it is not the main blocking point on the call chain, first maintain the status quo？

ilixiaocui:
@liam-thunder Hope [this artical](https://mp.weixin.qq.com/s/Q0G-JpXTJocyuuj6HwFwQg) can help you.

The monitoring deployment of curvefs can refer to: https://github.com/opencurve/curveadm/wiki/curvefs-monitor-deployment.

rocksdb is used to store metadata, and the operation frequency is relatively high to add, delete, modify and check inodes and dentry. After mounting the curvefs file system, you can execute `ls`, `mkdir` and other commands, and then check the time-consuming of each processing part involved through monitoring.

wuhongsong:
@liam-thunder Keep on going? or any help？

liam-thunder:
> @liam-thunder Keep on going? or any help？

Sorry for late response, I'm quite busy recently, and also couldn't spare enough test machines to do the performance test. I'll try to get back to this issue later.

liam-thunder:
@ilixiaocui @wu-hanqing @wuhongsong I'm working on this PR again, and try to build a single node CurveFS to do the benchmark test.

Cyber-SiKu:
@liam-thunder Keep on going? or any help？

Issue #1918, opened 2022-09-14
Directly reading the file caches a lot of redundant data


YunhuiChen:
**Describe the bug (描述bug)**
Directly reading the file caches a lot of redundant data

**To Reproduce (复现方法)**
1、open cto and Use multiple mounts
    mount fs1 to  test1
    mount fs1 to test4
2、Run the following script several times，During each run delete the cyh file

#/bin/bash
for i in `seq 1 1000`;do
echo "test $i" >> /home/nbs/failover/test1/cyh
cat /home/nbs/failover/test1/cyh
cat /home/nbs/failover/test4/cyh
#sleep 0.5
echo "test $i" >> /home/nbs/failover/test4/cyh
cat /home/nbs/failover/test1/cyh
cat /home/nbs/failover/test4/cyh
done

3、mount fs1 to test3
![image](https://user-images.githubusercontent.com/67582932/190117100-dc11cdb0-0778-4468-8ff3-a0289ca7381c.png)

![image](https://user-images.githubusercontent.com/67582932/190117159-4c2cc887-f080-4934-91f9-6add82a5d8ad.png)

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

wuhongsong:
there are two reasons:
1. when use S3, a block is the most basic units to split the file. so the context->len we will fill is larger than blockSize in the fuction of ReadFromS3. so even when the s3 obj is less than blockSize, the  context->len is set to blockSize, and we will pass context->len to the WriteReadDirect

2. after  `echo`  every time, it will call close(), so curve-fuse will flush the datacache to the  cachedisk.
   and the most basic units of  `du `  is block, as we known,the default blocksize of local file system is 4K.
   So even if we're writing less than 4K at a time,  We will use at least 4K of disk space

wuhongsong:
@wuhongsong

ilixiaocui:
cc @wuhongsong  Please resubmit the previous repair pr to master support.

Cyber-SiKu:
![image](https://user-images.githubusercontent.com/15775037/211997729-afe1d9ad-cdaa-47c8-be44-3ee16352f889.png)

![image](https://user-images.githubusercontent.com/15775037/211997754-80de4086-b545-440f-871f-e4fad5e86820.png)

PR #2203, opened 2023-01-12
curvefs/client: fix bug that read redundant data(#1918)

wuhongsong:

Issue Number: #1918


wuhongsong:
cicheck

PR #2218, opened 2023-01-30
build: exit immediately if build any failed.

Wine93:


Wine93:
recheck

wuhongsong:
@Wine93   keep on?

Issue #1931, opened 2022-09-22
Improve the performance of warmup with massive small files


wuhongsong:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**
Improve the performance of warmup with massive small files.
Currently, it takes about 3 hours to load 360w small files of about 1k in size with nos-s3. We expect this time to be reduced to about 30 minutes.

**Describe the solution you'd like (描述你期望的解决方法)**

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

wuhongsong:
after the warmup performance  optimize，we can get the expected performance.if we set the iops limit is 3000,  warmup 360W+ files will costs probably 20min
![image](https://user-images.githubusercontent.com/13897133/192180857-7db3f10a-2b4e-497e-88d6-a0c08f850193.png)

wuhongsong:
doing

PR #2236, opened 2023-02-07
curvefs/client:ptimize warmup performance

wuhongsong:

Issue Number: #1931


wuhongsong:
cicheck

wuhongsong:
cicheck

PR #2242, opened 2023-02-09
curvefs: change the data organization format

201341:
Signed-off-by: swj <1186093704@qq.com>

To improve the s3 data storage sync performance

Issue Number: #2380 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/conf/tools.conf
+++ b/curvefs/conf/tools.conf
@@ -35,6 +35,9 @@ s3.bucket_name=bucket
 s3.blocksize=4194304
 s3.chunksize=67108864
 s3.useVirtualAddressing=false
+# s3 objectPrefix, if set 0, means no prefix, if set 1, means inode prefix
+# if set 2 and other values mean hash prefix
+s3.objectPrefix=0
 # statistic info in xattr, hardlink will not be supported when enable
 enableSumInDir=true
 
--- a/curvefs/proto/common.proto
+++ b/curvefs/proto/common.proto
@@ -58,6 +58,7 @@ message S3Info {
     required string bucketname = 4;
     required uint64 blockSize = 5;
     required uint64 chunkSize = 6;
+    optional uint32 objectPrefix = 7;
 }
 
 enum PartitionStatus {
--- a/curvefs/src/client/common/config.h
+++ b/curvefs/src/client/common/config.h
@@ -135,6 +135,7 @@ struct S3ClientAdaptorOption {
     uint32_t baseSleepUs;
     uint32_t maxReadRetryIntervalMs;
     uint32_t readRetryIntervalMs;
+    uint32_t objectPrefix;
     DiskCacheOption diskCacheOpt;
 };
 
--- a/curvefs/src/client/s3/client_s3_adaptor.cpp
+++ b/curvefs/src/client/s3/client_s3_adaptor.cpp
@@ -62,6 +62,7 @@ S3ClientAdaptorImpl::Init(
     chunkFlushThreads_ = option.chunkFlushThreads;
     maxReadRetryIntervalMs_ = option.maxReadRetryIntervalMs;
     readRetryIntervalMs_ = option.readRetryIntervalMs;
+    objectPrefix_ = option.objectPrefix;
     client_ = client;
     inodeManager_ = inodeManager;
     mdsClient_ = mdsClient;
--- a/curvefs/src/client/s3/disk_cache_base.cpp
+++ b/curvefs/src/client/s3/disk_cache_base.cpp
@@ -37,13 +37,13 @@ namespace client {
 #define CACHE_READ_DIR  "cacheread"
 
 void DiskCacheBase::Init(std::shared_ptr<PosixWrapper> wrapper,
-                         const std::string cacheDir) {
+                         const std::string cacheDir, uint32_t objectPrefix) {
     cacheDir_ = cacheDir;
     posixWrapper_ = wrapper;
+    objectPrefix_ = objectPrefix;
 }
 
 int DiskCacheBase::CreateIoDir(bool writreDir) {
-    struct stat statFile;
     bool ret;
     std::string FullDirPath;
 
@@ -85,6 +85,37 @@ std::string DiskCacheBase::GetCacheIoFullDir() {
     return fullPath;
 }
 
+int DiskCacheBase::CreateDir(const std::string dir) {
+    size_t p = dir.find_last_of('/');
+    std::string dirPath = dir;
+    if (p != -1) {
+        dirPath.erase(dirPath.begin()+p, dirPath.end());
+    }
+    std::vector<std::string> names;
+    ::curve::common::SplitString(dirPath, "/", &names);
+    // root dir must exists
+    if (0 == names.size())
+        return 0;
+
+    std::string path;
+    for (size_t i = 0; i < names.size(); ++i) {
+        if (0 == i && dirPath[0] != '/')
+            path = path + names[i];
+        else
+            path = path + "/" + names[i];
+
+        if (IsFileExist(path)) {
+            continue;
+        }
+        // dir needs 755 permission，or “Permission denied”
+        if (posixWrapper_->mkdir(path.c_str(), 0755) < 0) {
+            LOG(WARNING) << "mkdir " << path << " failed. "<< strerror(errno);
+            return -errno;
+        }
+    }
+    return 0;
+}
+
 int DiskCacheBase::LoadAllCacheFile(std::set<std::string> *cachedObj) {
     std::string cachePath = GetCacheIoFullDir();
     bool ret = IsFileExist(cachePath);
@@ -94,26 +125,44 @@ int DiskCacheBase::LoadAllCacheFile(std::set<std::string> *cachedObj) {
     }
 
     VLOG(3) << "LoadAllCacheFile start, dir: " << cachePath;
-    DIR *cacheDir = NULL;
-    struct dirent *cacheDirent = NULL;
-    cacheDir = posixWrapper_->opendir(cachePath.c_str());
-    if (!cacheDir) {
-        LOG(ERROR) << "LoadAllCacheFile, opendir error, errno = " << errno;
-        return -1;
-    }
-    while ((cacheDirent = posixWrapper_->readdir(cacheDir)) != NULL) {
-        if ((!strncmp(cacheDirent->d_name, ".", 1)) ||
-            (!strncmp(cacheDirent->d_name, "..", 2)))
-            continue;
-        std::string fileName = cacheDirent->d_name;
-        cachedObj->emplace(fileName);
-        VLOG(9) << "LoadAllCacheFile obj, name = " << fileName;
-    }
+    std::function<bool(const std::string &path,
+                       std::set<std::string> *cacheObj)> listDir;
 
-    int rc = posixWrapper_->closedir(cacheDir);
-    if (rc < 0) {
-        LOG(ERROR) << "LoadAllCacheFile, opendir error, errno = " << errno;
-        return rc;
+    listDir = [&listDir, this](const std::string &path,
+                               std::set<std::string> *cacheObj) -> bool {
+        DIR *dir;
+        struct dirent *ent;
+        std::string fileName, nextdir;
+        if ((dir = posixWrapper_->opendir(path.c_str())) != NULL) {
+            while ((ent = posixWrapper_->readdir(dir)) != NULL) {
+                VLOG(9) << "LoadAllCacheFile obj, name = " << ent->d_name;
+                if (strncmp(ent->d_name, ".", 1) == 0 ||
+                        strncmp(ent->d_name, "..", 2) == 0) {
+                    continue;
+                } else if (ent->d_type == 8) {
+                    fileName = std::string(ent->d_name);
+                    VLOG(9) << "LoadAllCacheFile obj, name = " << fileName;
+                    cacheObj->emplace(fileName);
+                } else {
+                    nextdir = std::string(ent->d_name);
+                    nextdir = path + '/' + nextdir;
+                    if (!listDir(nextdir, cacheObj)) {
+                        return false;
+                    }
+                }
+            }
+            int ret = posixWrapper_->closedir(dir);
+            if (ret < 0) {
+                LOG(ERROR) << "close dir "  << dir << ", error = " << errno;
+            }
+            return ret >= 0;
+        }
+        LOG(ERROR) << "LoadAllCacheFile Opendir error, path =" << path;
+        return false;
+    };
+    ret = listDir(cachePath, cachedObj);
+    if (!ret) {
+        return -1;
     }
     VLOG(3) << "LoadAllCacheReadFile end, dir: " << cachePath;
     return 0;
--- a/curvefs/src/client/s3/disk_cache_read.cpp
+++ b/curvefs/src/client/s3/disk_cache_read.cpp
@@ -37,9 +37,9 @@ namespace curvefs {
 namespace client {
 
 void DiskCacheRead::Init(std::shared_ptr<PosixWrapper> posixWrapper,
-                         const std::string cacheDir) {
+                         const std::string cacheDir, uint32_t objectPrefix) {
     posixWrapper_ = posixWrapper;
-    DiskCacheBase::Init(posixWrapper, cacheDir);
+    DiskCacheBase::Init(posixWrapper, cacheDir, objectPrefix);
 }
 
 int DiskCacheRead::ReadDiskFile(const std::string name, char *buf,
@@ -85,7 +85,7 @@ int DiskCacheRead::LinkWriteToRead(const std::string fileName,
                                    const std::string fullWriteDir,
                                    const std::string fullReadDir) {
     VLOG(6) << "LinkWriteToRead start. name = " << fileName;
-    std::string fullReadPath, fullWritePath;
+    std::string fullReadPath, fullWritePath, dirPath;
     fullWritePath = fullWriteDir + "/" + fileName;
     fullReadPath = fullReadDir + "/" + fileName;
     int ret;
@@ -94,6 +94,16 @@ int DiskCacheRead::LinkWriteToRead(const std::string fileName,
                    << ", file = " << fullWritePath;
         return -1;
     }
+
+    if (objectPrefix_ != 0) {
+        ret = CreateDir(fullReadPath);
+        if (ret < 0 && errno != EEXIST) {
+            LOG(ERROR) << "Mkdir error. ret = " << ret << ", errno = " << errno
+                       << ", path is " << fullReadPath;
+            return -1;
+        }
+    }
+
     ret = posixWrapper_->link(fullWritePath.c_str(), fullReadPath.c_str());
     if (ret < 0 &&
       errno != EEXIST ) {
--- a/curvefs/src/common/s3util.h
+++ b/curvefs/src/common/s3util.h
@@ -31,13 +31,37 @@ namespace s3util {
 
 inline std::string GenObjName(uint64_t chunkid, uint64_t index,
                               uint64_t compaction, uint64_t fsid,
-                              uint64_t inodeid) {
-    return std::to_string(fsid) + "_" + std::to_string(inodeid) + "_" +
-           std::to_string(chunkid) + "_" + std::to_string(index) + "_" +
-           std::to_string(compaction);
+                              uint64_t inodeid,
+                              uint32_t objectPrefix) {
+    std::string objName;
+    if (objectPrefix == 0) {
+        objName = std::to_string(fsid) + "_" +
+            std::to_string(inodeid) + "_" +
+            std::to_string(chunkid) + "_" + std::to_string(index) + "_" +
+            std::to_string(compaction);
+    } else if (objectPrefix == 1) {
+        objName = std::to_string(fsid) + "/" +
+            std::to_string(inodeid/1000/1000) + "/" +
+            std::to_string(inodeid/1000) + "/" +
+            std::to_string(fsid) + "_" + std::to_string(inodeid) + "_" +
+            std::to_string(chunkid) + "_" + std::to_string(index) + "_" +
+            std::to_string(compaction);
+    } else {
+        objName = std::to_string(fsid) + "/" +
+            std::to_string(inodeid%256) + "/" +
+            std::to_string(inodeid/1000) + "/" +
+            std::to_string(fsid) + "_" + std::to_string(inodeid) + "_" +
+            std::to_string(chunkid) + "_" + std::to_string(index) + "_" +
+            std::to_string(compaction);
+    }
+    return objName;
 }
 
-bool ValidNameOfInode(const std::string &inode, const std::string &objName);
+bool ValidNameOfInode(const std::string &inode, const std::string &objName,
+                      uint32_t objectPrefix);
+
+
+std::string GenPathByObjName(const std::string &objName, uint32_t objectPrefix);
 
 }  // namespace s3util
 }  // namespace common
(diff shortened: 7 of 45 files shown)


201341:
cicheck

wuhongsong:
image

@201341  cpplint is failed, you can   run cpplint locally like `cpplint hello.cpp`

curvefs/src/common/s3util.h
```diff
@@ -32,11 +32,16 @@ namespace s3util {
 inline std::string GenObjName(uint64_t chunkid, uint64_t index,
                               uint64_t compaction, uint64_t fsid,
                               uint64_t inodeid) {
-    return std::to_string(fsid) + "_" + std::to_string(inodeid) + "_" +
+    return std::to_string(fsid) + "/" +
+           std::to_string(inodeid/1000/1000) + "/" +
+           std::to_string(inodeid/1000) + "/" +
+           std::to_string(fsid) + "_" + std::to_string(inodeid) + "_" +
```
wuhongsong, line 38: fsid repeated？

201341:
> image
> 
> @201341 cpplint is failed, you can run cpplint locally like `cpplint hello.cpp`

ok, I will fix it later

curvefs/src/common/s3util.h
```diff
@@ -32,11 +32,16 @@ namespace s3util {
 inline std::string GenObjName(uint64_t chunkid, uint64_t index,
                               uint64_t compaction, uint64_t fsid,
                               uint64_t inodeid) {
-    return std::to_string(fsid) + "_" + std::to_string(inodeid) + "_" +
+    return std::to_string(fsid) + "/" +
```
wuhongsong, line 35: Two suggestions:
1. Maybe we don't need the directory prefix as the object name？
2. maybe we can use hash or % replace ？

curvefs/src/client/s3/disk_cache_read.cpp
```diff
@@ -94,6 +94,20 @@ int DiskCacheRead::LinkWriteToRead(const std::string fileName,
                    << ", file = " << fullWritePath;
         return -1;
     }
+    
+    dirPath = fullReadPath;
+    size_t p = fullReadPath.find_last_of('/');
```
wuhongsong, line 99: can we move the mkdir to DiskCacheBase?

curvefs/src/common/s3util.h
201341: Hi, Adding directory prefix has three advantages.
1:  improve the s3 data storage sync performance when we sync different cloud provider using the tools rclone
2: improve s3 write performance when deploying on aws as the [doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html) suggested
3: improve disk cache write performance.
So I suggerest using the directory prefix.

wuhongsong: > Hi, Adding directory prefix has three advantages. 1: improve the s3 data storage sync performance when we sync different cloud provider using the tools rclone 2: improve s3 write performance when deploying on aws as the [doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html) suggested 3: improve disk cache write performance as writing 1000w file into 1 directory maybe has poor write performance. So I suggerest using the directory prefix.


yeah, you are right. lets do it.  but can we use hash or % replace `inodeid/1000` ? then the files will more evenly distributed,？

201341: > > Hi, Adding directory prefix has three advantages. 1: improve the s3 data storage sync performance when we sync different cloud provider using the tools rclone 2: improve s3 write performance when deploying on aws as the [doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html) suggested 3: improve disk cache write performance as writing 1000w file into 1 directory maybe has poor write performance. So I suggerest using the directory prefix.
> 
> yeah, you are right. lets do it. but can we use hash or % replace `inodeid/1000` ? then the files will more evenly distributed,？

yep, maybe we can make it an optional as [juicefs](https://github.com/juicedata/juicefs/blob/main/pkg/chunk/cached_store.go#L74) does.

wuhongsong: good job, lets do it.

wuhongsong: > > > Hi, Adding directory prefix has three advantages. 1: improve the s3 data storage sync performance when we sync different cloud provider using the tools rclone 2: improve s3 write performance when deploying on aws as the [doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html) suggested 3: improve disk cache write performance as writing 1000w file into 1 directory maybe has poor write performance. So I suggerest using the directory prefix.
> > 
> > 
> > yeah, you are right. lets do it. but can we use hash or % replace `inodeid/1000` ? then the files will more evenly distributed,？
> 
> yep, maybe we can make it an optional as [juicefs](https://github.com/juicedata/juicefs/blob/main/pkg/chunk/cached_store.go#L74) does.

good job, lets do it

curvefs/src/common/s3util.h
```diff
@@ -32,11 +32,16 @@ namespace s3util {
...
+           std::to_string(inodeid/1000/1000) + "/" +
```
ilixiaocui, line 36: Compatibility with existing systems needs to be considered.

curvefs/src/common/s3util.h
```diff
@@ -32,11 +32,16 @@ namespace s3util {
...
+           std::to_string(inodeid/1000) + "/" +
```
ilixiaocui, line 37: We need a comment that exemplifies what each field means.

ilixiaocui: Why do you need inodeid/1000 and inodeid/1000/1000?

wuhongsong: > Why do you need inodeid/1000 and inodeid/1000/1000?
for  multi directory and  s3 prefix，but its better to  replace it with hash or %

curvefs/src/common/s3util.h
201341: There is no better way to handle with compatibility. If changing the data organization, we have to sync the existing file system to new file system.

ilixiaocui:
@201341 What is the current status of this PR?Whether the review opinions have been modified or have any questions？

201341:
@ilixiaocui I'm working on this review opinions. For compatibility with existing systems, I have no good idea, so I don't push this code.

ilixiaocui:
> @ilixiaocui I'm working on this review opinions. For compatibility with existing systems, I have no good idea, so I don't push this code.

For deployed clusters, it should not be realistic to format all existing objects.

One idea is it supports both encodings, which brings a problem. Every time you go to s3 to get it, you have to get it twice according to different shard names.

The second method is to add a flag configuration, the old cluster has always used the old encoding method, and the newly deployed cluster opens the new configuration.In this way, the old cluster cannot be upgraded to the new encoding method.

201341:
> > @ilixiaocui I'm working on this review opinions. For compatibility with existing systems, I have no good idea, so I don't push this code.
> 
> For deployed clusters, it should not be realistic to format all existing objects.
> 
> One idea is it supports both encodings, which brings a problem. Every time you go to s3 to get it, you have to get it twice according to different shard names.
> 
> The second method is to add a flag configuration, the old cluster has always used the old encoding method, and the newly deployed cluster opens the new configuration.In this way, the old cluster cannot be upgraded to the new encoding method.

Second method sounds good, I'll do it.

ilixiaocui:
> > > @ilixiaocui I'm working on this review opinions. For compatibility with existing systems, I have no good idea, so I don't push this code.
> > 
> > 
> > For deployed clusters, it should not be realistic to format all existing objects.
> > One idea is it supports both encodings, which brings a problem. Every time you go to s3 to get it, you have to get it twice according to different shard names.
> > The second method is to add a flag configuration, the old cluster has always used the old encoding method, and the newly deployed cluster opens the new configuration.In this way, the old cluster cannot be upgraded to the new encoding method.
> 
> Second method sounds good, I'll do it.

The second method has an update, and the updated one should be more reasonable.

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
pjdtest failed, As I change the client.conf, the curveadm can't mount the filesystem successfully, I can't handle this

201341:
@wuhongsong @ilixiaocui  would you mind spend some time on reviewing this pr?

ilixiaocui:
> @wuhongsong @ilixiaocui would you mind spend some time on reviewing this pr?

Sorry for making you wait so long. We will review as soon as possible. I will add this issue to the developer event in 2023Q2.

wuhongsong:
sorry, I will review it next week, so busy lately.

wuhongsong:
cicheck

curvefs/src/tools/curvefs_tool_define.cpp
```diff
@@ -249,6 +250,12 @@ std::function<void(curve::common::Configuration*, google::CommandLineFlagInfo*)>
                   std::placeholders::_2, "s3_chunksize", "s3.chunksize",
                   &FLAGS_s3_chunksize);
 
+std::function<void(curve::common::Configuration*, google::CommandLineFlagInfo*)>
+    SetS3_objectprefix =
+        std::bind(&SetDiffFlagInfo<uint32_t>, std::placeholders::_1,
+                  std::placeholders::_2, "s3_objectprefix", "s3.objectprefix",
```
wuhongsong, line 256: s3.objectPrefix needed

wuhongsong:
> pjdtest failed, As I change the client.conf, the curveadm can't mount the filesystem successfully, I can't handle this

I have looked at the reasons for the pjdtest failure. there is one error in s3.objectprefix, s3.objectPrefix needed.

201341:
> > pjdtest failed, As I change the client.conf, the curveadm can't mount the filesystem successfully, I can't handle this
> 
> I have looked at the reasons for the pjdtest failure. there is one error in s3.objectprefix, s3.objectPrefix needed.

yes, s3.objectPrefix is needed, and I have to modify the curveadm's code to support this after this pr is merged just as github can't merge the two repo's pr at the same time.

wuhongsong:
> > pjdtest failed, As I change the client.conf, the curveadm can't mount the filesystem successfully, I can't handle this
> 
> I have looked at the reasons for the pjdtest failure. there is one error in s3.objectprefix, s3.objectPrefix needed.

root@pubbeta2-curve3:/data/logs/curvefs# sudo docker logs 27c3b00eb625
F 2023-04-07T11:34:25.596459+0800 196976 configuration.cpp:318] Get s3.objectprefix from /etc/curvefs/tools.conf fail
*** Check failure stack trace: ***
*** Aborted at 1680838465 (unix time) try "date -d @1680838465" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGABRT (@0x30170) received by PID 196976 (TID 0x7f2102de7240) from PID 196976; stack trace: ***
    @     0x7f21010a00e0 (unknown)
    @     0x7f20ff757fff gsignal
    @     0x7f20ff75942a abort
    @     0x55fcf3844f54 google::FlushAndAbort()
    @     0x55fcf38421a6 google::LogMessage::Fail()
    @     0x55fcf38420ef google::LogMessage::SendToLog()
    @     0x55fcf3841918 google::LogMessage::Flush()
    @     0x55fcf38448e8 google::LogMessageFatal::~LogMessageFatal()
    @     0x55fcf333f9ec curve::common::Configuration::GetValueFatalIfFail()
    @     0x55fcf305fa1b curvefs::tools::SetDiffFlagInfo<>()
    @     0x55fcf3068272 _ZNSt5_BindIFPFvPN5curve6common13ConfigurationEPN6gflags19CommandLineFlagInfoERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESE_PjESt12_PlaceholderILi1EESI_ILi2EEPKcSM_SF_EE6__callIvJOS3_OS6_EJLm0ELm1ELm2ELm3ELm4EEEET_OSt5tupleIJDpT0_EESt12_Index_tupleIJXspT1_EEE
    @     0x55fcf3066b14 std::_Bind<>::operator()<>()
    @     0x55fcf3064c60 std::_Function_handler<>::_M_invoke()
    @     0x55fcf30a3141 std::function<>::operator()()
    @     0x55fcf30a1224 curvefs::tools::CurvefsToolRpc<>::UpdateFlags()
    @     0x55fcf30a072d curvefs::tools::CurvefsToolRpc<>::Init()
    @     0x55fcf315215d curvefs::tools::create::CreateFsTool::Init()
    @     0x55fcf3144b99 curvefs::tools::CurvefsTool::Run()
    @     0x55fcf3057b39 main
    @     0x7f20ff7452e1 __libc_start_main
    @     0x55fcf30577aa _start
    @                0x0 (unknown)
/client.sh: line 9: 196976 Aborted                 (core dumped) $g_curvefs_tool $g_curvefs_tool_operator $g_fsname $g_fstype $g_rpc_timeout_ms
CREATEFS FAILED
root@pubbeta2-curve3:/data/logs/curvefs#

Cyber-SiKu:
> > > pjdtest failed, As I change the client.conf, the curveadm can't mount the filesystem successfully, I can't handle this
> > 
> > 
> > I have looked at the reasons for the pjdtest failure. there is one error in s3.objectprefix, s3.objectPrefix needed.
> 
> yes, s3.objectPrefix is needed, and I have to modify the curveadm's code to support this after this pr is merged just as github can't merge the two repo's pr at the same time.

You should give it an appropriate default value so that it can be created successfully

201341:
cicheck

Cyber-SiKu:
```bash
create fs success.
fileList: []
E 2023-04-10T17:16:31.497514+0800 987453 disk_cache_base.cpp:135] LoadAllCacheFile, cache read dir is not exist.
E 2023-04-10T17:16:31.497532+0800 987453 disk_cache_read.cpp:123] LoadAllCacheReadFile, load all cacched read file fail ret = -1
E 2023-04-10T17:16:31.497556+0800 987453 disk_cache_manager.cpp:104] load all cache read file error. ret = -1
E 2023-04-10T17:16:31.497562+0800 987453 disk_cache_manager_impl.cpp:47] DiskCacheManagerImpl init error.
E 2023-04-10T17:16:31.497567+0800 987453 client_s3_adaptor.cpp:76] Init disk cache failed
E 2023-04-10T17:16:31.497625+0800 987453 main.cpp:231] init fuse client fail, conf = /curvefs/client/conf/client.conf
Begin to mount fs test to /curvefs/client/mnt/var/lib/jenkins/workspace/curve_failover/pjdtest
```

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cichek

201341:
cicheck

curvefs/conf/tools.conf
```diff
@@ -35,6 +35,7 @@ s3.bucket_name=bucket
 s3.blocksize=4194304
 s3.chunksize=67108864
 s3.useVirtualAddressing=false
+s3.objectPrefix=0
```
wuhongsong, line 38: maybe replace 0 with false is better?

curvefs/src/client/common/config.h
```diff
@@ -135,6 +135,7 @@ struct S3ClientAdaptorOption {
     uint32_t baseSleepUs;
     uint32_t maxReadRetryIntervalMs;
     uint32_t readRetryIntervalMs;
+    uint32_t objectPrefix;
```
wuhongsong, line 138: ditto

curvefs/proto/common.proto
```diff
@@ -58,6 +58,7 @@ message S3Info {
     required string bucketname = 4;
     required uint64 blockSize = 5;
     required uint64 chunkSize = 6;
+    optional uint32 objectPrefix = 7;
```
wuhongsong, line 61: ditto, bool is better?

curvefs/src/client/s3/client_s3_adaptor.cpp
```diff
@@ -62,6 +62,7 @@ S3ClientAdaptorImpl::Init(
     chunkFlushThreads_ = option.chunkFlushThreads;
     maxReadRetryIntervalMs_ = option.maxReadRetryIntervalMs;
     readRetryIntervalMs_ = option.readRetryIntervalMs;
+    objectPrefix_ = option.objectPrefix;
```
wuhongsong, line 65: maybe objectPrefix should be a member of FsInfo just like FSType?

curvefs/conf/tools.conf
201341: No, there are 3 kinds of prefixs now, maybe more later. So bool isn't ok. Uint32 or Uint8?

curvefs/src/client/s3/client_s3_adaptor.cpp
201341: objectPrefix is inited in the [FuseopInit](https://github.com/opencurve/curve/blob/master/curvefs/src/client/fuse_s3_client.cpp#L63) process  ` ::curvefs::client::common::S3Info2FsS3Option(s3Info, &fsS3Option);` And s3Info just come from FsInfo.

curvefs/src/client/s3/disk_cache_base.cpp
```diff
@@ -85,6 +85,37 @@ std::string DiskCacheBase::GetCacheIoFullDir() {
...
+    }
+    std::vector<std::string> names;
+    ::curve::common::SplitString(dirPath, "/", &names);
+    // root dir must exists
+    if (0 == names.size())
+        return 0;
+
+    std::string path;
+    for (size_t i = 0; i < names.size(); ++i) {
+        if (0 == i && dirPath[0] != '/')  // 相对路径
```
wuhongsong, line 102: english is needed

curvefs/src/client/s3/disk_cache_read.cpp
```diff
@@ -94,6 +94,15 @@ int DiskCacheRead::LinkWriteToRead(const std::string fileName,
                    << ", file = " << fullWritePath;
         return -1;
     }
+
+    if (objectPrefix_ != 0) {
+        ret = CreateDir(fullReadPath);
+        if (ret < 0 && errno != EEXIST) {
+            LOG(ERROR) << "Mkdir error. ret = " << ret << ", errno = " << errno
+                       << ", path is " << fullReadPath;
```
wuhongsong, line 102: need return

curvefs/conf/tools.conf
201341: ok, I will add some comments here.

ilixiaocui, line 38: It may be more appropriate to use `objectPrefixVersion` instead of `objectPrefix`.
Please add a comment to explain the difference between different versions.

In addition, it is recommended to merge 6 commits into 1 for submission
image

ilixiaocui, line 38: With this design, users can only use the metadata format of the specified version, and cannot upgrade it?

201341: ok

curvefs/src/client/s3/client_s3_adaptor.cpp
ilixiaocui, line 65: The acquisition of fsinfo is through https://github.com/opencurve/curve/blob/1f41ffe8c4f601d77f9aab302d5927ed4d8a6902/curvefs/src/client/curve_fuse_op.cpp#L88

The processing on the mds side is in
https://github.com/opencurve/curve/blob/1f41ffe8c4f601d77f9aab302d5927ed4d8a6902/curvefs/src/mds/mds_service.cpp#L199

mds needs to be assigned according to whether the optional prefixVersion exists.

201341: Hi, objectPrefix is inited in the [FuseopInit](https://github.com/opencurve/curve/blob/master/curvefs/src/client/fuse_s3_client.cpp#L63) process  ::curvefs::client::common::S3Info2FsS3Option(s3Info, &fsS3Option); And s3Info just come from FsInfo.

curvefs/conf/tools.conf
201341: Yes, users cann't change the prefix once the prefix is decided just as the s3.blocksize. If the users want to change the prefix, There are two methods, one is to create another fs, and sync form the previous fs. The other is to sync the s3 object from one bucket to another bucket, While syncing the object, change the object format. Finally,  using the tools to change the bucket and prefix.

201341: done

curvefs/src/client/s3/disk_cache_base.cpp
```diff
@@ -85,6 +85,37 @@ std::string DiskCacheBase::GetCacheIoFullDir() {
...
+            path = path + names[i];
+        else
+            path = path + "/" + names[i];
+
+        if (IsFileExist(path)) {
+            continue;
+        }
+        // 目录需要755权限，不然会出现“Permission denied”
```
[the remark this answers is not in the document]
201341: done

curvefs/src/client/s3/disk_cache_read.cpp
201341: done

curvefs/conf/tools.conf
201341: I think `objectPrefixVersion ` can't express the meaning. 0 means no prefix, And 1/2/3/... means different kinds of prefix.

201341:
cicheck

curvefs/src/client/s3/client_s3_adaptor.cpp
ilixiaocui: > Hi, objectPrefix is inited in the [FuseopInit](https://github.com/opencurve/curve/blob/master/curvefs/src/client/fuse_s3_client.cpp#L63) process ::curvefs::client::common::S3Info2FsS3Option(s3Info, &fsS3Option); And s3Info just come from FsInfo.

FsInfo of FuseopInit come from 
https://github.com/opencurve/curve/blob/1f41ffe8c4f601d77f9aab302d5927ed4d8a6902/curvefs/src/client/curve_fuse_op.cpp#L88

curvefs/conf/tools.conf
ilixiaocui: image

It's not that the prefix is ​​different, but the organizational form of the objectname has changed. Or `objectNameVersion`?

`objectNameVersion=0` means fsid_inodeid_chunkid_index_compaction
`objectNameVersion=1` means fsid_inodeid/1000/1000_inodeid/1000_fsid_inodeid_chunkid_index_compaction

201341: There's no meaning, just adds prefix to improve write and sync performance.

wuhongsong (approved):
LGTM

curvefs/src/client/s3/client_s3_adaptor.cpp
201341: Hi, I think we shouldn't change the mds code, which require upgrading the binary. I can check whether the objectprefix has bee set in  the client, what do you think it is ok ?
`fsS3Opt->objectPrefix = s3.has_objectprefix() ? s3.objectprefix() : 0;`

ilixiaocui: > Hi, I think we shouldn't change the mds code, which require upgrading the binary. I can check whether the objectprefix has bee set in the client, do you think it is ok ? `fsS3Opt->objectPrefix = s3.has_objectprefix() ? s3.objectprefix() : 0;`

LGTM

201341:
cicheck

201341:
cicheck

201341:
cicheck

wuhongsong:
cicheck

Issue #2185, opened 2023-01-05
print error code with more meaningful name


wu-hanqing:
curve has three types of error codes
- protobuf enum type
- C++ enum type, e.g., https://github.com/opencurve/curve/blob/14938624132ba0e88396f6b46b0c2f715e40b3c4/include/client/libcurve_define.h#L30
- raw error value, e.g., https://github.com/opencurve/curve/blob/14938624132ba0e88396f6b46b0c2f715e40b3c4/src/mds/common/mds_define.h#L76

now, in some error path, we only print the raw value of them instead of more meaningful name which result in bad experience.

yamicro:
Hi, I am a newbie in c++ and interested in this issue to try it out. I would like to know if there is some additional information that can help me understand this issue and this project better.

beertocode:
can i be assigned this issue

ilixiaocui:
> can i be assigned this issue

Thank you very much for your interest in this issue, this issue will be assigned to you.

wuhongsong:
@beertocode Are you working on this task? any help?

Cyber-SiKu:
@beertocode Is this still going on?

gueFDF:
Please assign me.

ilixiaocui:
@beertocode Due to no feedback for a long time, the issue is reassigned.

ilixiaocui:
> Please assign me.

Welcome!

gueFDF:
Hello, I encountered an error when submitting pr, but I don't seem to understand what happened. The error message is as follows:
>GitHub pull request #2286 of commit d817215abc36b148ac880d39d0601314707cde2f, no merge conflicts.
Running as SYSTEM
Setting status of d817215abc36b148ac880d39d0601314707cde2f to PENDING with url http://59.111.91.248:8080/job/waiting_in_line/2082/ and message: 'Build started for merge commit.'
[EnvInject] - Loading node environment variables.
Building remotely on [failover-slave3](http://59.111.91.248:8080/computer/failover-slave3) (curve-failover-build3) in workspace /home/nbs/jenkins/workspace/waiting_in_line
[waiting_in_line] $ /bin/bash /tmp/jenkins4459845953071838805.sh
No build required,fail
ERROR: [GitHub Commit Status Setter] - Cannot retrieve Git metadata for the build, setting build result to FAILURE
Build step 'Set build status on GitHub commit [deprecated]' changed build result to FAILURE
[...]

PR #2278, opened 2023-03-02
print error code with more meaningful name

gueFDF:

Issue Number: #2185  

Problem Summary:

What's Changed:
Added some log information, print error code with more meaningful name.

How it Works:
Call the LibCurveErrorName(LIBCURVE_ERROR ret) function to return the specific error information corresponding to the error number.

Basically no side effects.
### Check List


caoxianfei1:
@gueFDF Is this pr is ok?

Issue #2321, opened 2023-03-20
Pass `std::string` by reference instead of value or use `absl::string_view`


Ziy1-Tan:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**
Sometimes we just access `std::string` instead of holding or modify it. We can consider `std::string&` or `absl::string_view` in these case.
example:
https://github.com/opencurve/curve/blob/d0544e2c221e334c2cedc24f1d7e32d25406ad2a/curvefs/src/client/s3/disk_cache_base.h#L47-L53

**Describe the solution you'd like (描述你期望的解决方法)**
See https://blog.hiebl.cc/posts/how-to-use-std-string-view/#string_view-as-function-parameter
https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/
**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

Cyber-SiKu:
> @Cyber-SiKu I'd like to work on it.

:thumbsup:

MizukiCry:
@Ziy1-Tan Hi, should I check all the `std::string` in this project and replace them if possible?

Ziy1-Tan:
Right, You can divide them into several PR to avoid too many changes at once :).

Ziy1-Tan:
> @Ziy1-Tan Hi, should I check all the `std::string` in this project and replace them if possible?

Right, You can divide them into several PR to avoid too many changes at once :).

PR #2326, opened 2023-03-23
[perf] curvefs: improve `std::string` usage in curvefs

MizukiCry:

- Replace the `std::string` by reference or `absl::string_view` when we just access it.

Issue Number: #2321

Problem Summary:

What's Changed:

- Relevant function parameters in curvefs

How it Works:

- Use `const std::string&` or `absl::string_view`

### Check List

- [ ] Relevant documentation/comments are changed or added


MizukiCry:
cicheck

Issue #2094, opened 2022-11-21
[curvefs] When the fs mount is mounted multiple times with different enableSumInDir switches, the xattr information recorded in the metadata will be inaccurate


cw123:
curvefs 有个特性，为了加快getfattr查询目录下的所有文件的length的长度之和，使用了一个enableSumInDir的特性，当开启这个开关的时候，每次在目录下创建、删除文件，修改文件的长度的时候，都会在目录的attr字段中更新相应的字段。可以在curvefs/src/client/fuse_client.cpp 搜索 `enableSumInDir_` 查看对应的实现逻辑。

但是，如果一个fs，挂载了多次，且挂载的时候，有时开启这个特性，有时不开这个特性，就会造成文件的长度信息统计不准。需要想办法避免这种情况。怎么才能在开启这个特性的时候让这个值保持准确？或者，怎么才能在这个统计值不准的时候，不要使用这个值，而是通过遍历所有文件的方式统计数据，来避免拿到错误的数据？

Curvefs has a feature. In order to speed up getfattr to query the sum of the lengths of all files in the directory, a feature of enableSumInDirt is used. When this switch is turned on, every time a file is created or deleted in the directory, the length of the file is modified. At any time, the corresponding field will be updated in the attr field of the directory. You can search `enableSumInDir_` in curvefs/src/client/fuse_client.cpp to view the corresponding implementation logic.

However, if an fs is mounted multiple times, and when mounting, sometimes this feature is enabled, and sometimes this feature is not enabled, the statistics of the length information of the file will be inaccurate. A way needs to be found to avoid this. How can I keep this value accurate when this feature is turned on? Or, how can we not use this value when the statistical value is inaccurate, but get the data by traversing all files to avoid getting wrong data?

```
curvefs/src/client/fuse_client.cpp

enableSumInDir_ = fsInfo_->enablesumindir() && !FLAGS_enableCto;
```

tangwz:
assign

wuhongsong:
Are you having some trouble? If so, please let us know so that we can communicate together    @tangwz

wuhongsong:
@tangwz  keep going when you have time ？ If you're still busy, you can do it later.

ilixiaocui:
cc @tangwz Is this issue still being followed up? Did you encounter any problems？

tangwz:
> cc @tangwz Is this issue still being followed up? Did you encounter any problems？

I'm doing this these days. I'm reading the code and think how to do it.

[...]

PR #2340, opened 2023-03-28
fix_enable

bit-dance:
In multi mount situation,the xattr information will wrong and we can not use the sum to calculate the file bytes.
Issue Number: #2094 

Problem Summary:
When the fs mount is mounted multiple times with different enableSumInDir switches, the xattr information recorded in the metadata will be inaccurate.
If mds find there are multi mounts and then it will close the enableSumInDir forever.
What's Changed:
fuse_client.cpp fs_info_wrapper.cpp
How it Works:
just like above
Maybe we can't use the sum xattr later if we close the enableSumInDir.
### Check List


zhanghuidinah:
cicheck

bit-dance:
cicheck

zhanghuidinah:
cicheck

caoxianfei1:
image
curve_fuse is not found @bit-dance

bit-dance:
Ok,i will research this part later

SeanHai:
cicheck

curvefs/src/mds/fs_info_wrapper.cpp
```diff
@@ -113,6 +113,9 @@ void FsInfoWrapper::AddMountPoint(const Mountpoint& mp) {
     *p = mp;
 
     fsInfo_.set_mountnum(fsInfo_.mountnum() + 1);
+    if (fsInfo_.mountnum() > 1) {
```
SeanHai, line 116: // stats are not cached in xattr in multi-mount scenarios
if (fsInfo_.enablesumindir() && fsInfo_.mountnum() > 1) {
    fsInfo_.set_enablesumindir(false);
}

SeanHai:
@bit-dance 
The current implementation is incomplete. In the multi-mount scenario, the flag in fsInfo is set to false, and subsequent mount points perceive this change (enableSumInDir_ in the client will be a new value), but the mounted client The enableSumInDir_ in is still old, so it is necessary to reversely update the change of this value from mds. At present, the connection from clinet to mds is realized by session. You can consider taking back the flag from mds in the renewal response and updating it to the client.

Cyber-SiKu:
@bit-dance  Will you continue to follow up this PR?

bit-dance:
Yes,these times i am dealing with my exams and later i will continue following codes.

caoxianfei1:
@bit-dance feel free to continue.

bit-dance:
@caoxianfei1 I am so sorry but i will have time to continue next week.

PR #2415, opened 2023-04-20
[fix] build document and sh

Cyber-SiKu:

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/docs/cn/build_and_run.md
+++ b/docs/cn/build_and_run.md
@@ -9,6 +9,7 @@
 3. 以下镜像和编译过程目前仅支持 x86 系统
 4. 如要编译[arm分支](https://github.com/opencurve/curve/pull/2408)，请根据 [Dockerfile](https://github.com/opencurve/curve/blob/master/docker/debian9/compile/Dockerfile)打包编译镜像
 5. 目前master分支不支持在arm系统上编译运行
+6. 推荐 debian 10及以上版本的操作系统
 
 ## 使用Docker进行编译（推荐方式）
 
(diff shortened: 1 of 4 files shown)


Cyber-SiKu:
cicheck

docs/cn/build_and_run.md
```diff
@@ -30,16 +30,23 @@ docker build -t opencurvedocker/curve-base:build-debian9
 ### 在docker镜像中编译
 
 ```bash
-docker run -it opencurvedocker/curve-base:build-debian9 /bin/bash
-cd <workspace>
 git clone https://github.com/opencurve/curve.git 或者 git clone https://gitee.com/mirrors/curve.git
+cd curve
+# 如果你想在容器内完成编译+制作+上传镜像的操作，可以添加以下参数
+# -v /var/run/docker.sock:/var/run/docker.sock -v /root/.docker:/root/.docker
+docker run -v $(pwd):/curve -w /curve -v /root/.cache/bazel:/root/.cache/bazel  --privileged -it opencurvedocker/curve-base:build-debian9 bash
```
ilixiaocui, line 37: 是否可以设置为容器的默认启动参数

Cyber-SiKu: > 是否可以设置为容器的默认启动参数

大部分为映射目录,不能写在dockerfile中, -w是设置工作目录,这个和前面的`-v $(pwd):/curve`相关连,写在dockerfile中也没有必要

docs/cn/build_and_run.md
```diff
@@ -30,16 +30,23 @@ docker build -t opencurvedocker/curve-base:build-debian9
...
 # curve v2.0 之前
 bash mk-tar.sh （编译 curvebs 并打tar包）
 bash mk-deb.sh （编译 curvebs 并打debian包）
+
 # （当前）curve v2.0 及之后
-编译 curvebs: cd curve && make build stor=bs dep=1
-编译 curvefs: cd curve && make build stor=fs dep=1
+# 注意以下方式不会编译单元测试的代码
+# 编译 curvebs:
+make dep stor=bs && make build stor=bs only=src,tools
```
SeanHai, line 47: Does nebd and nbd will compiled with this command?

Cyber-SiKu: > Does nebd and nbd will compiled with this command?

fix

docs/cn/build_and_run.md
```diff
@@ -30,16 +30,26 @@ docker build -t opencurvedocker/curve-base:build-debian9
 ### 在docker镜像中编译
 
 ```bash
-docker run -it opencurvedocker/curve-base:build-debian9 /bin/bash
-cd <workspace>
 git clone https://github.com/opencurve/curve.git 或者 git clone https://gitee.com/mirrors/curve.git
+cd curve
+# 如果你想在容器内完成编译+制作+上传镜像的操作，可以添加以下参数
+# -v /var/run/docker.sock:/var/run/docker.sock -v /root/.docker:/root/.docker
+docker run -v $(pwd):/curve -w /curve -v /root/.cache/bazel:/root/.cache/bazel  --privileged -it opencurvedocker/curve-base:build-debian9 bash
```
charlie0129, line 37:
```suggestion
docker run -it --rm -v $(pwd):/curve -w /curve -v ${HOME} /.cache/bazel:/root/.cache/bazel -u $(id -u):$(id -g) opencurvedocker/curve-base:build-debian9 /bin/bash
```

What about bind-mounting build cache to the home dir of current user instead of root? So that we can possibly share cache.

User is kept the same as the user on host.

`--rm` deletes the container once we are done, instead of creating a new one everytime. There is no need to preserve runtime-data since everything is bind-mounted.

SeanHai (approved):

docs/cn/build_and_run.md
Cyber-SiKu: Is `--rm` necessary? If you are a docker proficient user, you can continue to use this container to complete some work through exec or attach.

charlie0129: Yes, you are right. Without `rm`, the user can start the container and reuse it.

The reason to use `rm` is that it seems this container acts as an ephemeral container, or like Jobs, and doesn't store persistent data. So I think creating a new one is fine in this case. Since `bash` is not a long-running command, so the container will exit once `bash` exits, and we cannot attach to it before it is started manually again.

Cyber-SiKu: `bind-mounting build cache to the home dir of current user` is good idea.

Cyber-SiKu: > Yes, you are right. Without `rm`, the user can start the container and reuse it. That's fine if they know how to do it.
> 
> The reason to use `rm` is that it seems this container acts as an ephemeral container, or like Jobs, and doesn't store persistent data. It is mostly used as a one-time job. So I think creating a new one is fine in this case. Since `bash` is not a long-running command, so the container will exit once `bash` exits, and we cannot attach to it before it is started manually again.

It can also be added to make a prompt. If the user wants to do some work such as testing, a container that is deleted after exiting is not friendly.

charlie0129: That makes sense

ilixiaocui (approved):

Issue #2319, opened 2023-03-17
[optimze] optimize some metric style


fansehep:
In ```/curvefs/src/client/metric/client_metric.h```:

``` c++
struct FSMetric {
    const std::string prefix = "curvefs_client";

...

struct S3Metric {
    const std::string prefix = "curvefs_s3";

...
struct DiskCacheMetric {
    const std::string prefix = "curvefs_disk_cache";

struct KVClientMetric {
    const std::string prefix = "curvefs_kvclient";

...

struct KVClientMetric {
    const std::string prefix = "curvefs_kvclient";

...
```

Use ```const std::string xxx``` in a ```struct``` type is strange. ```static``` is right.

Optimize them in all *.h and *.cpp;

Veeshavanachu:
i will take this

ashutosh887:
@fansehep Let me work on this

caoxianfei1:
@ashutosh887 the work going on?

wu-hanqing:
```cpp
struct FSMetric {
    static constexpr const char* prefix = "curvefs_client";
};
```

maybe this's better, if you use `static const std::string`, you have to define the string in cpp file.

[godbolt](https://godbolt.org/z/6vvh69hqo)

fansehep:
```const char*``` if we have a function like ``` void xxx(const std::string& str_name)```, we input a ```const char*``` here, that will construct a string.

But in fact, static variables are initialized at runtime, and char* can be compile-time constants. It may depend on the scenario.

I think here ```std::string``` is better than ```const char*```.

**Comment by zhanghuidinah** [comment]
- **This issue has been chosen for community developer activities.  Welcome to feel free to try it. After the PR is merged, you will receive a gift package and a certificate as a community contributor.**

- **Developer Activities Registration Steps please ref https://github.com/opencurve/curve/issues/2334#issuecomment-1504723771**

**Comment by Xinlong-Chen** [comment]
The issue seems more simple than other issues, can I get this task?
I agree with the `const static std::string` plan because it is more high-efficiency.
I will add a cpp file to initialize all the string variables.

**Comment by Xinlong-Chen** [comment]
Why does following variables not define const? (such as `const std::string prefix;`，other class in this file do it)
In `/curvefs/src/client/metric/client_metric.h`：
```cpp
struct MDSClientMetric {
    std::string prefix;
}

struct MetaServerClientMetric {
    std::string prefix;
}

struct ClientOpMetric {
    std::string prefix;
}
```

[...]

PR #2417, opened 2023-04-21
[fix] optimize some metric style in curvefs

Xinlong-Chen:

Issue Number: #2319

Problem Summary: Use `static const std::string xxx` to replace `const std::string xxx`  in a struct type.

What's Changed: some metric class in `curvefs/src/client/metric`, add a cpp file to initialize it;

How it Works: static variables only have one copy for all objects, so it occupy less memory in this way.

### Check List

### DIFF
--- a/curvefs/src/client/metric/client_metric.cpp
+++ b/curvefs/src/client/metric/client_metric.cpp
@@ -1,0 +1,43 @@
+/*
+ *  Copyright (c) 2023 NetEase Inc.
+ *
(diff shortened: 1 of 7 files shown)


Xinlong-Chen:
cicheck

Xinlong-Chen:
cicheck

wu-hanqing (approved):

ilixiaocui (approved):

Issue #2225, opened 2023-02-03
curvefs/client: add qos for curve-fuse


wuhongsong:
Description：
add qos for curve-fuse, then  we can limits the bandwidth and  iops

we want to limit the iops and bandwidth of the fuse(read/write) . and we can achieve this by controlling the corresponding fuse interface implementation in curve (like FuseOpWrite)

Proposed solution：
The qos module already exists in the curve project   [here](https://github.com/opencurve/curve/blob/master/src/common/throttle.h)  , and we need to apply it to curve-fuse

YLShiJustFly:
Sorry, Dear. I'm too busy last year, so delay to do this issue. I'd like to try it now again, thanks.

wuhongsong:
> Sorry, Dear. I'm too busy last year, so delay to do this issue. I'd like to try it now again, thanks.

yeah, take it easy. lets do it

zhanghuidinah:
- **This issue has been chosen for community developer activities.  Welcome to feel free to try it. After the PR is merged, you will receive a gift package and a certificate as a community contributor.**

- **Developer Activities Registration Steps please ref https://github.com/opencurve/curve/issues/2334#issuecomment-1504723771**

UniverseParticle:
I will try to do this .Please assign me.

PR #2424, opened 2023-04-24
curvefs/client: add qos for curve-fuse,  we can limits the bandwidth

UniverseParticle:

Issue Number: #2225 

Problem Summary: add qos for curve-fuse, then we can limits the bandwidth or IOPS

What's Changed: 
1.  Read the configuration from the  client.conf
2.  enable the throttle

It does not start by default
### Check List

### DIFF
--- a/curvefs/conf/client.conf
+++ b/curvefs/conf/client.conf
@@ -114,6 +114,36 @@ fuseClient.maxDataSize=1024
 fuseClient.refreshDataIntervalSec=30
 fuseClient.warmupThreadsNum=10
 
+
+# the write throttle bps of fuseClient, default no limit
+fuseClient.throttle.avgWriteBytes=0
+# the write burst bps of fuseClient, default no limit
+fuseClient.throttle.burstWriteBytes=0
+# the times that write burst bps can continue, default 180s
+fuseClient.throttle.burstWriteBytesSecs=180
+
+# the write throttle iops of fuseClient, default no limit
+fuseClient.throttle.avgWriteIops=0
+# the write burst iops of fuseClient, default no limit
+fuseClient.throttle.burstWriteIops=0
+# the times that write burst Iops can continue, default 180s
+fuseClient.throttle.burstWriteIopsSecs=180
+
+# the read throttle bps of fuseClient, default no limit
+fuseClient.throttle.avgReadBytes=0
+# the read burst bps of fuseClient, default no limit
+fuseClient.throttle.burstReadBytes=0
+# the times that read burst bps can continue, default 180s
+fuseClient.throttle.burstReadBytesSecs=180
+
+# the read throttle iops of fuseClient, default no limit
+fuseClient.throttle.avgReadIops=0
+# the read burst Iops of fuseClient, default no limit
+fuseClient.throttle.burstReadIops=0
+# the times that read burst Iops can continue, default 180s
+fuseClient.throttle.burstReadIopsSecs=180
+
+
 #### volume
 volume.bigFileSize=1048576
 volume.volBlockSize=4096
--- a/curvefs/src/client/common/config.cpp
+++ b/curvefs/src/client/common/config.cpp
@@ -51,6 +51,55 @@ DEFINE_bool(useFakeS3, false,
             "Use fake s3 to inject more metadata for testing metaserver");
 DEFINE_bool(supportKVcache, false, "use kvcache to speed up sharing");
 
+/**
+ * use curl -L fuseclient:port/flags/fuseClientAvgWriteBytes?setvalue=true
+ * for dynamic parameter configuration
+ */
+static bool pass_uint64(const char *, uint64_t) { return true; }
+
+DEFINE_uint64(fuseClientAvgWriteBytes, 0,
+              "the write throttle bps of fuse client");
+DEFINE_validator(fuseClientAvgWriteBytes, &pass_uint64);
+DEFINE_uint64(fuseClientBurstWriteBytes, 0,
+              "the write burst bps of fuse client");
+DEFINE_validator(fuseClientBurstWriteBytes, &pass_uint64);
+DEFINE_uint64(fuseClientBurstWriteBytesSecs, 180,
+              "the times that write burst bps can continue");
+DEFINE_validator(fuseClientBurstWriteBytesSecs, &pass_uint64);
+
+
+DEFINE_uint64(fuseClientAvgWriteIops, 0,
+              "the write throttle iops of fuse client");
+DEFINE_validator(fuseClientAvgWriteIops, &pass_uint64);
+DEFINE_uint64(fuseClientBurstWriteIops, 0,
+              "the write burst iops of fuse client");
+DEFINE_validator(fuseClientBurstWriteIops, &pass_uint64);
+DEFINE_uint64(fuseClientBurstWriteIopsSecs, 180,
+              "the times that write burst iops can continue");
+DEFINE_validator(fuseClientBurstWriteIopsSecs, &pass_uint64);
+
+
+DEFINE_uint64(fuseClientAvgReadBytes, 0,
+              "the Read throttle bps of fuse client");
+DEFINE_validator(fuseClientAvgReadBytes, &pass_uint64);
+DEFINE_uint64(fuseClientBurstReadBytes, 0,
+              "the Read burst bps of fuse client");
+DEFINE_validator(fuseClientBurstReadBytes, &pass_uint64);
+DEFINE_uint64(fuseClientBurstReadBytesSecs, 180,
+              "the times that Read burst bps can continue");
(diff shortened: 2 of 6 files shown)


UniverseParticle:
cicheck

wuhongsong:
cicheck

wuhongsong:
@UniverseParticle `To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).`

wuhongsong:
good job

curvefs/src/client/fuse_client.cpp
```diff
@@ -143,6 +142,15 @@ CURVEFS_ERROR FuseClient::Init(const FuseClientOption &option) {
         warmupManager_->SetFsInfo(fsInfo_);
     }
 
+    ReadWriteThrottleParams params;
```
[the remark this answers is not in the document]
UniverseParticle: OK, It's better this way

UniverseParticle:
cicheck

UniverseParticle:
cicheck

ilixiaocui:
image
It is better to merge four commits into one commit.

ilixiaocui:
image

Please use git commit -s -m "****"

wuhongsong:
good job. LGTM.  no any more questions, except `git commit -s`  as said above.

UniverseParticle:
cicheck

UniverseParticle:
cicheck

UniverseParticle:
cicheck

wuhongsong:
`client branch coverage is not ok, actual branch coverage ratio: 42.2714, expected branch coverage ratio: 59.  `  maybe the unitest is abort, you can run unittest in your test environment and look at the reasons for the failure

wuhongsong:
just miss package. run "make stor=fs dep"



---- Replied Message ----
| From | ***@***.***> |
| Date | 05/08/2023 21:48 |
| To | ***@***.***> |
| Cc | ***@***.***>、Review ***@***.***> |
| Subject | Re: [opencurve/curve] curvefs/client: add qos for curve-fuse, we can limits the bandwidth (PR #2424) |

Unit test compilation failed

curvefs/conf/client.conf
```diff
@@ -114,6 +114,19 @@ fuseClient.maxDataSize=1024
...
+# the write burst bps of fuseClient, default no limit
+fuseClient.throttle.burstWriteBytes=0
+# the times that write burst bps can continue, default 180s
+fuseClient.throttle.burstSecs=180
+# the write throttle iops of fuseClient, default no limit
+fuseClient.throttle.avgWriteIops=0
+# the read throttle bps of fuseClient, default no limit
+fuseClient.throttle.avgReadBytes=0
+# the read throttle iops of fuseClient, default no limit
+fuseClient.throttle.avgReadIops=0
```
aspirer, line 128: burst should contains four types:
  - read
      - bps
      - iops
  - write
      - bps : just this one here, we'd better add anothers.
      - iops

curvefs/src/client/common/config.cpp
```diff
@@ -322,8 +331,8 @@ void SetFuseClientS3Option(FuseClientOption *clientOption,
     clientOption->s3Opt.s3AdaptrOpt.bucketName = fsS3Opt.bucketName;
 }
 
-void S3Info2FsS3Option(const curvefs::common::S3Info& s3,
-                       S3InfoOption* fsS3Opt) {
+void S3Info2FsS3Option(const curvefs::common::S3Info &s3,
+                       S3InfoOption *fsS3Opt) {
```
aspirer, line 335: these changes look like only about code styles, the old and new styles in C++ are all good, we can leave them to old style, so the reviewer can pay more attentions to your added codes.

curvefs/conf/client.conf
aspirer: I means we can fix this by a new pr.

UniverseParticle:
cicheck

UniverseParticle:
微信图片_20230514103146

wuhongsong:
cicheck

wuhongsong:
> 微信图片_20230514103146

just wget failed, recheck

UniverseParticle:
cicheck

wuhongsong:
cicheck

curvefs/src/client/fuse_client.cpp
```diff
@@ -1471,5 +1548,32 @@ FuseClient::SetMountStatus(const struct MountOption *mountOption) {
...
+                                     FLAGS_fuseClientBurstReadIops,
+                                     FLAGS_fuseClientBurstReadIopsSecs);
+
+    params.bpsRead = ThrottleParams(FLAGS_fuseClientAvgReadBytes,
+                                    FLAGS_fuseClientBurstReadBytes,
+                                    FLAGS_fuseClientBurstReadBytesSecs);
+
+    throttle_.UpdateThrottleParams(params);
+
+    int ret = bthread_timer_add(&throttleTimer_, butil::seconds_from_now(1),
```
wuhongsong, line 1571: not  periodic tasks? maybe can reference [timer in curve](https://github.com/opencurve/curve/blob/master/curvefs/src/metaserver/heartbeat.h#L147)

UniverseParticle: This is a timer task that refreshes variables once per second

wuhongsong: > This is a timer task that refreshes variables once per second
yes
6d2cc34d-2b00-4cb5-a46a-f2804b95ccff

wuhongsong (approved):

curvefs/src/client/common/config.h
```diff
@@ -205,6 +205,34 @@ struct FuseClientOption {
     bool disableXattr = false;
     uint32_t downloadMaxRetryTimes;
     uint32_t warmupThreadsNum = 10;
+
+    // the write throttle bps of fuseClient
+    uint64_t avgWriteBytes;
```
ilixiaocui, line 210: ditto

curvefs/src/client/common/config.cpp
```diff
@@ -308,6 +308,35 @@ void InitFuseClientOption(Configuration *conf, FuseClientOption *clientOption) {
         clientOption->entryTimeOut = 0;
     }
 
+
+    conf->GetValueFatalIfFail("fuseClient.throttle.avgWriteBytes",
+                              &clientOption->avgWriteBytes);
```
ilixiaocui, line 313: Here you can consider initializing FLAGS_XX directly, without going through the option to switch to one layer.

curvefs/src/client/fuse_client.cpp
```diff
@@ -1471,5 +1548,32 @@ FuseClient::SetMountStatus(const struct MountOption *mountOption) {
     return CURVEFS_ERROR::OK;
 }
 
+void FuseClient::InitQosParam() {
```
ilixiaocui, line 1551: Add new code, please add the corresponding unit test.

ilixiaocui:
image
Please merge into one commit to submit.

curvefs/src/client/common/config.cpp
UniverseParticle: I referred to the qos usage of S3 disk cache。

UniverseParticle: There is a situation where users only need to initialize from the configuration file and do not need to dynamically modify the qos

curvefs/src/client/fuse_client.cpp
UniverseParticle: ok

curvefs/src/client/common/config.cpp
ilixiaocui: If so, the following regular InitQosParam will be problematic.

Even if the user does not set the value of FLAG_XX, the value read from the configuration file will be overwritten.

ilixiaocui: > There is a situation where users only need to initialize from the configuration file and do not need to dynamically modify the qos

It can be considered that FLAG_XX is directly read in the code, and there are two sources of its assignment, one is the configuration file, and the other is the setting of the http interface.

UniverseParticle: The fuse client initialization will read from the configuration file and assign values to FLAG_XX.

UniverseParticle: There are two ways to initializing values . Is this the standard writing method in the future? （initializing FLAGS_XX directly)

UniverseParticle:
cicheck

UniverseParticle:
cicheck

ilixiaocui (approved):

ilixiaocui:
LGTM!

PR #2447, opened 2023-04-27
curvefs: client: remove duplicate code

201341:

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/rpcclient/channel_manager.h
+++ b/curvefs/src/client/rpcclient/channel_manager.h
@@ -72,11 +72,6 @@ ChannelManager<T>::GetOrCreateChannel(const T &id,
     }
 
     curve::common::WriteLockGuard guard(rwlock_);
-    auto iter = channelPool_.find(id);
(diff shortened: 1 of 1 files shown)


201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

curvefs/src/client/rpcclient/channel_manager.h
```diff
@@ -72,11 +72,6 @@ ChannelManager<T>::GetOrCreateChannel(const T &id,
     }
 
     curve::common::WriteLockGuard guard(rwlock_);
-    auto iter = channelPool_.find(id);
```
ilixiaocui, line 75: First add a read lock to judge whether the channel exists; add a write lock, and then judge again, because it is possible that other threads have already added the channel like:
image

PR #2448, opened 2023-04-27
[fix] bug: raft concurrent error

Xinlong-Chen:

Issue Number: #111 

Problem Summary: fix raft concurrent error between `on_leader_start` and `ConcurrentApplyModule`.
before:
image
after:
image

What's Changed:
in `src/chunkserver/copyset_node.cpp`, the order of following code：
```cpp
concurrentapply_->Flush();
leaderTerm_.store(term, std::memory_order_release);
```

How it Works:
Change the race condition `leaderTerm_`, let other request invalid.

### Check List

### DIFF
--- a/src/chunkserver/copyset_node.cpp
+++ b/src/chunkserver/copyset_node.cpp
@@ -495,9 +495,9 @@ int CopysetNode::on_snapshot_load(::braft::SnapshotReader *reader) {
 }
 
 void CopysetNode::on_leader_start(int64_t term) {
-    leaderTerm_.store(term, std::memory_order_release);
     ChunkServerMetric::GetInstance()->IncreaseLeaderCount();
     concurrentapply_->Flush();
+    leaderTerm_.store(term, std::memory_order_release);
     LOG(INFO) << "Copyset: " << GroupIdString()
               << ", peer id: " << peerId_.to_string()
               << " become leader, term is: " << leaderTerm_;


Xinlong-Chen:
cicheck

Xinlong-Chen:
cicheck

ilixiaocui:
The read interface decides whether to follow the consistency protocol to read according to the `applyIndex`.

When raft processes a write request, it first writes the log, and then the state apply log. At this time, the request of `on_**` will be called，Before `on_leader_start` finish, new logs will not be processed by raft.

xu-chaojie (approved):

Xinlong-Chen:
> The read interface decides whether to follow the consistency protocol to read according to the `applyIndex`.
> 
> When raft processes a write request, it first writes the log, and then the state apply log. At this time, the request of `on_**` will be called，Before `on_leader_start` finish, new logs will not be processed by raft.

if we set leaderterm first as following:
https://github.com/opencurve/curve/blob/de447961be61ce546ee2b417964792e8a9d1dc2d/src/chunkserver/copyset_node.cpp#L497-L504
when we process read request after `leaderTerm_.store(term, std::memory_order_release); `, previous code will return true in here:
https://github.com/opencurve/curve/blob/de447961be61ce546ee2b417964792e8a9d1dc2d/src/chunkserver/op_request.cpp#L244-L247
`IsLeaderTerm` implement:
https://github.com/opencurve/curve/blob/de447961be61ce546ee2b417964792e8a9d1dc2d/src/chunkserver/copyset_node.cpp#L652-L656
and push this task to ConcurrentApplyModule:
https://github.com/opencurve/curve/blob/de447961be61ce546ee2b417964792e8a9d1dc2d/src/chunkserver/op_request.cpp#L283-L288
then flush it:
```
concurrentapply_->Flush(); 
```
we will get a wrong answer which is difference with our design.
> “在on_leader_start接口实现中，调用ConcurrentApplyModule的flush接口，保证leader切换时，之前leader上已经apply的数据全部落盘；在on_leader_start调用成功之前，不允许读请求处理。”

ilixiaocui:
> nd push this task to ConcurrentApplyModule:

but `applyIndex` is update after write:

https://github.com/opencurve/curve/blob/de447961be61ce546ee2b417964792e8a9d1dc2d/src/chunkserver/op_request.cpp#LL479C5-L479C5

if `readApplyIndex > applyIndex`， read will proccess by raft:
https://github.com/opencurve/curve/blob/de447961be61ce546ee2b417964792e8a9d1dc2d/src/chunkserver/op_request.cpp#L254

xu-chaojie (commented):
become leader immediately，then leave the request propose to raft node bill be better.

xu-chaojie (changes requested):
become leader immediately，then leave the request propose to raft node bill be better.

Xinlong-Chen:
@ilixiaocui 
this is a example:
> 1. 原先的Leader是A，然后客户端先发起了两个写请求W6 W7都在A上完成了。
> 2. 客户端准备去读 R6，这个请求会带上appliedindex
> 3. 如果这个时候A突然宕机了，那这个RPC请求就失败了。
> 4. 失败之后，会到新的Leader B上继续重试这次R6，不过重试的时候还是带了之前的appliedindex的。

and also can see the [reference](https://zhuanlan.zhihu.com/p/454942497).

Xinlong-Chen:
> become leader immediately，then leave the request propose to raft node bill be better.

According to forementioned plan, we need change client that makes some restart read request don't have appliedindex.
It seems that it is harder than this pr's plan(change the code order).
I wonder to know why it is necessary for us to do the above plan(leave the request propose to raft).

ilixiaocui (approved):

ilixiaocui (approved):

ilixiaocui:
LGTM!

aspirer:
@Xinlong-Chen we'd better add some comments then other developers can get why we do this change, then they will care about the race condition issue when they chage this function.

xu-chaojie (approved):

Xinlong-Chen:
> @Xinlong-Chen we'd better add some comments then other developers can get why we do this change, then they will care about the race condition issue when they chage this function.

I wiil fix it(make summary && add comment). 
Please wait a moment, do not merge now.

Xinlong-Chen:
> ping @Xinlong-Chen , if you have no time, we can merge this pr now. the comments is not very necessary, I think who change this func later can find this pr, too. just a little trouble.

Sorry, I've been busy with my school project recently. It is still difficult to clearly describe this problem, and I will add comments later. u can merge this pr now. thx!

aspirer:
> 

done, tkx

Xinlong-Chen:
> @Xinlong-Chen we'd better add some comments then other developers can get why we do this change, then they will care about the race condition issue when they chage this function.

had fixed in https://github.com/opencurve/curve/pull/2543

Issue #2427, opened 2023-04-25
fuse client hang


wuhongsong:
**Describe the bug (描述bug)**

if `s3.writeCacheMaxByte` set zero, fuse client will hang because the buf cache is full all the time

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

PR #2492, opened 2023-05-20
curvefs/client: fuse client exit when cache size less than 8MB

Ziy1-Tan:

Issue Number: https://github.com/opencurve/curve/issues/2427

Problem Summary:

What's Changed:

- `FuseS3client` exit when `writeCacheMaxByte <= 8MB`

How it Works:

### Check List

### DIFF
--- a/curvefs/conf/client.conf
+++ b/curvefs/conf/client.conf
@@ -149,6 +149,7 @@ s3.baseSleepUs=500
 s3.threadScheduleInterval=3
 # data cache flush wait time
 s3.cacheFlushIntervalSec=5
+# write cache < 8,388,608 (8MB) is not allowed
 s3.writeCacheMaxByte=838860800
 s3.readCacheMaxByte=209715200
 # file cache read thread num
--- a/curvefs/src/client/fuse_s3_client.cpp
+++ b/curvefs/src/client/fuse_s3_client.cpp
@@ -67,12 +67,20 @@ CURVEFS_ERROR FuseS3Client::Init(const FuseClientOption &option) {
 
     auto s3Client = std::make_shared<S3ClientImpl>();
     s3Client->Init(opt.s3Opt.s3AdaptrOpt);
+
+    const uint64_t writeCacheMaxByte =
+        opt.s3Opt.s3ClientAdaptorOpt.writeCacheMaxByte;
+    if (writeCacheMaxByte < MIN_WRITE_CACHE_SIZE) {
+        LOG(ERROR) << "writeCacheMaxByte is too small"
+                   << ", at least " << MIN_WRITE_CACHE_SIZE << " (8MB)"
+                      ", writeCacheMaxByte = " << writeCacheMaxByte;
+        return CURVEFS_ERROR::CACHETOOSMALL;
+    }
+
     auto fsCacheManager = std::make_shared<FsCacheManager>(
         dynamic_cast<S3ClientAdaptorImpl *>(s3Adaptor_.get()),
-        opt.s3Opt.s3ClientAdaptorOpt.readCacheMaxByte,
-        opt.s3Opt.s3ClientAdaptorOpt.writeCacheMaxByte,
-        opt.s3Opt.s3ClientAdaptorOpt.readCacheThreads,
-        kvClientManager_);
+        opt.s3Opt.s3ClientAdaptorOpt.readCacheMaxByte, writeCacheMaxByte,
+        opt.s3Opt.s3ClientAdaptorOpt.readCacheThreads, kvClientManager_);
     if (opt.s3Opt.s3ClientAdaptorOpt.diskCacheOpt.diskCacheType !=
         DiskCacheType::Disable) {
         auto s3DiskCacheClient = std::make_shared<S3ClientImpl>();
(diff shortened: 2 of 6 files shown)


Ziy1-Tan:
cicheck

Ziy1-Tan:
cicheck

Ziy1-Tan:
cicheck

curvefs/src/client/s3/client_s3_adaptor.cpp
```diff
@@ -113,30 +112,27 @@ int S3ClientAdaptorImpl::Write(uint64_t inodeId, uint64_t offset,
...
+        std::lock_guard<std::mutex> lockGuard(ioMtx_);
         pendingReq_.fetch_add(1, std::memory_order_seq_cst);
         VLOG(6) << "pendingReq_ is: " << pendingReq_;
-        uint64_t pendingReq = pendingReq_.load(std::memory_order_seq_cst);
         fsCacheManager_->DataCacheByteInc(length);
         uint64_t size = fsCacheManager_->GetDataCacheSize();
-        uint64_t maxSize = fsCacheManager_->GetDataCacheMaxSize();
-        if ((size + pendingReq * fuseMaxSize_) >= maxSize) {
+        const uint64_t maxSize = fsCacheManager_->GetDataCacheMaxSize();
+        if (size >= maxSize) {
```
wuhongsong, line 122: pendingReq_ can also be deleted, and i think you maybe use another pr to remove fusemax and pendingReq,it will be clearer(this pr just edit cache size only)

Ziy1-Tan: Good suggestion1! Let me rearrange them.

Ziy1-Tan:
cicheck

wuhongsong (approved):

aspirer:
Hi @Ziy1-Tan , thanks for your commit, it's good.
but we can do it better, my advices are:
1. we should not add a new config item, add too many configs will confuse users.
2. we should know that the memory buffer cache size should never less than fuse kernel io size(default is 128KB, user can change it if the compile the fuse.ko), so the s3.writeCacheMaxByte should be a resonable value
3. so we can add some code to check s3.writeCacheMaxByte is resonable value, such as assert(s3.writeCacheMaxByte >= 8M), if not, we exit and do not start the fuse client. 8M can be defined in cpp file like #define MINBUFSIZE 8M which same as fuse.ko.
4. add a comment for s3.writeCacheMaxByte to declare the minbufsize is 8M in curvefs/conf/client.conf

curvefs/src/client/fuse_s3_client.cpp
```diff
@@ -67,12 +67,14 @@ CURVEFS_ERROR FuseS3Client::Init(const FuseClientOption &option) {
 
     auto s3Client = std::make_shared<S3ClientImpl>();
     s3Client->Init(opt.s3Opt.s3AdaptrOpt);
+    const uint64_t writeCacheMaxByte =
```
aspirer, line 70: check the config value >= 8M or not, if not, exit fuse client.

curvefs/conf/client.conf
```diff
@@ -153,6 +153,8 @@ s3.threadScheduleInterval=3
 s3.cacheFlushIntervalSec=5
 s3.writeCacheMaxByte=838860800
 s3.readCacheMaxByte=209715200
+# enable when write cache is 0
+s3.defaultWriteCacheByte=4194304
```
aspirer, line 157: remove it

curvefs/conf/client.conf
```diff
@@ -153,6 +153,8 @@ s3.threadScheduleInterval=3
 s3.cacheFlushIntervalSec=5
 s3.writeCacheMaxByte=838860800
```
aspirer, line 154: add a comment for this config, tell users they should not config a value less than 8M.

curvefs/src/client/fuse_s3_client.cpp
aspirer: add a `#define MINBUFSIZE 8M` above

aspirer (changes requested):

Ziy1-Tan:
> Hi @Ziy1-Tan , thanks for your commit, it's good. but we can do it better, my advices are:
> 
> 1. we should not add a new config item, add too many configs will confuse users.
> 2. we should know that the memory buffer cache size should never less than fuse kernel io size(default is 128KB, user can change it if the compile the fuse.ko), so the s3.writeCacheMaxByte should be a resonable value
> 3. so we can add some code to check s3.writeCacheMaxByte is resonable value, such as assert(s3.writeCacheMaxByte >= 8M), if not, we exit and do not start the fuse client. 8M can be defined in cpp file like #define MINBUFSIZE 8M which same as fuse.ko.
> 4. add a comment for s3.writeCacheMaxByte to declare the minbufsize is 8M in curvefs/conf/client.conf

Let us make them perfect. Thx!

curvefs/conf/client.conf
Ziy1-Tan: Fixed.

curvefs/conf/client.conf
Ziy1-Tan: Fixed.

curvefs/src/client/fuse_s3_client.cpp
Ziy1-Tan: Fixed.

aspirer (approved):
LGTM！thx~

wuhongsong (approved):

Ziy1-Tan:
cicheck

Ziy1-Tan:
cicheck

PR #2496, opened 2023-05-22
[fix]tool-v2: rewrite bs check copyset

lng2020:

Issue Number:  

Problem Summary: The `curve bs check copyset` command is being revamped for `curve bs check chunkserver`.

What's Changed:
1. The function `CheckCopySetHealth` is renamed to `CheckFsCopySetHealth` in `internal/utils/copyset.go` for better clarity. All relevant references in `curvefs` have been updated accordingly.
2. `copyset.go` has been rewritten in `pkg/cli/command/curvebs/check/copyset`, and the file `gap.go` has been added in the same directory. These changes aim to improve error information for the `check chunkserver` command.
3. Relevant flags in the configuration file have been updated.

How it Works:
Here is the current result for `curve bs check copyset`
```bash
root@598452503af0:~# ./curve bs check copyset --copysetid 1 --logicalpoolid 1
+------------+-----------+--------+---------+
| COPYSETKEY | COPYSETID | POOLID | RESULT  |
+------------+-----------+--------+---------+
| 4294967346 | 1         | 1     | healthy |
+------------+-----------+--------+---------+
```

### Check List


tools-v2/internal/utils/base.go
```diff
@@ -0,0 +1,266 @@
+package cobrautil
```
Cyber-SiKu, line 1: Is it the same as pkg/cli/command/base.go?

Cyber-SiKu: If it is to solve the circular reference, you can create another package called copyset under the internal, so that the circular reference can be avoided?

Cyber-SiKu: It is not elegant to maintain two senses

tools-v2/internal/utils/gap.go
```diff
@@ -110,9 +110,9 @@ func NewCopysetLeaderInfo() *CopysetLeaderInfo {
 // snapshot_status: IDLE
 // replicator_3298534883333@10.219.192.50:6800:0: next_index=4  flying_append_entries_size=0 idle hc=150785 ac=2 ic=0
 // replicator_2203318222854@10.219.192.50:6802:0: next_index=4  flying_append_entries_size=0 idle hc=150785 ac=2 ic=0
-func GetLeaderCopysetGap(addr string, key2LeaderInfo *sync.Map, timeout time.Duration) *cmderror.CmdError {
-	metric := basecmd.NewMetric([]string{addr}, RAFT_STAT_URL, timeout)
-	result, _ := basecmd.QueryMetric(metric)
+func GetLeaderCopysetGap(addr string, timeout time.Duration) (CopysetLeaderInfo, *cmderror.CmdError) {
```
Cyber-SiKu, line 113: Wouldn't it be possible to obtain only one modification at a time in this way, and to query multiple times?

montaguelhz:
@lng2020 I'm working on `curve bs check chunkserver` now, and it depends on your code, maybe we can discuss it.

tools-v2/pkg/cli/command/curvebs/status/copyset/copyset.go
```diff
@@ -135,7 +136,7 @@ func (cCmd *GetCopysetStatusCommand) AddFlags() {
 	config.AddBSPeersConfFlag(cCmd.Cmd)
 }
 
-func GetCopysetStatus(caller *cobra.Command) (*map[string]*copyset.CopysetStatusResponse, *cmderror.CmdError) {
+func GetCopysetStatus(caller *cobra.Command) (map[string]*copyset.CopysetStatusResponse, *cmderror.CmdError) {
```
Cyber-SiKu, line 139: If it returns a value, will it make a copy? If the amount of data is large, it should affect performance more.

Cyber-SiKu:
When I run it on a normal cluster, the command seems to go into an infinite loop.

Cyber-SiKu:
The command seems to get stuck when I check for a copyset that doesn't exist.
```bash
curve bs check copyset --copysetid 2 --logicalpoolid 2
```

caoxianfei1:
@lng2020 still follow up the pr?

lng2020:
This pr improves the information provided in the output of the `check copyset` command. It adds more details and context to help users better understand the copyset status. Although a full rewrite of `check copyset` would require significant effort, this change can serve as a reference for anyone looking to enhance this functionality in the future.

PR #2503, opened 2023-05-29
change cbdclient python binding read/write to bytes instead of str

h0hmj:

Issue Number: #xxx 

Problem Summary:

in python3, str need encode/decode, but what we need is binary data. old code works in python2, but not python3.

What's Changed:

cbdclient read/write accept bytes instead of str

How it Works:

### Check List


h0hmj:
cicheck

h0hmj:
cicheck

wu-hanqing (approved):

ilixiaocui (approved):

PR #2529, opened 2023-06-12
[fix]get copyset in clusetr

Cyber-SiKu:
The judgment condition is wrong. If it is the old way, as long as filterScanning is set, the filter will be started regardless of the value, which is inconsistent with expectations.

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


Cyber-SiKu:
cicheck

Wine93 (approved):

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

wu-hanqing (approved):

Issue #2531, opened 2023-06-13
[CurveBS] CurveBS supports modifying the maximum file length limit through the configuration file


ilixiaocui:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

Now max volume size can be ceated by CurveVS is limited by [maxFileLength_](https://github.com/opencurve/curve/blob/f76ed2d5f2c64af384a7d1b7760dbc7c0e8b1fcc/src/mds/nameserver2/curvefs.cpp#L282) which defined by `FLAGS_minFileLength` :
https://github.com/opencurve/curve/blob/f76ed2d5f2c64af384a7d1b7760dbc7c0e8b1fcc/src/mds/main/main.cpp#L65 

The problem is the current code design, we can neither modify the size dynamically through FLAG, nor configure the size through the configuration file.

**Describe the solution you'd like (描述你期望的解决方法)**

In order to adapt to user scenarios more flexibly, we hope that this value can be dynamically modified by using `FLAGS_maxFileLength`, or the size can be configured through the configuration file.

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

Ziy1-Tan:
Seem that flag has been supported, just need to make it configurable in file?

ilixiaocui:
> Seem that flag has been supported, just need to make it configurable in file?

The flag variable is defined, but it still does not support dynamic modification of the parameter through the flag.

It is best to support both config from configuration file and dynamic modification.

Ziy1-Tan:
> > Seem that flag has been supported, just need to make it configurable in file?
> 
> The flag variable is defined, but it still does not support dynamic modification of the parameter through the flag.
> 
> It is best to support both config from configuration file and dynamic modification.

Got it, What is the dynamic modification? Does it mean to reassign the value at runtime? Are there other configuration items that support dynamic modification?

Ziy1-Tan:
> > > > Seem that flag has been supported, just need to make it configurable in file?
> > > 
> > > 
> > > The flag variable is defined, but it still does not support dynamic modification of the parameter through the flag.
> > > It is best to support both config from configuration file and dynamic modification.
> > 
> > 
> > Got it, What is the dynamic modification? Does it mean to reassign the value at runtime? Are there other configuration items that support dynamic modification?
> 
> Sorry, needed a correction, reading from configuration files is now supported. We only need to support dynamic modification just like:
> 
[...]

PR #2552, opened 2023-06-23
curvebs: max/min file length support reload

Ziy1-Tan:

Issue Number: https://github.com/opencurve/curve/issues/2531

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/src/mds/main/main.cpp
+++ b/src/mds/main/main.cpp
@@ -39,8 +39,15 @@ using ::curve::mds::kTB;
 
 DEFINE_uint64(chunkSize, 16 * kMB, "chunk size");
 DEFINE_uint64(segmentSize, 1 * kGB, "segment size");
+/**
+ * use curl -L mds:port/flags/minFileLength?setvalue=10737418240
+ * for dynamic parameter configuration
+ */
+static bool pass_uint64(const char *, uint64_t) { return true; }
 DEFINE_uint64(minFileLength, 10 * kGB, "min filelength");
+DEFINE_validator(minFileLength, &pass_uint64);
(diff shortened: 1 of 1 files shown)


Ziy1-Tan:
cicheck

Cyber-SiKu:
Is this modification necessary?
1. The same effect should be achieved by modifying the configuration file and then restarting mds
2. Does the judgment of the code directly depend on these two flags for the judgment logic of these two values? If not, this modification will not work.

wuhongsong:
> Is this modification necessary?
> 
> 1. The same effect should be achieved by modifying the configuration file and then restarting mds
> 2. Does the judgment of the code directly depend on these two flags for the judgment logic of these two values? If not, this modification will not work.

@Ziy1-Tan any question？

wuhongsong:
> Is this modification necessary?
> 
> 1. The same effect should be achieved by modifying the configuration file and then restarting mds
> 2. Does the judgment of the code directly depend on these two flags for the judgment logic of these two values? If not, this modification will not work.

yep, its will not work.  @Ziy1-Tan

Cyber-SiKu:
IMO, it need more work. @Ziy1-Tan

Ziy1-Tan:
> IMO, it need more work. @Ziy1-Tan

ok, dynamic reloading seems to be enabled by default when defining the flag. I'm not sure what else needs to be modified.

Cyber-SiKu:
@Ziy1-Tan You need to add some unit tests or manual tests to see if your modification takes effect, that is, whether your modification has been passed to the corresponding value.

caoxianfei1:
@Ziy1-Tan We are temporarily closing this pull request. If you want to continue, you can reopen it.

PR #2562, opened 2023-06-26
[fix] warmup to mem error

Cyber-SiKu:
It turns out that mem and disk are not separated, the storage of mem is asynchronous, and the storage of disk is synchronous, so when it is deleted later, it is generally written. Later, after the separation of the two, an exception occurred in the asynchronous save of mem.

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


Cyber-SiKu:
cicheck

wu-hanqing (approved):

Wine93 (approved):

Issue #2493, opened 2023-05-22
[CurveBS/CurveFS] Change hostName to optional in server


ilixiaocui:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

In the current Curve file storage and block storage topo, the server must be configured with hostname and ip. 
https://github.com/opencurve/curve/blob/6f533ae4013b4b1835fc7559f49dd55341a7d27e/curvefs/proto/topology.proto#L82
https://github.com/opencurve/curve/blob/6f533ae4013b4b1835fc7559f49dd55341a7d27e/curvefs/proto/topology.proto#L93
https://github.com/opencurve/curve/blob/6f533ae4013b4b1835fc7559f49dd55341a7d27e/proto/topology.proto#L93
https://github.com/opencurve/curve/blob/6f533ae4013b4b1835fc7559f49dd55341a7d27e/proto/topology.proto#L224

During the deployment process of external users, we found that many users did not configure hostName on the machine. Considering that the communication of the entire cluster depends on IP, we hope to remove the mandatory dependency of hostname to simplify the configuration dependency of deployment.

**Describe the solution you'd like (描述你期望的解决方法)**

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

201341:
I will try this!

ilixiaocui:
> I will try this!

welcome!

Cyber-SiKu:
@201341 Have you encountered any difficulties?

201341:
> @201341 Have you encountered any difficulties?

I split this issues into two parts - curvebs and curvefs. And the curvefs part pr has been submitted

caoxianfei1:
@201341 I see the curvefs has merged, and you can continue to modifications about curvebs.

201341:
> @201341 I see the curvefs has merged, and you can continue to modifications about curvebs.

ok, I will submit the pr later.

Cyber-SiKu:
@201341 continue?

wuhongsong:
This work is in progress

PR #2574, opened 2023-06-27
[feat]curvefs: change hostname to optional

201341:

Issue Number: #2493  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/proto/topology.proto
+++ b/curvefs/proto/topology.proto
@@ -79,7 +79,7 @@ message ZoneData {
 
 message ServerData {
     required uint32 serverId = 1;
-    required string hostName = 2;
+    optional string hostName = 2;
     required string internalIP = 3;
     required uint32 internalPort = 4;
     required string externalIP = 5;
@@ -90,7 +90,7 @@ message ServerData {
 
 message MetaServerData {
     required uint32 MetaServerId = 1;
-    required string hostName = 2;
+    optional string hostName = 2;
     required string token = 3;
     required string internalIP = 4;
     required uint32 internalPort = 5;
@@ -131,7 +131,7 @@ message MetaServerInfo {
 
 // MetaServer message
 message MetaServerRegistRequest {
-    required string hostName = 1;
+    optional string hostName = 1;
     required string internalIp = 2;
     required uint32 internalPort = 3;
     required string externalIp = 4;
@@ -188,7 +188,7 @@ message ServerInfo {
 }
 
 message ServerRegistRequest {
-    required string hostName = 1;
+    optional string hostName = 1;
     required string internalIp = 2;
     optional uint32 internalPort = 3;
     required string externalIp = 4;
--- a/curvefs/src/metaserver/register.cpp
+++ b/curvefs/src/metaserver/register.cpp
@@ -62,11 +62,10 @@ int Register::RegisterToMDS(MetaServerMetadata *metadata) {
     char hostname[128];
     int ret = gethostname(hostname, sizeof(hostname));
     if (ret == -1) {
-        LOG(ERROR) << "get hostname failed!";
-        return -1;
+        LOG(INFO) << "get hostname failed!";
+    } else {
+        req.set_hostname(hostname);
     }
-
-    req.set_hostname(hostname);
     req.set_internalip(ops_.metaserverInternalIp);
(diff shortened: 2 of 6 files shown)


201341:
cicheck

curvefs/src/metaserver/register.cpp
```diff
@@ -62,11 +62,10 @@ int Register::RegisterToMDS(MetaServerMetadata *metadata) {
     char hostname[128];
     int ret = gethostname(hostname, sizeof(hostname));
     if (ret == -1) {
-        LOG(ERROR) << "get hostname failed!";
-        return -1;
+        LOG(INFO) << "get hostname failed!";
+    } else {
+        req.set_hostname(hostname);
```
SeanHai, line 67: You need change the hostName of MetaServerRegistRequest in topology.proto to optional.

curvefs/src/tools/create/curvefs_create_topology_tool.cpp
```diff
@@ -628,7 +628,9 @@ int CurvefsBuildTopologyTool::CreateServer() {
     for (auto it : serverDatas) {
         ServerRegistRequest request;
         ServerRegistResponse response;
-        request.set_hostname(it.name);
+        if (!it.name.empty()) {
+            request.set_hostname(it.name);
```
SeanHai, line 632: You need change hostName of ServerRegistRequest in topology.topo to optional.

curvefs/src/mds/topology/topology_manager.cpp
```diff
@@ -125,7 +125,12 @@ void TopologyManager::RegistMetaServer(const MetaServerRegistRequest *request,
         }
     }
 
-    MetaServer metaserver(metaServerId, request->hostname(), token, serverId,
+    std::string hostname;
+    if (request->has_hostname()) {
```
SeanHai, line 129: not need, request->hostname() is empty string if hostname not set.

SeanHai:
You should check all code that uses hostname to see if remove hostname affects it.
For example: https://github.com/opencurve/curve/blob/ace571a80e90c12239b9a8ba0dbc0a230090bd3f/curvefs/src/mds/topology/topology.cpp#L1654

curvefs/proto/topology.proto
```diff
@@ -79,7 +79,7 @@ message ZoneData {
 
 message ServerData {
     required uint32 serverId = 1;
-    required string hostName = 2;
+    optional string hostName = 2;
```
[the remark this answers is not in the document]
201341: Yes, I will change the curvebs code in another pr for passing ci.

curvefs/src/metaserver/register.cpp
201341: ok

curvefs/src/tools/create/curvefs_create_topology_tool.cpp
201341: ok

curvefs/src/mds/topology/topology_manager.cpp
201341: ok

201341:
cicheck

SeanHai:
> 1. You should check all code that uses hostname to see if remove hostname affects it.
>    For example: https://github.com/opencurve/curve/blob/ace571a80e90c12239b9a8ba0dbc0a230090bd3f/curvefs/src/mds/topology/topology.cpp#L1654
> 2. You need add some unit test of you modified when not set hostname.


 The modify of hostname is not only affect register, you need check other place used hostname  like above example.

201341:
> > 1. You should check all code that uses hostname to see if remove hostname affects it.
> >    For example: https://github.com/opencurve/curve/blob/ace571a80e90c12239b9a8ba0dbc0a230090bd3f/curvefs/src/mds/topology/topology.cpp#L1654
> > 2. You need add some unit test of you modified when not set hostname.
> 
> The modify of hostname is not only affect register, you need check other place used hostname like above example.

Yes, I find no other codes that effect by hostname after search keyword 'hostname'.

SeanHai (approved):

SeanHai:
LGTM!

ilixiaocui (approved):

Issue #2590, opened 2023-06-29
[curvefs]mds:improve check s3info when create fs


Cyber-SiKu:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

Currently, curvefs will check whether the s3 information is available through headbuck when creating fs (s3 type). However, sometimes some s3 service providers will disable this interface for security reasons, resulting in failure to create fs. Please consider how to optimize the related experience.

**Describe the solution you'd like (描述你期望的解决方法)**

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

201341:
If one bucket can only be  used for one filesystem, we can put the uuid object into bucket. And use headObject  to check the filesystem instead of headbucket?

caoxianfei1:
@Cyber-SiKu

PR #2625, opened 2023-07-14
[feat]: curvefs: replace BucketExist with putobject

201341:

Issue Number: #2590 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/mds/fs_manager.cpp
+++ b/curvefs/src/mds/fs_manager.cpp
@@ -257,6 +257,19 @@ bool FsManager::CheckFsName(const std::string& fsName) {
     return true;
 }
 
+bool FsManager::TestS3(const std::string& fsName) {
+    const Aws::String aws_key(fsName.c_str(), fsName.size());
+    std::string value(1024, 'a');
+    if (0 != s3Adapter_->PutObject(aws_key, value)) {
+        return false;
+    } else {
+        if (0 != s3Adapter_->DeleteObject(aws_key)) {
+            return false;
+        }
+    }
+    return true;
+}
+
 FSStatusCode FsManager::CreateFs(const ::curvefs::mds::CreateFsRequest* request,
                                  FsInfo* fsInfo) {
     const auto& fsName = request->fsname();
@@ -308,7 +321,7 @@ FSStatusCode FsManager::CreateFs(const ::curvefs::mds::CreateFsRequest* request,
         option_.s3AdapterOption.s3Address = s3Info.endpoint();
         option_.s3AdapterOption.bucketName = s3Info.bucketname();
         s3Adapter_->Reinit(option_.s3AdapterOption);
-        if (!s3Adapter_->BucketExist()) {
+        if (!TestS3(fsName)) {
             LOG(ERROR) << "CreateFs " << fsName
                        << " error, s3info is not available!";
             return FSStatusCode::S3_INFO_ERROR;
--- a/curvefs/src/mds/fs_manager.h
+++ b/curvefs/src/mds/fs_manager.h
@@ -248,6 +248,8 @@ class FsManager {
 
     bool FillVolumeInfo(common::Volume* volume);
 
+    bool TestS3(const std::string& fsName);
+
  private:
     std::shared_ptr<FsStorage> fsStorage_;
     std::shared_ptr<SpaceManager> spaceManager_;
--- a/curvefs/test/mds/fs_manager_test.cpp
+++ b/curvefs/test/mds/fs_manager_test.cpp
@@ -262,7 +262,8 @@ class FSManagerTest : public ::testing::Test {
             .WillOnce(DoAll(
                 SetArgPointee<2>(response),
                 Invoke(RpcService<false>{})));
-        EXPECT_CALL(*s3Adapter_, BucketExist()).WillOnce(Return(true));
+        EXPECT_CALL(*s3Adapter_, PutObject(_, _)).WillOnce(Return(0));
+        EXPECT_CALL(*s3Adapter_, DeleteObject(_)).WillOnce(Return(0));
 
         return fsManager_->CreateFs(&s3Req, &s3FsInfo);
     }
@@ -538,7 +539,8 @@ TEST_F(FSManagerTest, test_fail_create_s3_fs_on_failed_root_node_creation) {
         .WillOnce(DoAll(
             SetArgPointee<2>(response),
             Invoke(RpcService<false>{})));
-    EXPECT_CALL(*s3Adapter_, BucketExist()).WillOnce(Return(true));
+    EXPECT_CALL(*s3Adapter_, PutObject(_, _)).WillOnce(Return(0));
+    EXPECT_CALL(*s3Adapter_, DeleteObject(_)).WillOnce(Return(0));
     EXPECT_CALL(*topoManager_, DeletePartition(_))
         .WillOnce(Return(TopoStatusCode::TOPO_OK));
     ret = fsManager_->CreateFs(&s3Req, &s3FsInfo);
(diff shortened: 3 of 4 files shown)


201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

Cyber-SiKu:
I took a look at bs and I have the same problem, please modify it

201341:
> I took a look at bs and I have the same problem, please modify it. @201341

ok.

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
@SeanHai ci pjdtest Exceeded error, what happend?`+ echo 'Exceeded max check time of 600 seconds, exiting with failure'
Exceeded max check time of 600 seconds, exiting with failure`

201341:
cicheck

201341:
cicheck

SeanHai:
cicheck

curvefs/src/mds/fs_manager.cpp
```diff
@@ -308,10 +308,17 @@ FSStatusCode FsManager::CreateFs(const ::curvefs::mds::CreateFsRequest* request,
...
+        if (s3Adapter_->ObjectExist(aws_key)) {
+            LOG(ERROR) << s3Info.bucketname() << " has been mounted, "
+                       << "use another bucket or delete the bucket objects!";
             return FSStatusCode::S3_INFO_ERROR;
+        } else {
+            if (s3Adapter_->PutObject(aws_key, "curvefs") != 0) {
+                LOG(ERROR) << "CreateFs " << fsName
+                       << " error, s3info is not available!";
+               return FSStatusCode::S3_INFO_ERROR;
+            }
```
Cyber-SiKu, line 321: Can I understand this here, upload an s3 object named fsname to the bucket, and if the corresponding object exists, it is considered that fs already exists?
This doesn't feel very reasonable:
1. If fs is deleted, this object must be deleted together. Although there is a delete statement later, but the link is too long, if it fails in the middle, the object will remain.
2. If the cluster is dismantled directly. Then recreate it, obviously at this time the fs with the same name cannot be created
3. Considering the idempotency of creating fs, if multiple requests to create the same fs are sent at one time, only one will succeed.
It is recommended to upload an s3 object directly.

201341: 1: If it fails in step one and two, three, the fs isn't deleted, it's ok. And If the fs status become Deleting in step 3, I will delete the object in the scanfs.
2: Do you mean two clusters will create the same fsname with the same bucket? 
3: multiple requests to create the same fs, I think all of it will succeed

Cyber-SiKu: > 1: If it fails in step one and two, three, the fs isn't deleted, it's ok. And If the fs status become Deleting in step 3, I will delete the object in the scanfs. 2: Do you mean two clusters will create the same fsname with the same bucket? 3: multiple requests to create the same fs, I think all of it will succeed because of name lock

1. On the one hand, it is necessary to confirm that the intermediate creation fails to ensure that fs will not be deleted. In addition, considering backward compatibility, I think this kind of guarantee is unnecessary.

Cyber-SiKu: > 1: If it fails in step one and two, three, the fs isn't deleted, it's ok. And If the fs status become Deleting in step 3, I will delete the object in the scanfs. 2: Do you mean two clusters will create the same fsname with the same bucket? 3: multiple requests to create the same fs, I think all of it will succeed because of name lock

1. On the one hand, it is necessary to confirm that the intermediate creation fails to ensure that fs will not be deleted. In addition, considering backward compatibility, I think this kind of guarantee is unnecessary.
2. In an extreme case, create a cluster first, then directly dismantle the cluster and create a new cluster when the s3 object is uploaded but not deleted. If the s3 is not cleaned up, create an fs with the same name in the new cluster. This situation should fail?
3. I think as long as you can upload an S3 object that fs will never use, this can be deleted or not.

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
@Cyber-SiKu bs part will be in another pr just for passing ci.

Cyber-SiKu (approved):

curvefs/src/mds/fs_manager.cpp
```diff
@@ -308,10 +308,17 @@ FSStatusCode FsManager::CreateFs(const ::curvefs::mds::CreateFsRequest* request,
         option_.s3AdapterOption.s3Address = s3Info.endpoint();
         option_.s3AdapterOption.bucketName = s3Info.bucketname();
         s3Adapter_->Reinit(option_.s3AdapterOption);
-        if (!s3Adapter_->BucketExist()) {
+        const Aws::String aws_key(fsName.c_str(), fsName.size());
```
SeanHai, line 311: 1. Current modifications are not universal, you can add a function named like 'BucketExistCheckPut' and put a random content(maybe time-dependent) in it.
2. This requires two interface calls. Consider whether you can use the head bucket to judge first, and if it fails (all failures or feature error codes), then make a put judgment?

201341: 1: Current modifications doesn't cover the condition that if the bucket doesn't exist, we should create bucket 
automaticlly.
2: I think it's not necessary to call head bucket, When s3 providers disable the interface, we need to call put object, 
When having the permission to call interface head bucket, we need also call put object to make sure the fs can upload object normally.

wuhongsong:
@SeanHai PTAL

Cyber-SiKu:
@SeanHai PTAL

201341:
cicheck

201341:
cicheck

201341:
cicheck

curvefs/src/mds/fs_manager.cpp
201341: fixed

201341:
@SeanHai  PTAL

curvefs/src/mds/fs_manager.cpp
```diff
@@ -257,6 +257,19 @@ bool FsManager::CheckFsName(const std::string& fsName) {
     return true;
 }
 
+bool FsManager::TestS3(std::string fsName) {
```
SeanHai, line 260: `bool FsManager::TestS3(std::string fsName) {` -> `bool FsManager::TestS3(const std::string &fsName) {`

curvefs/src/mds/fs_manager.cpp
```diff
@@ -257,6 +257,19 @@ bool FsManager::CheckFsName(const std::string& fsName) {
...
+    const Aws::String aws_key(fsName.c_str(), fsName.size());
+    std::string value(1024, 'a');
+    if (s3Adapter_->PutObject(aws_key, value)) {
+        return false;
+    } else {
+        if (s3Adapter_->DeleteObject(aws_key)) {
```
SeanHai, line 266: ditto

201341:
cicheck

201341:
cicheck

201341:
cicheck

curvefs/src/mds/fs_manager.cpp
```diff
@@ -257,6 +257,19 @@ bool FsManager::CheckFsName(const std::string& fsName) {
     return true;
 }
 
+bool FsManager::TestS3(std::string fsName) {
+    const Aws::String aws_key(fsName.c_str(), fsName.size());
+    std::string value(1024, 'a');
+    if (s3Adapter_->PutObject(aws_key, value)) {
```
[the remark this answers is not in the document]
201341: done

curvefs/src/mds/fs_manager.cpp
201341: done

curvefs/src/mds/fs_manager.cpp
201341: done

SeanHai (approved):

Issue #2588, opened 2023-06-28
[tools-v2]:add bs recover volume


Cyber-SiKu:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

**Describe the solution you'd like (描述你期望的解决方法)**

https://github.com/opencurve/curve/blob/44949da50bacd046e2143448fb524ca1dece908b/curvefs_python/curvefs_tool.py#L93-L96

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

CrystalAnalyst:
Hi, I'm a junior major in computer science in HangzhouDianzi University currently. I am familiar with programming languages like Golang and C++. (Also, I'm also a newbie to Rust). I'm interested in WebFramework Development, Cloud-Native, Distributed System and Storage System. I'm good at Golang development and familiar with the go stand library. I wish I can take on this issue to pracitce my skills and capability further. Please assign it to me.

PR #2638, opened 2023-07-21
[feat]tools-v2:curve bs recover volume

lng2020:

Issue Number: #2588  

Problem Summary: add `curve bs recover volume` command

What's Changed: ./pkg/cli/command/curvebs/recover, ./pkg/cli/command/curvebs/bs.go and ./pkg/internal/error/error.go

How it Works: send `RecoverFileRequest` to recover volume

### Check List

### DIFF
--- a/tools-v2/internal/error/error.go
+++ b/tools-v2/internal/error/error.go
@@ -880,4 +880,15 @@ var (
 		}
 		return NewRpcReultCmdError(code, message)
 	}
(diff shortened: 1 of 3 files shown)


Cyber-SiKu:
recover volume

Cyber-SiKu:
@lng2020 continue？

Cyber-SiKu:
This should use the sdk of curvebs, which is a hard problem. You can consider using the go plugin (I have completed part of the work and have not submitted it yet) to achieve it. You can consider other problems.

Issue #2585, opened 2023-06-28
[tools-v2]:add bs create copyset snapshot


Cyber-SiKu:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

**Describe the solution you'd like (描述你期望的解决方法)**

https://github.com/opencurve/curve/blob/44949da50bacd046e2143448fb524ca1dece908b/tools/snaptool/snaptool.py#L38-L39

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

montaguelhz:
assign me for a try /doge

ApiaoSamaa:
I'd also like to have a try :) /doge

PR #2663, opened 2023-08-01
[feat]curvefs: add bs create volume snapshot

ApiaoSamaa:

Issue Number: #2585  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/tools-v2/pkg/cli/command/curvebs/create/volume/snapshot/snapshot.go
+++ b/tools-v2/pkg/cli/command/curvebs/create/volume/snapshot/snapshot.go
@@ -1,0 +1,174 @@
+/*
+ *  Copyright (c) 2023 NetEase Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/*
+ * Project: tools-v2
+ * Created Date: 2023-7-24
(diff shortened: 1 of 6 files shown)


tools-v2/pkg/cli/command/curvebs/create/volume/snapshot/snapshot.go
```diff
@@ -0,0 +1,158 @@
+/*
```
Cyber-SiKu, line 1: add License

Cyber-SiKu (commented):
Maybe the problem is not clearly described, what I want to achieve is this:
https://github.com/opencurve/curve/blob/6a7415134548286998c28f486e48e685d397daa0/tools/snaptool/curltool.py#L85-L95

Cyber-SiKu:
> Maybe the problem is not clearly described, what I want to achieve is this:
> 
> https://github.com/opencurve/curve/blob/6a7415134548286998c28f486e48e685d397daa0/tools/snaptool/curltool.py#L85-L95

There should be a problem with your implementation. What you want to achieve is this, which should be to create an http request.

ApiaoSamaa:
Hi! I found the problem, which was caused due to my being misled into calling the wrong interface, and now two solutions come into my mind: 1. The client contracts an http packet. 2. Add a new roc interface. If you want a uniform format, we may use solution 2, but 1 is easier to achieve. @Cyber-SiKu

Cyber-SiKu:
> Hi! I found the problem, which was caused due to my being misled into calling the wrong interface, and now two solutions come into my mind: 1. The client contracts an http packet. 2. Add a new roc interface. If you want a uniform format, we may use solution 2, but 1 is easier to achieve. @Cyber-SiKu

Option 1 is just fine, you can refer to this [pr](https://github.com/opencurve/curve/pull/2713)

Cyber-SiKu:
@ApiaoSamaa Did you encounter any problems?

Cyber-SiKu:
pls fix dco && conflicts

SeanHai:
Do you want to continue this issue?

ApiaoSamaa:
> Do you want to continue this issue?↳

Sorry for my delay... I might quit TuT

wuhongsong:
pr stale,you can reopen it later.

PR #2679, opened 2023-08-07
speed up bazel build

h0hmj:
bazel build full target is too slow.

get_target make all build serial now, if we build all, we should be parallel...


wu-hanqing (approved):

h0hmj:
cicheck

Wine93 (approved):

PR #2719, opened 2023-08-27
update: print the error code with meaningful name

NopeDl:

Issue Number: #2185 

Problem Summary:

What's Changed:

How it Works:

### Check List


wuhongsong:
`cicheck`  will trigger ci

wuhongsong:
cicheck

Ziy1-Tan:
Enum instead of const is preferred.

src/mds/chunkserverclient/chunkserver_client.cpp
```diff
@@ -48,6 +48,32 @@ namespace curve {
 namespace mds {
 namespace chunkserverclient {
 
+void ChunkServerClient::printErrorCode(int errorCode){
```
wu-hanqing, line 51: Hi, you shouldn't print error directly in here.

What we need it a function that can transfer error code to corresponding name like https://github.com/NopeDl/curve/blob/0a8642a53c801fc3f29c4e5ebdd5bb304a1e9301/src/client/client_common.h#L241, and when we encounter an error, we can use that function to print its name.

BTW, there are a number of status codes that need to be modified and adjusted, and you can start by doing a general search and listing them.

PR #2720, opened 2023-08-28
curvebs: change hostname from required to optional

201341:

Issue Number: #2493 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/proto/topology.proto
+++ b/proto/topology.proto
@@ -98,7 +98,7 @@ message ZoneData {
 
 message ServerData {
     required uint32 serverId = 1;
-    required string hostName = 2;
+    optional string hostName = 2;
     required string internalHostIP = 3;
     required uint32 internalPort = 4;
     required string externalHostIP = 5;
@@ -231,7 +231,7 @@ message SetChunkServerStatusResponse {
 //server message
 message ServerInfo {
     required uint32 serverID = 1;
-    required string hostName = 2;
+    optional string hostName = 2;
     required string internalIp = 3;
     required uint32 internalPort = 4;
     required string externalIp = 5;
@@ -244,7 +244,7 @@ message ServerInfo {
 }
 
 message ServerRegistRequest {
-    required string hostName = 1;
+    optional string hostName = 1;
     required string internalIp = 2; // for internal uses(between mds & chunkserver)
     optional uint32 internalPort = 3;
     required string externalIp = 4; // for client use (read & write & snapshot and so on)
--- a/src/common/net_common.h
+++ b/src/common/net_common.h
@@ -27,6 +27,7 @@
 #include <netdb.h>
 #include <netinet/in.h>    // in_addr
 #include <arpa/inet.h>     // inet_pton, inet_ntop
+#include <ifaddrs.h>
 #include <glog/logging.h>
 #include <string>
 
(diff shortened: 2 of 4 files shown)


201341:
cicheck

201341:
cicheck

YunhuiChen:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

YunhuiChen:
cicheck

201341:
cicheck

src/mds/topology/topology.cpp
```diff
@@ -1478,7 +1478,11 @@ std::string TopologyImpl::GetHostNameAndPortById(ChunkServerIdType csId) {
     }
 
     // get hostName of the chunkserver
-    return server.GetHostName() + ":" + std::to_string(cs.GetPort());
+    std::string hostname = server.GetHostName();
+    if (hostname.empty()) {
+        hostname = server.GetInternalHostIp();
+    }
+    return hostname + ":" + std::to_string(cs.GetPort());
```
Ziy1-Tan, line 1485: `absl::StrCat` to avoid temporary strings

wuhongsong:
@Ziy1-Tan  @wu-hanqing  @xu-chaojie  PTAL？

proto/topology.proto
```diff
@@ -98,7 +98,7 @@ message ZoneData {
 
 message ServerData {
     required uint32 serverId = 1;
-    required string hostName = 2;
+    optional string hostName = 2;
```
xu-chaojie, line 101: This requirement is strange. When hostName is not available, you can use ip instead. HostName is just an identifier internally. There is no need to change it to optional. Discuss with the issue author whether we should change it this way.

SeanHai:
Is there any progress on this PR so far?

proto/topology.proto
201341: While deploying the bs,  not setting the hostname will fail. So change it to optional?

xu-chaojie (approved):

wuhongsong:
@wu-hanqing PTAL

Ziy1-Tan (approved):

src/common/net_common.h
```diff
@@ -69,23 +70,20 @@ class NetCommon {
...
-
-        if (t == nullptr) {
-            LOG(INFO) << "get host ip failed!";
-            return false;
-        }
-
-        char* ip = inet_ntoa(*(struct in_addr*)t->h_addr_list[0]);
+        char ip[INET_ADDRSTRLEN];
+        struct in_addr adr = ((struct sockaddr_in *)ifa->ifa_addr)->sin_addr;
+        inet_ntop(AF_INET, &adr, ip, INET_ADDRSTRLEN);
```
wu-hanqing, line 86: I got `127.0.0.1` using this new approach

image

proto/topology.proto
wu-hanqing, line 101: Since you have changed this field into optional, do the corresponding codes need to be modified？

proto/topology.proto
```diff
@@ -231,7 +231,7 @@ message SetChunkServerStatusResponse {
 //server message
 message ServerInfo {
     required uint32 serverID = 1;
-    required string hostName = 2;
+    optional string hostName = 2;
```
wu-hanqing, line 234: ditto

wu-hanqing (changes requested):

src/common/net_common.h
201341: ok，It's a long time, I will test it to check what happened.

proto/topology.proto
201341: I think I have changed all the corresponding function. I will check it again.

PR #2723, opened 2023-08-30
[feat]curvefs/client:support warmup sym link

Cyber-SiKu:

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/warmup/warmup_manager.cpp
+++ b/curvefs/src/client/warmup/warmup_manager.cpp
@@ -22,14 +22,18 @@
 
 #include "curvefs/src/client/warmup/warmup_manager.h"
 
+#include <fmt/format.h>
 #include <glog/logging.h>
 #include <unistd.h>
 
 #include <algorithm>
 #include <atomic>
+#include <cstdint>
 #include <deque>
 #include <list>
 #include <memory>
+#include <string>
+#include <unordered_map>
 #include <utility>
 
 #include "curvefs/src/client/common/common.h"
@@ -37,9 +41,9 @@
 #include "curvefs/src/client/kvclient/kvclient_manager.h"
 #include "curvefs/src/client/s3/client_s3_cache_manager.h"
 #include "curvefs/src/common/s3util.h"
+#include "fuse3/fuse_lowlevel.h"
 #include "src/common/concurrent/concurrent.h"
 #include "src/common/string_util.h"
-
 
 namespace curvefs {
 namespace client {
@@ -49,6 +53,13 @@ using curve::common::WriteLockGuard;
 
 #define WARMUP_CHECKINTERVAL_US (1000 * 1000)
 
+#define ROOT_PATH_NAME "/"
+
+static bool pass_uint32(const char*, uint32_t) { return true; }
+DEFINE_uint32(warmupMaxSymLink, 1 << 2,
+              "The maximum number of times to parse sym link");
+DEFINE_validator(warmupMaxSymLink, &pass_uint32);
+
 bool WarmupManagerS3Impl::AddWarmupFilelist(fuse_ino_t key,
                                             WarmupStorageType type) {
     if (!mounted_.load(std::memory_order_acquire)) {
@@ -57,7 +68,7 @@ bool WarmupManagerS3Impl::AddWarmupFilelist(fuse_ino_t key,
     }
     // add warmup Progress
     if (AddWarmupProcess(key, type)) {
-        VLOG(9) << "add warmup list task:" << key;
+        LOG(INFO) << "add warmup list task:" << key;
         WriteLockGuard lock(warmupFilelistDequeMutex_);
         auto iter = FindWarmupFilelistByKeyLocked(key);
         if (iter == warmupFilelistDeque_.end()) {
@@ -75,7 +86,7 @@ bool WarmupManagerS3Impl::AddWarmupFilelist(fuse_ino_t key,
     return true;
 }
 
-bool WarmupManagerS3Impl::AddWarmupFile(fuse_ino_t key, const std::string &path,
+bool WarmupManagerS3Impl::AddWarmupFile(fuse_ino_t key, const std::string& path,
                                         WarmupStorageType type) {
     if (!mounted_.load(std::memory_order_acquire)) {
         LOG(ERROR) << "not mounted";
@@ -83,7 +94,7 @@ bool WarmupManagerS3Impl::AddWarmupFile(fuse_ino_t key, const std::string &path,
     }
     // add warmup Progress
     if (AddWarmupProcess(key, type)) {
-        VLOG(9) << "add warmup single task:" << key;
+        LOG(INFO) << "add warmup single task:" << key;
         FetchDentryEnqueue(key, path);
     }
     return true;
@@ -95,13 +106,13 @@ void WarmupManagerS3Impl::UnInit() {
         bgFetchThread_.join();
     }
 
-    for (auto &task : inode2FetchDentryPool_) {
+    for (auto& task : inode2FetchDentryPool_) {
         task.second->Stop();
     }
     WriteLockGuard lockDentry(inode2FetchDentryPoolMutex_);
     inode2FetchDentryPool_.clear();
 
-    for (auto &task : inode2FetchS3ObjectsPool_) {
+    for (auto& task : inode2FetchS3ObjectsPool_) {
         task.second->Stop();
     }
     WriteLockGuard lockS3Objects(inode2FetchS3ObjectsPoolMutex_);
@@ -116,7 +127,7 @@ void WarmupManagerS3Impl::UnInit() {
     WarmupManager::UnInit();
 }
 
-void WarmupManagerS3Impl::Init(const FuseClientOption &option) {
+void WarmupManagerS3Impl::Init(const FuseClientOption& option) {
     WarmupManager::Init(option);
     bgFetchStop_.store(false, std::memory_order_release);
     bgFetchThread_ = Thread(&WarmupManagerS3Impl::BackGroundFetch, this);
@@ -134,8 +145,8 @@ void WarmupManagerS3Impl::BackGroundFetch() {
     }
 }
 
-void WarmupManagerS3Impl::GetWarmupList(const WarmupFilelist &filelist,
-                                        std::vector<std::string> *list) {
+void WarmupManagerS3Impl::GetWarmupList(const WarmupFilelist& filelist,
+                                        std::vector<std::string>* list) {
     struct fuse_file_info fi {};
     fi.flags &= ~O_DIRECT;
     size_t rSize = 0;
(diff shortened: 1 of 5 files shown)


Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

curvefs/src/client/warmup/warmup_manager.cpp
```diff
@@ -173,59 +174,35 @@ void WarmupManagerS3Impl::LookPath(fuse_ino_t key, std::string file) {
...
-    bool isRoot = false;
-    if (file == "/") {
-        splitPath.push_back(file);
-        isRoot = true;
-    } else {
-        curve::common::AddSplitStringToResult(file, "/", &splitPath);
+
+    if (!curve::common::StringStartWith(file, "/")) {
+        LOG(ERROR) << fmt::format("{} isn't absolute path", file);
+        file.erase(0, 1);
```
SeanHai, line 180: What if it is not an absolute path? And what is the purpose of `file.erase(0, 1); `

curvefs/src/client/warmup/warmup_manager.cpp
```diff
@@ -265,7 +242,25 @@ void WarmupManagerS3Impl::FetchDentry(fuse_ino_t key, fuse_ino_t ino,
         return;
 
     } else if (FsFileType::TYPE_SYM_LINK == dentry.type()) {
-        // skip links
+        std::string symLink;
+        CURVEFS_ERROR statusCode =
+            fuseOpReadLink_(0, dentry.inodeid(), &symLink);
```
SeanHai, line 247: Maybe nullptr is more readable than 0.

curvefs/src/client/warmup/warmup_manager.h
```diff
@@ -315,6 +325,23 @@ class WarmupManagerS3Impl : public WarmupManager {
...
+     * for example: parent = 1(/a), subPath{b,c}(b/c) then the
+     * return is the inodeid of /a/b
+     *
+     * @param parent
+     * @param subPath
+     * @param ret return value
+     * @return false: subPath isn't belong to inode
+     */
+    bool GetInodeSubPathParent(fuse_ino_t inode,
+                               std::vector<std::string> subPath,
```
wu-hanqing, line 342:
```suggestion
                               const std::vector<std::string>& subPath,
```

curvefs/src/client/warmup/warmup_manager.cpp
```diff
@@ -302,7 +297,26 @@ void WarmupManagerS3Impl::FetchChildDentry(fuse_ino_t key, fuse_ino_t ino) {
             };
             AddFetchDentryTask(key, task);
             VLOG(9) << "FetchChildDentry: " << dentry.inodeid();
-        } else if (FsFileType::TYPE_SYM_LINK == dentry.type()) {  // need todo
+        } else if (FsFileType::TYPE_SYM_LINK == dentry.type()) {
```
SeanHai, line 300: The logic of functions `FetchDentry` and `FetchChildDentry` is very similar, so you can consider abstracting the common parts.

curvefs/src/client/warmup/warmup_manager.cpp
```diff
@@ -719,6 +732,74 @@ void WarmupManager::CollectMetrics(InterfaceMetric *interface, int count,
     interface->latency << (butil::cpuwide_time_us() - start);
 }
 
+bool WarmupManagerS3Impl::GetInodeSubPathParent(
```
SeanHai, line 735: If there are source files and corresponding link files in the preheating directory, will the file be preheated multiple times?

curvefs/src/client/warmup/warmup_manager.cpp
```diff
@@ -719,6 +732,74 @@ void WarmupManager::CollectMetrics(InterfaceMetric *interface, int count,
...
+    fuse_ino_t nextInode;
+    std::string currentPath = subPath.front();
+    if (currentPath == "..") {
+        std::shared_ptr<InodeWrapper> inodeWrapper;
+        CURVEFS_ERROR statusCode = inodeManager_->GetInode(inode, inodeWrapper);
+        if (statusCode != CURVEFS_ERROR::OK) {
+            LOG(ERROR) << "get inode fail, inodeid = " << inode;
+            return false;
+        }
+        curve::common::UniqueLock lck = inodeWrapper->GetUniqueLock();
```
wu-hanqing, line 755: do we support this case, e.g.,

```
fsmount/
  -> A(symlink to ../fsmount/B)
  -> B
```

curvefs/src/client/warmup/warmup_manager.cpp
```diff
@@ -719,6 +732,74 @@ void WarmupManager::CollectMetrics(InterfaceMetric *interface, int count,
...
+                           << ", path = " << currentPath;
+                return false;
+            }
+            std::vector<std::string> splitSymLink;
+            curve::common::AddSplitStringToResult(symLink, "/", &splitSymLink);
+            subPath.insert(subPath.begin() + 1, splitSymLink.begin(),
+                           splitSymLink.end());
+            nextInode = inode;
+        } else {
+            nextInode = dentry.inodeid();
```
wu-hanqing, line 794: It's better to use iterative approach instead of recursion.

Cyber-SiKu:
cicheck

curvefs/src/client/warmup/warmup_manager.cpp
```diff
@@ -57,7 +59,7 @@ bool WarmupManagerS3Impl::AddWarmupFilelist(fuse_ino_t key,
     }
     // add warmup Progress
     if (AddWarmupProcess(key, type)) {
-        VLOG(9) << "add warmup list task:" << key;
+        LOG(INFO) << "add warmup list task:" << key;
```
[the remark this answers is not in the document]
Cyber-SiKu: > Will this log too much?

Only when preheating is added, one will be output, which feels good.

curvefs/src/client/warmup/warmup_manager.cpp
```diff
@@ -83,7 +85,7 @@ bool WarmupManagerS3Impl::AddWarmupFile(fuse_ino_t key, const std::string &path,
     }
     // add warmup Progress
     if (AddWarmupProcess(key, type)) {
-        VLOG(9) << "add warmup single task:" << key;
+        LOG(INFO) << "add warmup single task:" << key;
```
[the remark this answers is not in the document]
Cyber-SiKu: > ditto

ditto

curvefs/src/client/warmup/warmup_manager.cpp
Cyber-SiKu: > Maybe nullptr is more readable than 0.

Since this is the inodeid, wouldn't nullptr cause confusion: this should be a pointer?

Cyber-SiKu: > Maybe nullptr is more readable than 0.

fix

curvefs/src/client/warmup/warmup_manager.cpp
Cyber-SiKu: > What if it is not an absolute path? And what is the purpose of `file.erase(0, 1); `

If it is not an absolute path, the warm-up task will be skipped.
There should be a problem with the latter one, it has been fixed

curvefs/src/client/warmup/warmup_manager.cpp
Cyber-SiKu: > The logic of functions `FetchDentry` and `FetchChildDentry` is very similar, so you can consider abstracting the common parts.

agree with you anymore @wuhongsong

curvefs/src/client/warmup/warmup_manager.cpp
Cyber-SiKu: > If there are source files and corresponding link files in the preheating directory, will the file be preheated multiple times?

This depends on the situation:
If it is preheated to diskcache, although it will still be preheated, it will be skipped because diskcache has the logic to determine whether the cache exists.
However, there is no corresponding logic for kvcache, and it will be warmed up repeatedly.

curvefs/src/client/warmup/warmup_manager.cpp
```diff
@@ -302,7 +297,26 @@ void WarmupManagerS3Impl::FetchChildDentry(fuse_ino_t key, fuse_ino_t ino) {
...
+            std::vector<std::string> splitSymLink;
+            curve::common::AddSplitStringToResult(symLink, "/", &splitSymLink);
+            fuse_ino_t parent;
+            if (!GetInodeSubPathParent(dentry.inodeid(), splitSymLink,
+                                       &parent)) {
+                LOG(ERROR) << "GetInodeSubPathParent fail, path: " << symLink;
+                continue;
+            }
+            std::string lastName = splitSymLink.back();
+            auto task = [this, key, parent, lastName]() {
```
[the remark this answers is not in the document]
Cyber-SiKu: > `task` isn't being used.

fix

curvefs/src/client/warmup/warmup_manager.cpp
Cyber-SiKu: > do we support this case, e.g.,
> 
> ```
> fsmount/
>   -> A(symlink to ../fsmount/B)
>   -> B
> ```

no

curvefs/src/client/warmup/warmup_manager.cpp
Cyber-SiKu: > It's better to use iterative approach instead of recursion.

Here is to analyze the soft link, I don't think there will be too deep level calls. However, this can be modified. if you insist

curvefs/src/client/warmup/warmup_manager.h
Cyber-SiKu: > 

subPath will modify in function

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

SeanHai (approved):

wuhongsong (approved):

PR #2727, opened 2023-08-31
improve partition range management

SeanHai:
The previous logic of allocate inode range to partition is: index of partiton in this fs * range form conf.
For example, there are five partitions in fs and each manage 100 inodes:
p1 0-99
p2 100-199
p3 200-299
p4 300-399
p5 400-499

If you make  this conf value smaller(maybe 80) and upgrade mds:
p1 0-99
p2 100-199
p3 200-299
p4 300-399
p5 400-499

p6: 400-479    new allocated partition will have repetitive parts with previous partition.

Issue Number: #xxx 

Problem Summary:

Now we will change allocation algorithm and record max inodeId of fs.

previous:
截屏2023-08-31 16 25 50

after fix:
截屏2023-08-31 16 35 03

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/mds/topology/topology_item.h
+++ b/curvefs/src/mds/topology/topology_item.h
@@ -55,6 +55,8 @@ struct ClusterInformation {
     std::string clusterId;
     // <fsId, partition index of this fs>
     std::map<uint32_t, uint32_t> partitionIndexs;
+    // <fsId, next inode id>
+    std::unordered_map<uint32_t, uint64_t> fsNextInodeId;
 
     ClusterInformation() = default;
     explicit ClusterInformation(const std::string &clusterId)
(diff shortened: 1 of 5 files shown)


SeanHai:
cicheck

Wine93 (approved):

curvefs/src/mds/topology/topology_item.h
```diff
@@ -55,6 +55,8 @@ struct ClusterInformation {
     std::string clusterId;
     // <fsId, partition index of this fs>
     std::map<uint32_t, uint32_t> partitionIndexs;
+    // <fsId, next inode id>
+    std::unordered_map<uint32_t, uint64_t> fsMaxAllocatedInodeId;
```
wu-hanqing, line 59: How about `fsNextInodeId`? And there are also some related functions that can be modified according to this.

wu-hanqing (commented):
Do you need to add unit tests?

SeanHai:
cicheck

SeanHai:
> Do you need to add unit tests?

add utests

curvefs/src/mds/topology/topology_item.h
SeanHai: fixed

wu-hanqing (approved):

PR #2741, opened 2023-09-12
fix curvefs mds get metaserver

SeanHai:
The mds maybe get the if of retired metaserver when deal with heartbeat from metaserver:
截屏2023-09-12 15 09 54
截屏2023-09-12 15 11 18
截屏2023-09-12 15 19 46

This bug was triggered after the metaserver was added with a new role after changing the disk：
截屏2023-09-12 15 25 31
截屏2023-09-12 15 15 18

Issue Number: #xxx 

Problem Summary:

What's Changed:

How it Works:

### Check List


SeanHai:
cicheck

Wine93 (approved):
LGTM!

Issue #2322, opened 2023-03-21
Build: Support compilation of specified cc_library


ilixiaocui:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

The current [Makefile](https://github.com/opencurve/curve/blob/master/Makefile
) of Curve only supports the compilation of cc_binary. 

```
curve$ make
## list
Usage:
    make list stor=bs/fs
Examples:
    make list stor=bs

## build
Usage:
    make build stor=bs/fs only=TARGET dep=0/1 release=0/1 os=OS
Examples:
    make build stor=bs only=//src/chunkserver:chunkserver
    make build stor=bs only=src/* dep=0
    make build stor=fs only=test/* os=debian9
    make build stor=fs release=1
```

It is expected to support the compilation of cc_library.

**Describe the solution you'd like (描述你期望的解决方法)**

Support like: 

```
make build stor=bs only=//src/chunkserver:chunkserver_datastore
make build stor=fs only=//curvefs/src/metaserver:curvefs_metaserver
```

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

zztaki:
Hi, @ilixiaocui 
Sorry for previous communication email after I found the assignment [合并脚本 mk-tar.sh 和 mk-deb.sh](https://summer-ospp.ac.cn/org/prodetail/232990061?list=org&navpage=org) have been assigned. So I want to work on this issue and have it as my OSPP 2023 program. I will be appreciated if you could choose me to complete this assignment. 😊

Have a nice day!

ilixiaocui:
> Hi, @ilixiaocui Sorry for previous communication email after I found the assignment [合并脚本 mk-tar.sh 和 mk-deb.sh](https://summer-ospp.ac.cn/org/prodetail/232990061?list=org&navpage=org) have been assigned. So I want to work on this issue and have it as my OSPP 2023 program. I will be appreciated if you could choose me to complete this assignment. 😊
> 
> Have a nice day!

OSPP is still in the registration stage. Multiple people can sign up for one project, and one person can also sign up for multiple projects.If you are really interested, you can try to sign up.

If you are interested in the current topic, you also need to register in the summer of open source.

zztaki:
[...]

PR #2755, opened 2023-09-15
Build: Support compilation of specified cc_library

quas-modo:

Issue Number: #2322 

Problem Summary: The current [Makefile](https://github.com/opencurve/curve/blob/master/Makefile) of Curve only supports the compilation of cc_binary.

What's Changed: /utils/build-in-image.sh has been changed

How it Works: If anyone run `make build/ make ci-build` and have the parameter only to sepecify the target as cc_library, then the change will support to compiler the cc_library.n Also, when running `make list`, the change will list cc_library target.

Side effects(Breaking backward compatibility? Performance regression?): None

### Check List

- [] Relevant documentation/comments is changed or added
- [] I acknowledge that all my contributions will be made under the project's license


quas-modo:
cicheck

quas-modo:
cicheck

quas-modo:
cicheck

quas-modo:
cicheck

PR #2788, opened 2023-10-09
delete partition after fail

yanchaomei:

Problem Summary:
When the partition is successfully created on the metaserver, but fails to persist in the topology, we should go to the metaserver to delete the partition

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/mds/heartbeat/topo_updater.cpp
+++ b/curvefs/src/mds/heartbeat/topo_updater.cpp
@@ -222,7 +222,7 @@ void TopoUpdater::UpdatePartitionTopo(
             LOG(WARNING) << "hearbeat report partition which is not in topo"
                          << ", copysetId = " << copySetId
                          << ", partitionId = " << it.GetPartitionId();
-            
+
             const int maxRetries = 3;
             int retries = 0;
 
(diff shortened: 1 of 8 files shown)


curvefs/src/mds/heartbeat/topo_updater.cpp
```diff
@@ -222,6 +222,28 @@ void TopoUpdater::UpdatePartitionTopo(
             LOG(WARNING) << "hearbeat report partition which is not in topo"
                          << ", copysetId = " << copySetId
                          << ", partitionId = " << it.GetPartitionId();
+
```
SeanHai, line 225: It's not a good idea to delete here. If you really want to do it, you'd better check inode and dentry numbers of this partition again and make sure it is empty.
This piece of code logic can directly call TopologyManager::DeletePartition.

yanchaomei: I noticed that the logic in TopologyManager::DeletePartitio first fetches the corresponding partition from topology_. The situation we are facing now is that there is no corresponding partition in topology_.So, I added the DeleteAbnormalPartition function.

Cyber-SiKu:
pls fix `Code Formatter`

curvefs/src/mds/heartbeat/topo_updater.cpp
SeanHai: Failure to create is just one of the cases where the partition exists in metaserver but not in mds, so to be on the safe side, it is best to check whether the partition is empty first.

You can split the original TopologyManager::DeletePartitio into a function, such as TopologyManager::DeletePartitionDirectly, which does not check whether it exists in mds, so TopologyManager::DeletePartitionDirectly can be reused by you and the original code at the same time, and you can There is no need to rewrite the sending logic again.

caoxianfei1:
@yanchaomei Comlete it as soon as possible if have time.

yanchaomei:
> @yanchaomei Comlete it as soon as possible if have time.

get it

Issue #2779, opened 2023-09-27
[curvefs]mds: create fs


Cyber-SiKu:
**Describe the bug (描述bug)**

When using v2.4 client to mount v2.6 mds, a creation fs error will be reported.

Reason for error:
mds will check whether s3info is consistent, but an attribute is added in the middle: objectPrefix. This attribute does not exist in the old proto, so an error will occur when checking the consistency.

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**

201341:
Assign it to me, I will fix it later. @Cyber-SiKu

Cyber-SiKu:
> Assign it to me, I will fix it later. @Cyber-SiKu

done

caoxianfei1:
@Cyber-SiKu

PR #2791, opened 2023-10-11
[fix] curvefs: mds: createfs error

201341:

Issue Number: #2779  

Problem Summary:

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/mds/fs_manager.cpp
+++ b/curvefs/src/mds/fs_manager.cpp
@@ -791,9 +791,7 @@ FSStatusCode FsManager::GetFsInfo(const std::string& fsName, uint32_t fsId,
 }
 
 int FsManager::IsExactlySameOrCreateUnComplete(const std::string& fsName,
-                                               FSType fsType,
-                                               uint64_t blocksize,
-                                               const FsDetail& detail) {
+    FSType fsType, uint64_t blocksize, const FsDetail& detail) {
     FsInfoWrapper existFs;
 
     auto volumeInfoComparator = [](common::Volume lhs, common::Volume rhs) {
@@ -808,16 +806,34 @@ int FsManager::IsExactlySameOrCreateUnComplete(const std::string& fsName,
         return google::protobuf::util::MessageDifferencer::Equals(lhs, rhs);
     };
 
-    auto checkFsInfo = [fsType, volumeInfoComparator](const FsDetail& lhs,
-                                                      const FsDetail& rhs) {
+    auto s3InfoComparator = [](common::S3Info newFs, common::S3Info existFs) {
+        // If the s3info detail stored in mds doesn't have prefix, the new
+        // client which has prefix value bigger than 0 can't mount filesystem.
+        if (newFs.has_objectprefix() && !existFs.has_objectprefix() &&
+            newFs.objectprefix() != 0) {
+            return false;
+        }
+        // If the s3info detail stored in mds has prefix value bigger than 0,
+        // the old client which doesn't have prefix can't mount filesystem.
+        if (existFs.has_objectprefix() && !newFs.has_objectprefix() &&
+            existFs.objectprefix() != 0) {
+            return false;
+        }
+
+        return google::protobuf::util::MessageDifferencer::Equals(
+              newFs, existFs);
+    };
+
+    auto checkFsInfo = [fsType, volumeInfoComparator, s3InfoComparator](
+                           const FsDetail& newFs, const FsDetail& existFs) {
         switch (fsType) {
             case curvefs::common::FSType::TYPE_S3:
-                return MessageDifferencer::Equals(lhs.s3info(), rhs.s3info());
+                return s3InfoComparator(newFs.s3info(), existFs.s3info());
             case curvefs::common::FSType::TYPE_VOLUME:
-                return volumeInfoComparator(lhs.volume(), rhs.volume());
+                return volumeInfoComparator(newFs.volume(), existFs.volume());
             case curvefs::common::FSType::TYPE_HYBRID:
-                return MessageDifferencer::Equals(lhs.s3info(), rhs.s3info()) &&
-                       volumeInfoComparator(lhs.volume(), rhs.volume());
+                return s3InfoComparator(newFs.s3info(), existFs.s3info()) &&
+                       volumeInfoComparator(newFs.volume(), existFs.volume());
         }
 
         return false;
--- a/curvefs/test/mds/fs_manager_test2.cpp
+++ b/curvefs/test/mds/fs_manager_test2.cpp
@@ -241,6 +241,31 @@ TEST_F(FsManagerTest2, CreateFoundConflictFsNameAndNotIdenticalToPreviousOne) {
 
         EXPECT_EQ(FSStatusCode::FS_EXIST, fsManager_->CreateFs(&req, nullptr));
     }
+
+    // prefix is different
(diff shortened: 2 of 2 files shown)


201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
@Cyber-SiKu PTAL

curvefs/src/mds/fs_manager.cpp
```diff
@@ -795,15 +795,22 @@ int FsManager::IsExactlySameOrCreateUnComplete(const std::string& fsName,
         return google::protobuf::util::MessageDifferencer::Equals(lhs, rhs);
     };
 
-    auto checkFsInfo = [fsType, volumeInfoComparator](const FsDetail& lhs,
-                                                      const FsDetail& rhs) {
+    auto s3InfoComparator = [](common::S3Info lhs, common::S3Info rhs) {
+        // for compatible with old clients
+        lhs.clear_objectprefix();
+        rhs.clear_objectprefix();
```
Cyber-SiKu, line 801: I think we should not compare this directly, but determine whether there is this attribute (has_objectprefix).
1. If not, it means that the request is made by an old client. At this time, the attribute of fs should be required to be 0 (refer to curvefs/conf/tools.conf). If it is other, it should be wrong.
2. If available, just compare directly.

201341: there is no need to complicate this logic, the object prefix can't change anymore once create. So we don't need to compare this attribute.

Cyber-SiKu: > there is no need to complicate this logic, the object prefix can't change anymore once create. So we don't need to compare this attribute.

1. If the relevant attributes of the client and fs are different, it will lead to data inconsistency.
2. The other thing is to be fool-proof. This check can prevent users from using different objectprefix attributes to mount (curveadm's mounting process includes createfs).

201341: ok

201341:
cicheck

201341:
@Cyber-SiKu PTAL

curvefs/src/mds/fs_manager.cpp
```diff
@@ -795,15 +795,27 @@ int FsManager::IsExactlySameOrCreateUnComplete(const std::string& fsName,
...
-    auto checkFsInfo = [fsType, volumeInfoComparator](const FsDetail& lhs,
-                                                      const FsDetail& rhs) {
+    auto s3InfoComparator = [](common::S3Info lhs, common::S3Info rhs) {
+        // for compatible with old clients
+        if (lhs.has_objectprefix() && !rhs.has_objectprefix()) {
+            return lhs.objectprefix() == 0;
+        }
+        if (rhs.has_objectprefix() && !lhs.has_objectprefix()) {
+            return rhs.objectprefix() == 0;
+        }
```
Cyber-SiKu, line 805: Doesn't this compare to other content?

201341: fix

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

Cyber-SiKu (approved):

curvefs/src/mds/fs_manager.cpp
```diff
@@ -808,15 +808,29 @@ int FsManager::IsExactlySameOrCreateUnComplete(const std::string& fsName,
...
+        if (lhs.has_objectprefix() && !rhs.has_objectprefix() &&
+            lhs.objectprefix() != 0) {
+            return false;
+        }
+        if (rhs.has_objectprefix() && !lhs.has_objectprefix() &&
+            rhs.objectprefix() != 0) {
+            return false;
+        }
+
+        return google::protobuf::util::MessageDifferencer::Equals(lhs, rhs);
```
wu-hanqing, line 822:
```suggestion
        if (lhs.has_prefix() && rhs.has_prefix() && (lhs.prefix() != rhs.prefix())) {
            return false;
        }

        lhs.clear_prefix();
        rhs.clear_prefix();

        return Equals(lhs, rhs);
```

Are they equivalent? If so, this might look clearer. BTW, pls add corresponding tests.

201341: No, it's different. This change is mainly for compatibility with old clients which don't have object_prefix attribute. For test, I will add it later.

wu-hanqing: > No, it's different. This change is mainly for compatibility with old clients which don't have object_prefix attribute. For test, I will add it later.

The problem is clear, and, as my understanding, we just need to ensure if both have prefix, then they must be same, otherwise, we can ignore this field.

201341: It can't cover the condition that the old client which doesn't have prefix, and the new client has the prefix which is 1 .

201341:
cicheck

201341:
cicheck

201341:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

Cyber-SiKu:
cicheck

201341:
@wu-hanqing PTAL

curvefs/src/mds/fs_manager.cpp
```diff
@@ -808,15 +806,29 @@ int FsManager::IsExactlySameOrCreateUnComplete(const std::string& fsName,
         return google::protobuf::util::MessageDifferencer::Equals(lhs, rhs);
     };
 
-    auto checkFsInfo = [fsType, volumeInfoComparator](const FsDetail& lhs,
-                                                      const FsDetail& rhs) {
+    auto s3InfoComparator = [](common::S3Info lhs, common::S3Info rhs) {
+        // for compatible with old clients
```
wu-hanqing, line 810: Please add more comments to explain scenarios that require compatibility, and renaming parameters because it is impossible to clearly distinguish which is the MDS side and which is the fs the current client is preparing to create.

201341: The parameters' meaning depends on the caller, If changing the parameters, the caller must pass the parameter in the correct order. Is there any need to make changes?

201341:
cicheck

201341:
cicheck

201341:
cicheck

201341:
cicheck

curvefs/src/mds/fs_manager.cpp
wu-hanqing: > The parameters' meaning depends on the caller, If changing the parameters, the caller must pass the parameter in the correct order. Is there any need to make changes?

The parameters of a function are part of the calling convention and should not depend on the caller. The caller needs to pass in the parameters according to the parameters of the function as specified.

201341:
cicheck

201341:
cicheck

201341:
cicheck

curvefs/src/mds/fs_manager.cpp
201341: done

wu-hanqing (approved):

PR #2792, opened 2023-10-11
curvefs/client: optimizing the read amplification problem(especial me…

wuhongsong:
…mcached)

memcached does not support partial reads (object size：4MB in Curve) of an object, but the io will not exceed 128KB in linux kernel(1MB in kernel 4.20+).so for sequential reads, this will cause memcached write amplification is：4MB/128KB = 40 .

To solve this problem, we cache the first read memcached object locally.

### DIFF
--- a/curvefs/conf/client.conf
+++ b/curvefs/conf/client.conf
@@ -204,6 +204,18 @@ s3.writeCacheMaxByte=838860800
 s3.readCacheMaxByte=209715200
 # file cache read thread num
 s3.readCacheThreads=5
+
+# The data in the cache cluster download to local
+s3.memClusterToLocal=true
+# The data in the s3 storage download to local
+s3.s3ToLocal=true
+# read size bigger than this value will read until prefetch is finished
+s3.bigIoSize=131072
+# retry times when read big io failed
+s3.bigIoRetryTimes=100
+# retry interval when read big io failed
+s3.bigIoRetryIntervalUs=100
+
 # http = 0, https = 1
 s3.http_scheme=0
 s3.verify_SSL=False
@@ -247,7 +259,7 @@ diskCache.asyncLoadPeriodMs=5
 #        ok           nearfull               full
 #  |------------|-------------------|----------------------|
 #  0     trimRatio*safeRatio    safeRatio               fullRatio
-# 
+#
 #  1. 0<=ok<trimRatio*safeRatio;
 #  2. trimRatio*safeRatio<=nearfull<safeRatio
 #  3. safeRatio<=full<=fullRatio
--- a/curvefs/src/client/s3/client_s3_cache_manager.cpp
+++ b/curvefs/src/client/s3/client_s3_cache_manager.cpp
@@ -28,6 +28,8 @@
 #include <utility>
 
 #include "absl/cleanup/cleanup.h"
+#include "absl/strings/str_split.h"
+#include "absl/strings/string_view.h"
 #include "absl/synchronization/blocking_counter.h"
 #include "curvefs/src/client/kvclient/kvclient_manager.h"
 #include "curvefs/src/client/metric/client_metric.h"
@@ -37,6 +39,13 @@
 
 namespace curvefs {
 namespace client {
+
+DECLARE_bool(memClusterToLocal);
+DECLARE_bool(s3ToLocal);
+DECLARE_uint32(bigIoSize);
+DECLARE_uint32(bigIoRetryTimes);
+DECLARE_uint32(bigIoRetryIntervalUs);
+
 namespace common {
 DECLARE_bool(enableCto);
 }  // namespace common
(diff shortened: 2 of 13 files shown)


wuhongsong:
cicheck

wuhongsong:
cicheck

wuhongsong:
cicheck

wuhongsong:
cicheck

wuhongsong:
cicheck

wuhongsong:
cicheck

wuhongsong:
cicheck

wuhongsong:
cicheck

wu-hanqing (approved):

curvefs/conf/client.conf
```diff
@@ -204,6 +204,18 @@ s3.writeCacheMaxByte=838860800
 s3.readCacheMaxByte=209715200
 # file cache read thread num
 s3.readCacheThreads=5
+
+# The data in the cache cluster download to local
+s3.memClusterToLocal=true
```
201341, line 209: The s3.memClusterToLocal and s3.s3ToLocal can be merged into one options (maybe s3.S3ToLocalFirst)?

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -591,8 +614,34 @@ void FileCacheManager::ProcessKVRequest(const S3ReadRequest &req, char *dataBuf,
...
+                waitDownloading = true;
+            } else {
+                VLOG(9) << "obj is not in downloading: " << prefetchName
+                        << ", size: " << downloadingObj_.size() << ", "
+                        << downloadingObj_.size();
+                break;
+            }
+        }
+        if (waitDownloading) {
+            bthread_usleep(FLAGS_bigIoRetryIntervalUs * 10);
```
201341, line 638: Why multiply by 10？

curvefs/conf/client.conf
wuhongsong: these are two separate modules and functions. so ...

curvefs/src/client/s3/client_s3_cache_manager.cpp
wuhongsong: sorry， just miss it, fix it now.

curvefs/test/client/client_prefetch_test.cpp
```diff
@@ -0,0 +1,307 @@
+/*
+ *  Copyright (c) 2021 NetEase Inc.
```
Wine93, line 2: `2021` -> `2023`

curvefs/src/client/s3/client_s3_cache_manager.cpp
```diff
@@ -482,10 +491,24 @@ bool FileCacheManager::ReadKVRequestFromLocalCache(const std::string &name,
...
+        len >= FLAGS_bigIoSize && s3ClientAdaptor_->GetPrefetchBlocks()) {
+        int retry = 0;
+        do {
+            VLOG(6) << "wait for download object: " << name;
+            bthread_usleep(FLAGS_bigIoRetryIntervalUs);
+            if (++retry >= FLAGS_bigIoRetryTimes) {
+                LOG(WARNING) << "download object: " << name << " timeout";
+                return false;
+            }
+        } while (!s3ClientAdaptor_->GetDiskCacheManager()->IsCached(name));
```
Wine93, line 507: There are multi same operations, maybe you can use one function to replace this chain operation.

curvefs/conf/client.conf
201341: It is a little confused that s3.memClusterToLocal And s3.s3ToLocal are both true or false?

curvefs/test/client/client_prefetch_test.cpp
wuhongsong: fix

wuhongsong:
cicheck

wuhongsong:
cicheck

wuhongsong:
cicheck

curvefs/src/client/s3/client_s3_cache_manager.cpp
wuhongsong: fix

curvefs/conf/client.conf
wuhongsong: > It is a little confused that s3.memClusterToLocal And s3.s3ToLocal are both true or false?

they are independent, as you see follows：
` 
  

    if (!kvClientManager_ && FLAGS_s3ToLocal) {   // It is configurable whether to write to local cache or not
        // get from s3 directly
        PrefetchS3Objs(prefetchObjs, true);
    } else if (FLAGS_memClusterToLocal) {
        // get from memcached first, if failed, get from s3
        PrefetchS3Objs(prefetchObjs, false);
    }
`

- if memcached cluster is exist:
  - FLAGS_memClusterToLocal=true: download objs from memcached
  - FLAGS_memClusterToLocal=false: download is disabled
-  if memcached cluster is not exist:
  -  FLAGS_s3ToLocal=true: download objs from s3
  - FLAGS_s3ToLocal=false: download is disabled

Wine93 (approved):

wuhongsong:
@201341  PTAL

curvefs/conf/client.conf
201341: ok

201341 (approved):

PR #2800, opened 2023-10-12
Print error code

NopeDl:

Issue Number: #2185 

What's Changed: print error code with more meaningful name

How it Works: use the function to convert error code to string

### Check List


NopeDl:
cicheck

wu-hanqing:
cicheck

YunhuiChen:
cicheck

Cyber-SiKu:
cicheck

NopeDl:
cicheck

src/mds/common/mds_define.h
```diff
@@ -47,6 +47,27 @@ const int kCsClientReturnFail = -5;
 // error code: chunkserver offline
 const int kCsClientCSOffline = -6;
 
+const char* PrintMdsDescByErrorCode(int code) {
```
wu-hanqing, line 50:
```suggestion
inline const char* MdsErrCodeToName(int code) {
```

src/mds/common/mds_define.h
```diff
@@ -104,6 +125,59 @@ const int kTopoErrCodePoolsetNotFound = -20;
 const int kTopoErrCodeCannotDeleteDefaultPoolset = -21;
 const int kTopoErrCodeConflictBlockSizeAndChunkSize = -22;
 
+const char* PrintTopoErrCodeDescription(int code) {
```
wu-hanqing, line 128:
```suggestion
inline const char* TopoErrCodeToName(int code) {
```

wuhongsong:
@NopeDl  continue?

src/mds/common/mds_define.h
```diff
@@ -47,6 +47,27 @@ const int kCsClientReturnFail = -5;
 // error code: chunkserver offline
 const int kCsClientCSOffline = -6;
 
+inline char* MdsErrCodeToName(int code) {
```
Ziy1-Tan, line 50:
```suggestion
inline const char* MdsErrCodeToName(int code) {
```

Ziy1-Tan:
cicheck

Issue #2801, opened 2023-10-12
[tools-v2] character `%` in error messages produce unintended output


saicaca:
**Describe the bug (描述bug)**
The character `%` in an error message will be recognized as the prefix of a format verb, which produces `(MISSING)` in the output string.

**To Reproduce (复现方法)**

**Expected behavior (期望行为)**
The character `%` should be printed directly.

**Versions (各种版本)**
OS:
Compiler:
branch:
commit id:

**Additional context/screenshots (更多上下文/截图)**
![image](https://github.com/opencurve/curve/assets/25200299/191899ef-0462-4fae-b9b3-4472f7f38f13)

saicaca:
Replaced `fmt.Errorf(ce.Message)` with `errors.New(ce.Message)` in `CmdError#ToError` should fix the issue. #2802

PR #2802, opened 2023-10-12
[fix] tools-v2: fix incorrect error output

saicaca:

Issue Number: #2801

Problem Summary: Error messages contains character `%` cannot be output correctly because `%` is treated as the prefix of a format verb.

What's Changed: Replaced `fmt.Errorf(ce.Message)` with `errors.New(ce.Message)` in `ToError()`

How it Works: `errors.New` treats the error messages as plain text instead of format strings.

### Check List

### DIFF
--- a/tools-v2/internal/error/error.go
+++ b/tools-v2/internal/error/error.go
@@ -23,6 +23,7 @@
 package cmderror
 
 import (
+	"errors"
 	"fmt"
 
 	fscopyset "github.com/opencurve/curve/tools-v2/proto/curvefs/proto/copyset"
@@ -69,7 +70,7 @@ func (ce *CmdError) ToError() error {
 	if ce.Code == CODE_SUCCESS {
 		return nil
 	}
-	return fmt.Errorf(ce.Message)
+	return errors.New(ce.Message)
 }
 
 func NewSucessCmdError() *CmdError {


saicaca:
cicheck

tools-v2/internal/error/error.go
```diff
@@ -69,7 +70,7 @@ func (ce *CmdError) ToError() error {
 	if ce.Code == CODE_SUCCESS {
 		return nil
 	}
-	return fmt.Errorf(ce.Message)
+	return errors.New(ce.Message)
```
Cyber-SiKu, line 73: The effect is the same, why is it modified like this?

saicaca: As described in #2801, `fmt.Errorf` treats the input string as a format string. This can be problematic when the error message contains a `%` character (For example a web URL with non-ASCII characters). In such cases, the % character is mistakenly treated as the beginning of a format verb, leading to incorrect output. Using `errors.New` can fix this problem.

To illustrate, the following code snippet
```go
msg := "GET failed: https://example.com?msg=hello%20world"
err1 := fmt.Errorf(msg)
err2 := errors.New(msg)
fmt.Println(err1.Error())
fmt.Println(err2.Error())
```
outputs
```
GET failed: https://example.com?msg=hello%!w(MISSING)orld
GET failed: https://example.com?msg=hello%20world
```

baytan0720:
cicheck

Cyber-SiKu (approved):
LGTM!

baytan0720 (approved):

PR #2824, opened 2023-10-20
[feat]curvefs/client: split io attr/data latency

Cyber-SiKu:

Issue Number: #xxx 

Problem Summary:

image

What's Changed:

How it Works:

### Check List

### DIFF
--- a/curvefs/src/client/fuse_s3_client.cpp
+++ b/curvefs/src/client/fuse_s3_client.cpp
@@ -20,11 +20,12 @@
  * Author: xuchaojie
  */
 
+#include "curvefs/src/client/fuse_s3_client.h"
 
+#include <cstdint>
 #include <memory>
 #include <vector>
 
-#include "curvefs/src/client/fuse_s3_client.h"
 #include "curvefs/src/client/kvclient/memcache_client.h"
 
 namespace curvefs {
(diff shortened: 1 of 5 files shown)


Cyber-SiKu:
cicheck

SeanHai (approved):

curvefs/src/client/fuse_s3_client.cpp
```diff
@@ -220,6 +217,15 @@ CURVEFS_ERROR FuseS3Client::FuseOpWrite(fuse_req_t req, fuse_ino_t ino,
     }
 
     inodeWrapper->GetInodeAttrLocked(&fileOut->attr);
+
+    if (fsMetric_.get() != nullptr) {
```
wu-hanqing, line 221: should we remove this metric?

curvefs/src/client/fuse_s3_client.cpp
```diff
@@ -263,17 +271,18 @@ CURVEFS_ERROR FuseS3Client::FuseOpRead(fuse_req_t req, fuse_ino_t ino,
...
         fsMetric_->userRead.qps.count << 1;
         uint64_t duration = butil::cpuwide_time_us() - start;
         fsMetric_->userRead.latency << duration;
         fsMetric_->userReadIoSize.set_value(rRet);
     }
-
-    ::curve::common::UniqueLock lgGuard = inodeWrapper->GetUniqueLock();
-    inodeWrapper->UpdateTimestampLocked(kAccessTime);
-    inodeManager_->ShipToFlush(inodeWrapper);
+    ioLatencyMetric_->readDataLatency << butil::cpuwide_time_us() - mid;
```
wu-hanqing, line 285: read data latency should put before L274?

curvefs/src/client/fuse_s3_client.cpp
Cyber-SiKu: no, this pr only care about latency.

wu-hanqing (approved):

PR #2892, opened 2023-11-13
curvefs: support client mdsaddrs override

h0hmj:
we need to make client alive when migrate mds to new ones. so we add online client mds addrs override to broadcast by session. 

tool -> send SetClientMdsAddrsOverride to mds, will keep active mds
mds -> refresh session send to client
client -> change mds addr to override ones

use case:
client -> mds A,B,C
change to client -> mds A,B,C,D,E,F
change to client -> mds D,E,F

### DIFF
--- a/curvefs/proto/mds.proto
+++ b/curvefs/proto/mds.proto
@@ -236,6 +236,14 @@ message CommitTxResponse {
     required FSStatusCode statusCode = 1;
 }
 
+message SetClientMdsAddrsOverrideRequest {
+    required string clientMdsAddrsOverride = 1;
+}
+
+message SetClientMdsAddrsOverrideResponse {
+    required FSStatusCode statusCode = 1;
+}
+
 service MdsService {
     // fs interface
     rpc CreateFs(CreateFsRequest) returns (CreateFsResponse);
@@ -253,4 +261,7 @@ service MdsService {
 
     // client lease
     rpc RefreshSession(RefreshSessionRequest) returns (RefreshSessionResponse);
+
+    // client mds addrs override, for mds migration
+    rpc SetClientMdsAddrsOverride(SetClientMdsAddrsOverrideRequest) returns (SetClientMdsAddrsOverrideResponse);
 }
--- a/curvefs/proto/mds.proto
+++ b/curvefs/proto/mds.proto
@@ -194,6 +194,7 @@ message RefreshSessionRequest {
     required string fsName = 2;
     required Mountpoint mountpoint = 3;
     optional FsDelta fsDelta = 4;
+    optional string mdsAddrs = 5;
 }
 
 message RefreshSessionResponse {
@@ -202,6 +203,7 @@ message RefreshSessionResponse {
     optional bool enableSumInDir = 3;
     optional uint64 fsCapacity = 4;
     optional uint64 fsUsedBytes = 5;
+    optional string mdsAddrsOverride = 6;
 }
 
 message DLockValue {
--- a/curvefs/src/client/rpcclient/mds_client.cpp
+++ b/curvefs/src/client/rpcclient/mds_client.cpp
@@ -25,6 +25,7 @@
 #include <map>
 #include <utility>
 #include <vector>
+#include "absl/strings/str_join.h"
 
 #include "curvefs/proto/space.pb.h"
 #include "curvefs/src/client/rpcclient/fsdelta_updater.h"
@@ -500,12 +501,11 @@ FSStatusCode MdsClientImpl::AllocS3ChunkId(uint32_t fsId, uint32_t idNum,
     return ReturnError(rpcexcutor_.DoRPCTask(task, mdsOpt_.mdsMaxRetryMS));
 }
 
-FSStatusCode
-MdsClientImpl::RefreshSession(const std::vector<PartitionTxId> &txIds,
-                              std::vector<PartitionTxId> *latestTxIdList,
-                              const std::string& fsName,
-                              const Mountpoint& mountpoint,
-                              std::atomic<bool>* enableSumInDir) {
+FSStatusCode MdsClientImpl::RefreshSession(
+    const std::vector<PartitionTxId> &txIds,
+    std::vector<PartitionTxId> *latestTxIdList, const std::string &fsName,
+    const Mountpoint &mountpoint, std::atomic<bool> *enableSumInDir,
+    const std::string &mdsAddrs, std::string *mdsAddrsOverride) {
     auto task = RPCTask {
         (void)addrindex;
         (void)rpctimeoutMS;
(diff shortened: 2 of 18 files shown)


h0hmj:
cicheck

curvefs/src/mds/fs_manager.cpp
```diff
@@ -1280,5 +1289,10 @@ bool FsManager::FillVolumeInfo(common::Volume* volume) {
     return true;
 }
 
+void FsManager::SetClientMdsAddrsOverride(const std::string& addrs) {
+    WriteLockGuard lock(clientMdsAddrsOverrideMutex_);
+    clientMdsAddrsOverride_ = addrs;
```
wu-hanqing, line 1294: It's better to do some basic verification of the IP addresses, otherwise, it may cause client exception.

Another point is that it is better to update the addresses gradually, you cannot change all addresses directly. For example, you must change from (A,B,C) -> (A,B,C,D,E,F) -> (D, E, F).

wu-hanqing: BTW, should you persist mds addresses into etcd? Becuase there's chance that mds exit before client get new addresses.

curvefs/src/client/rpcclient/mds_client.cpp
```diff
@@ -550,7 +552,17 @@ MdsClientImpl::RefreshSession(const std::vector<PartitionTxId> &txIds,
             FsQuotaChecker::GetInstance().UpdateQuotaCache(
                 response.fscapacity(), response.fsusedbytes());
         }
-
+        if (response.has_mdsaddrsoverride()) {
```
wu-hanqing, line 555: it's not a good idea to put heavy logic in here, `MdsClientImpl` is just used for communicate with MDS

curvefs/src/client/rpcclient/mds_client.cpp
```diff
@@ -550,7 +552,17 @@ MdsClientImpl::RefreshSession(const std::vector<PartitionTxId> &txIds,
...
+            std::vector<std::string> mdsAddrsOverride = {};
+            curve::common::SplitString(response.mdsaddrsoverride(), ",",
+                                       &mdsAddrsOverride);
+            if (!mdsAddrsOverride.empty()) {
+                mdsOpt_.rpcRetryOpt.addrs = mdsAddrsOverride;
```
wu-hanqing, line 560: data race

src/client/mds_client.cpp
```diff
@@ -191,10 +195,15 @@ int RPCExcutorRetryPolicy::GetNextMDSIndex(bool needChangeMDS,
...
-    assert(mdsindex >= 0 &&
-           mdsindex < static_cast<int>(retryOpt_.addrs.size()));
-
-    const std::string &mdsaddr = retryOpt_.addrs[mdsindex];
+    std::string mdsaddr;
+    {
+        ReadLockGuard lock(retryOptLock_);
+        // happen when mds scaling down
+        if (mdsindex >= static_cast<int>(retryOpt_.addrs.size())) {
+            return -EHOSTDOWN;
```
wu-hanqing, line 203: return `-brpc::ELOGOFF` is better

curvefs/src/client/rpcclient/mds_client.cpp
h0hmj: mdsOpt_.rpcRetryOpt in MdsClientImpl is only used by Init() seems, so no data race i think, will leave a comment to hint

curvefs/src/mds/fs_manager.cpp
h0hmj: > It's better to do some basic verification of the IP addresses, otherwise, it may cause client exception.
> 
> Another point is that it is better to update the addresses gradually, you cannot change all addresses directly. For example, you must change from (A,B,C) -> (A,B,C,D,E,F) -> (D, E, F).

ok i will make sure active mds is in the override ones

h0hmj: > BTW, should you persist mds addresses into etcd? Becuase there's chance that mds exit before client get new addresses.

no. it's just a temporary online advanced setting.

src/client/mds_client.cpp
h0hmj: ok

curvefs/src/client/rpcclient/mds_client.cpp
h0hmj: ok

curvefs/src/client/rpcclient/mds_client.cpp
h0hmj: ok, will add safe get&set for rpcexcutor_'s retryOption

YunhuiChen:
cicheck

YunhuiChen:
cicheck

h0hmj:
cicheck

h0hmj:
cicheck

h0hmj:
cicheck

h0hmj:
cicheck

curvefs/proto/mds.proto
```diff
@@ -234,6 +236,14 @@ message CommitTxResponse {
     required FSStatusCode statusCode = 1;
 }
 
+message SetClientMdsAddrsOverrideRequest {
```
SeanHai, line 239: `SetClientMdsAddrsOverrideRequest` to `SetMdsAddrsOverrideRequest` ?

curvefs/src/mds/fs_manager.cpp
```diff
@@ -976,6 +976,15 @@ void FsManager::RefreshSession(const RefreshSessionRequest* request,
     FsUsage usage;
     fsStorage_->GetFsUsage(request->fsname(), &usage, true);
     response->set_fsusedbytes(usage.usedbytes());
+    {
+        ReadLockGuard lock(clientMdsAddrsOverrideMutex_);
```
SeanHai, line 980: Ditto `clientMdsAddrsOverrideMutex_` to `mdsAddrsOverrideMutex_`

curvefs/src/mds/fs_manager.cpp
```diff
@@ -976,6 +976,15 @@ void FsManager::RefreshSession(const RefreshSessionRequest* request,
...
+        VLOG(6) << "clientMdsAddrsOverride_ = " << clientMdsAddrsOverride_
+                << ", request->mdsaddrs() = " << request->mdsaddrs();
+        if (!clientMdsAddrsOverride_.empty() && request->has_mdsaddrs() &&
+            request->mdsaddrs() != clientMdsAddrsOverride_) {
```
SeanHai, line 984: Different mds address sequences may cause inequality

curvefs/src/mds/mds.cpp
```diff
@@ -163,6 +163,7 @@ void MDS::InitFsManagerOptions(FsManagerOption* fsManagerOption) {
            "default value: "
         << fsManagerOption->spaceReloadConcurrency;
 
+    fsManagerOption->mdsListenAddr = conf_->GetStringValue("mds.listen.addr");
```
SeanHai, line 166: `mds.listen.addr` here is all mds addr, not the leader only

SeanHai:
New unit test cases are needed to verify the code you added.

curvefs/src/mds/mds.cpp
h0hmj: it's the mds addr read from conf

h0hmj:
cicheck

h0hmj:
> New unit test cases are needed to verify the code you added.

yes. new uts have been added.

h0hmj:
cicheck

wu-hanqing (approved):

SeanHai (approved):

Issue #2783, opened 2023-10-07
[request help] curvefs support x-amz-storage-class


aspirer:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**
curvefs needs to support the x-amz-storage-class header so that I can store data into the specified s3 storage class level, such as Standard-IA storage (the default is standard storage).

**Describe the solution you'd like (描述你期望的解决方法)**

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

Songjf-ttk:
Hello, I am a student from Hunan University. My major is Software Engineering. I am familiar with C++ and Golang, and I am very interested in cloud-native and storage systems. My technical strengths include C++ server development and kernel virtualization techniques. I chose this issue because I have some knowledge of S3 and want to deepen my understanding through the process of completing the topic. I am confident in completing this topic, and I have previously submitted two PRs to the community：

1.[Add debug mode](https://github.com/opencurve/curveadm/pull/303)
2.[Fix variable position](https://github.com/opencurve/curveadm/pull/305)

Tianpingan:
Hello, i wanna solve this issue, please assign this to me, thanks. 
I am a graduate student in the Zhejiang university. I am familiar with programming languages C++ and Rust. I am interested in  distributed systems and storage systems. The reason why I chose this issue is because I believe I can complete it and I am eager to delve into the project's design and implementation details to further enhance my knowledge. Thank you for considering my request.

Subzeroo7:
> Hello, i wanna solve this issue, please assign this to me, thanks. I am a graduate student in the Zhejiang university. I am familiar with programming languages C++ and Rust. I am interested in distributed systems and storage systems. The reason why I chose this issue is because I believe I can complete it and I am eager to delve into the project's design and implementation details to further enhance my knowledge. Thank you for considering my request.

Hi~Use WeChat to add the Curve community assistant "Curve"**（ID：OpenCurve_bot）** as a WeChat friend, tell your GitHub ID, and send the Issue link of your choice. The assistant will invite you to communicate with the mentor  by WeChat.

PR #2926, opened 2023-11-23
curvefs: support x-amz-storage-class

Tianpingan:

Issue Number: #2783

Problem Summary:

What's Changed:
1. Add support for PutObjectOptions in PutObject and PutObjectAsync methods, enabling the setting of storage class.
2. Polish the client to align with updated PutObjectAsync method.
3. Polish the meta_server s3compaction to align with updated PutObject method.

How it Works:

When creating a request, the configuration for storage class is encompassed within the PutObjectOptions, applied alongside both PutObject and PutObjectAsync methods. 
This allows us to employ PutObjectOptions to construct the S3 request, including the storage class information.

### Check List


aspirer:
1. I don’t see that the s3info-related RPC request uses your newly added storageclass parameter. What is its function?
2. Is the storage class configured at the file system level or the fuse client level? I saw that the storageclass parameter was added to the rpc parameters of your s3info, but the persistent storage operation was not seen on the mds side. How to use it later? In addition, I see that there is also a configuration item analysis of this parameter. What is the relationship between the two?
3. Our goal is to configure it at the file system level. Each file system configures the storage class it wants to use.

h0hmj:
> 1. I don’t see that the s3info-related RPC request uses your newly added storageclass parameter. What is its function?
> 2. Is the storage class configured at the file system level or the fuse client level? I saw that the storageclass parameter was added to the rpc parameters of your s3info, but the persistent storage operation was not seen on the mds side. How to use it later? In addition, I see that there is also a configuration item analysis of this parameter. What is the relationship between the two?
> 3. Our goal is to configure it at the file system level. Each file system configures the storage class it wants to use.

Q2&3 s3info is part of fsinfo in etcd

h0hmj:
need set storageclass when createfs.
maybe related changes
1. tool
2. createfs rpc

Tianpingan:
> need set storageclass when createfs. maybe related changes
> 
> 1. tool
> 2. createfs rpc

Got it. since the storage class is already part of S3info, so i decide to use the existing createfs rpc for tools, simply adding the storage class parameter.

Tianpingan:
> need set storageclass when createfs. maybe related changes
> 
> 1. tool
> 2. createfs rpc

i have updated the tool in createfs/src/tools and tools-v2 to support storage class

h0hmj:
cicheck

h0hmj:
cicheck

Tianpingan:
cicheck

h0hmj:
cicheck

h0hmj:
cicheck

h0hmj:
cicheck

h0hmj:
cicheck

h0hmj:
cicheck

Issue #2978, opened 2023-12-25
Nit(build): download dependency unsuccessfully


zztaki:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

**Describe the solution you'd like (描述你期望的解决方法)**
mount `/etc/hosts` to docker

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**
When I build using `make build stor=fs dep=1`, it will compile etcdclient and download go1.12.8 from golang.google.cn or dl.google.com first. 

However, since we did not mount the local hosts to docker. In docker, it may not be able to resolve the correct DNS address, resulting in the failure to successfully download go1.12.8. 

![image](https://github.com/opencurve/curve/assets/50295684/8843cf08-caf7-4f4c-a3cc-ea1579cd6f51)

I will submit a PR soon :)

Issue #2602, opened 2023-06-30
[tools-v2]:better build tools-v2


Cyber-SiKu:
**Is your feature request related to a problem? (你需要的功能是否与某个问题有关?)**

Currently compiling tools-v2 is in the main process of compiling and cannot be skipped. Even the same as below will compile. Please optimize this.

```bash
make build stor=fs release=1 only=curvefs/src/metaserver/*
```

**Describe the solution you'd like (描述你期望的解决方法)**

**Describe alternatives you've considered (描述你想到的折衷方案)**

**Additional context/screenshots (更多上下文/截图)**

zztaki:
Hello, I have located the location that needs to be modified. I would like to discuss with you when we need to compile tools-v2 and when not. :)

zztaki:
A simple way is if we specify `only` (alias `g_target`) then we don't compile it.

zztaki:
Or we add add an independent parameter (eg. `skiptools`) to specify whether to skip the compilation of tools-v2

montaguelhz:
Adding new parameters seems unnecessary, welcome to do it.

montaguelhz:
Adding new parameters seems unnecessary, welcome to do it. @zztaki

zztaki:
> Adding new parameters seems unnecessary, welcome to do it. @zztaki

Thank you for your help :) @montaguelhz

PR #2989, opened 2024-01-13
build: better build tools-v2 and mount /etc/hosts

zztaki:
Issue Number: #2602 and #2978 

Problem Summary:
1. tools-v2 is in the main process of compiling and cannot be skipped.
2. go1.12.8 dependency can't be downloaded due to resolving incorrect DNS address.

What's Changed:
1. When only(akw. target) is specified when building, we will no longer compile tools-v2. (Warning: this will also happen if you manually specify all modules for only)
2. mount `/etc/hosts` to docker.

How it Works:

### Check List


zztaki:
cicheck

